PR #8, opened 2023-10-12
Use Spanned

imaqtkatt:


developedby:
I'll close this for now, since I don't think we'll go with this approach. If we ever need it for some reason (probably will implement somehting like this in Bend lang) then we can look at the closed PR

PR #13, opened 2023-10-13
Definition pruning

imaqtkatt:


src/term/check/unbound_vars.rs
```diff
@@ -4,7 +4,7 @@ use std::collections::HashMap;
 
 impl DefinitionBook {
   pub fn check_unbound_vars(&self) -> anyhow::Result<()> {
-    for def in &self.defs {
+    for (_, def) in &self.defs {
```
developedby, line 7: You can do `for def in self.defs.values()` to iterate here. Same goes for the other identical `for`s.

src/term/transform/definition_pruning.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+impl Term {
+  /// Recursively finds all used definitions on every term that can have a def_id.
+  fn find_used_definitions(&self, used: &mut Definitions, defs: &BTreeMap<DefId, Definition>) {
+    match &self {
+      Term::Ref { def_id } => {
+        if used.insert(*def_id) {
+          let Definition { body, .. } = defs.get(def_id).unwrap();
+          body.find_used_definitions(used, defs);
```
developedby, line 30: If you recurse on both the subterms and the definitions at the same time, this will easily stack overflow for any program that isn't super small.
I recommend recursion only in the subterms since we know users will usually write definitions of reasonable size.

You can implement this by returning the newly found definitions here with `Term::find_used_definition(...) -> HashSet<DefId>` and then in `DefinitionBook::prune()` you can `find_used_definitions` on the new set of found definitions. Something like:
```
let mut to_visit = HashSet::new();
while let Some(def_id) = to_visit.pop() {
  let body = &self.defs[def_id].body;
  let new_defs = body.find_used_definitions(used, &self.defs);
  used.extend(new_defs.iter().copied());
  to_visit.extend(new_defs.into_iter());
}
```

developedby (changes requested):

developedby (approved):

PR #14, opened 2023-10-17
[sc-61] Refactor net intermediate representation

LunaAmora:


src/net/mod.rs
```diff
@@ -1,8 +1,176 @@
...
+    self.0
+  }
+
+  /// Returns the slot of a port.
+  pub fn slot(self) -> SlotId {
+    self.1
+  }
+}
+
+impl Op {
```
developedby, line 123: I know I was the one who implemented this, but I think these functions shouldn't be here, but rather only on the file they're used (one file uses the first and another uses the second one).

src/term/mod.rs
```diff
@@ -199,18 +199,18 @@ impl Term {
...
-            succ.to_string(def_names),
-          )
-        } else {
-          eprintln!("{self:?}");
-          panic!("Invalid match expression on stringification");
-        }
+        let (pred, succ) = match succ.as_ref() {
+          Term::Lam { nam, bod } => (nam, bod),
+          _ => (&None, succ),
+        };
```
developedby, line 205: A comment here explaining this here would be nice

developedby (approved):

PR #15, opened 2023-10-17
[sc-122] Rename entrypoint from Main to main

LunaAmora:


developedby:
Sorry, I forgot to ask in the story, but the compiled file also needs to be generated with a "main" entrypoint. Right now it just uses the assigned id which is an error

developedby (approved):

PR #18, opened 2023-10-18
[sc-103] Implement Pairs in hvm-lang

imaqtkatt:


src/term/term_to_net.rs
```diff
@@ -144,6 +145,20 @@ fn encode_term(
       inet.link(up, Port(node, 0));
       Ok(Some(Port(node, 0)))
     }
+    Term::Let { pat: LetPat::Tup(l_nam, r_nam), val, nxt } => {
+      let dup = inet.new_node(Dup { lab: u8::try_from(*dups).unwrap() });
```
developedby, line 149: I think it would be cleaner to have a new `NodeKind::Tup` instead of using Dup, even though they have the same behaviour.

src/term/term_to_net.rs
```diff
@@ -174,6 +189,17 @@ fn encode_term(
 
       Ok(Some(Port(snd_node, 2)))
     }
+    Term::Tup { fst, snd } => {
+      let tup = inet.new_node(Con { lab: 1 });
```
developedby, line 193: Is there som specific reason why you added labels to Con nodes? It seems that it only used here and the only used values are either 0 or 1.

I think you should remove the `lab` field and either compile to a Dup node like you did for Tup destructuring in the let case above (or use a Tup node type as I suggested).

src/term/transform/linearize.rs
```diff
@@ -72,6 +72,35 @@ fn term_to_affine(
...
+      let snd_uses = r_nam.as_ref().map(|r_nam| *var_uses.get(r_nam).unwrap()).unwrap_or(0);
+
+      match (fst_uses, snd_uses) {
+        (0, 0) => *nxt,
+        (_, 0) => match *val {
+          Term::Tup { fst, .. } => {
+            let let_ = Term::Let { pat: LetPat::Var(l_nam.unwrap()), val: fst, nxt };
+            term_to_affine(let_, var_uses, let_bodies)?
+          }
+          _ => unimplemented!(),
```
developedby, line 86: Tup is not the only body that we expect here. For example:
`((@x let (a, b) = x; (+ a b)) (1, 2))`
This is a valid term that in the context of this pass should remain the same.

developedby (changes requested):

src/term/mod.rs
```diff
@@ -277,6 +290,25 @@ impl Term {
...
       }
+      Term::Tup { fst, snd } => {
+        fst.subst(from, to);
+        snd.subst(from, to);
+      }
+    }
+  }
+}
+
+impl LetPat {
```
developedby, line 301: You can make this into a fmt::Display implementation (clippy will probably report this as well)

src/term/net_to_term.rs
```diff
@@ -25,7 +25,7 @@ pub fn net_to_term_non_linear(net: &INet, book: &DefinitionBook) -> (Term, bool)
         (Term::Era, valid)
       }
       // If we're visiting a con node...
-      Con => match next.slot() {
+      Con { .. } => match next.slot() {
```
developedby, line 28: You can remove the {..} here

src/term/net_to_term.rs
```diff
@@ -503,6 +582,7 @@ impl Term {
         snd.fix_names(id_counter, book);
       }
       Term::Let { .. } => unreachable!(),
+      Term::Tup { .. } => todo!(),
```
developedby, line 585: This needs to be done, should be identical to Sup one. I don't understand how it's not crashing by leaving this as a todo!(). If it's actualyl unreadcheable then put that instead. Also, maybe have a look if this is disabled for some reason?

src/term/parser/parser.rs
```diff
@@ -130,14 +130,24 @@ where
...
+      .then(term.clone())
+      .delimited_by(just(Token::LParen), just(Token::RParen))
+      .map(|(fst, snd)| Term::Tup { fst: Box::new(fst), snd: Box::new(snd) })
+      .boxed();
+
+    // let a = ...
+    // let (a, b) = ...
     let let_ = just(Token::Let)
-      .ignore_then(name())
+      .ignore_then(pat())
```
developedby, line 145: Please call this `let_pat` for consistency and avoid clashing with rule patterns. (At some point I think we'll merge both)

developedby (changes requested):

developedby (approved):
good work

PR #20, opened 2023-10-19
[sc-137] Add support for "Main" entrypoint for HVM1 compatiblity

LunaAmora:


src/term/check/mod.rs
```diff
@@ -1,10 +1,14 @@
-use super::{DefId, DefinitionBook, Name, Term};
+use super::{DefId, DefNames, DefinitionBook, Name, Term};
 
 pub mod unbound_vars;
 
 impl DefinitionBook {
   pub fn check_has_main(&self) -> anyhow::Result<DefId> {
-    if let Some(main) = self.def_names.def_id(&Name::new("main")) {
+    if let Some(main) = self
```
developedby, line 7: We should only allow one of the accepted entry point names. Otherwise it becomes confusing for the user in case they for some reason define the two names and expect the other one to be the entrypoint

developedby (changes requested):
Please also add a test of a file using Main as an entry point that correctly reads back a reference to Main (and another to main).
A test that runs a file with both Main and main would also be nice.

developedby:
Also please check that pruning is working. I have the impression that after changing the entrypoint to "main" something went wrong there

developedby (commented):
Alright, just rebase and it's ready to merge

developedby (approved):
Looks great

PR #21, opened 2023-10-19
[sc-135] Keep truncated name of defs when compiling

LunaAmora:


src/lib.rs
```diff
@@ -47,9 +53,9 @@ pub fn run_compiled(book: &Book, mem_size: usize) -> (Net, RunStats) {
 }
 
 pub fn run_book(mut book: DefinitionBook, mem_size: usize) -> anyhow::Result<(Term, DefNames, RunInfo)> {
-  let compiled = compile_book(&mut book)?;
+  let (compiled, id_map) = compile_book(&mut book)?;
```
developedby, line 56: Since inside `compile_book` we only use the left-to-right direction of the bimap and on readback we only use the `right-to-left` direction, I think it would be better to use just a normal HashMap.
Inside `compile_book` it can be just `id_to_hvmc_name` and then before returning you create a new HashMap with keys and values swapped with `let hvmc_name_to_id = id_to_hvmc_name.into_iter().map(|(k, v)| (v, k)).collect::<HashMap<_>>();`

developedby, line 56: Not a big fan of the name id_map, since we already have pretty much the same thing inside the definition book, but there it's DefId to hvm-lang name, while here it's DefId to hvmc name.
`id_map` feels like it could be either, so a bit confusing.

developedby: Another possible approach is to add this BiMap to the DefinitionBook or to the DefNames type, where it could be together with the other id to name map, which may be slightly more organized.
You could then make it calculate the hvmc name automatically when adding a definition to the book and it becomes accessible everywhere else without having to add extra arguments to the functions.

developedby (changes requested):

src/term/term_to_net.rs
```diff
@@ -1,17 +1,50 @@
...
+use bimap::BiHashMap;
+use hvmc::{
+  ast::{name_to_val, val_to_name},
+  run::Val,
+};
 use std::collections::HashMap;
 
-pub fn book_to_compact_nets(book: &DefinitionBook) -> anyhow::Result<Vec<(DefId, INet)>> {
-  let mut nets = Vec::new();
+pub fn book_to_compact_nets(
```
developedby, line 10: If you could change the name of this function to `book_to_nets` (also it's misspelled as compact, lol)

src/term/term_to_net.rs
```diff
@@ -1,17 +1,50 @@
...
+    let name = if def.def_id == main { "main".to_string() } else { def_id_to_name(book, def.def_id, &nets) };
+
+    id_map.insert(def.def_id, name_to_val(&name));
+    nets.insert(name, net);
+  }
+
+  Ok((nets, id_map))
+}
+
+pub fn def_id_to_name(book: &DefinitionBook, def_id: DefId, nets: &HashMap<String, INet>) -> String {
```
developedby, line 28: Can you add a comment on what this function does and what it's needed? I think someone who looks at this without context in the future might be a bit confused.

developedby (changes requested):

src/lib.rs
LunaAmora: I don't think the second aproach would be much better as the two functions that would use one of this Hashmaps don't need the DefinitionBook or DefNames as argument. So it would just change one argument for the other. Unless that is a good enough reason, as it simplifies the api

developedby (approved):
Looks good

PR #25, opened 2023-10-31
[sc-155] Add pattern matching to HVM2

imaqtkatt:


developedby (changes requested):
Also, add some tests that catch non-exhaustive patterns to the compilation test suite

src/term/check/exhaustiveness.rs
```diff
@@ -0,0 +1,463 @@
...
+    self.rules.iter().map(Rule::get_row).collect()
+  }
+}
+
+fn wildcard() -> RulePat {
+  RulePat::Var(Name(String::from("_")))
+}
+
+/// Specializes a row based on the first pattern.
+fn specialize(row: &Row, label: &Name, size: usize) -> Matrix {
```
developedby, line 29: What does "specialization" mean in this context? Can you describe it here?

src/term/check/exhaustiveness.rs
```diff
@@ -0,0 +1,463 @@
...
+          })
+          .collect::<Vec<RulePat>>();
+
+        let matrix = def.get_matrix();
+        let rule_arity = def.rules[0].arity();
+        // this is the default case to check if a definition is exhaustive
+        let case = vec![wildcard(); rule_arity];
+
+        // ctx_cons, ex: [Ctr("T"), Ctr("F")]
+        let mut ctx_cons = HashMap::<String, Vec<RulePat>>::new();
```
developedby, line 264: Both `ctx_cons` and `ctx_types` seem to be just different projections of already existing data. Why not just use `self.adts` and `self.ctrs` directly?

src/term/check/exhaustiveness.rs
```diff
@@ -0,0 +1,463 @@
...
+        }
+      } else {
+        false
+      }
+    }
+  }
+}
+
+impl Book {
+  pub fn check_exhaustiveness(&self) -> anyhow::Result<()> {
```
developedby, line 244: Could you add a general description of how this algorithm works? I already forgot the details and I'm failing to understand it now

src/term/check/type_check.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use crate::term::{Book, DefId, Name, RulePat};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Type {
+  Any,
+  Adt(Name),
+}
+
+type TypedDefinitions = HashMap<DefId, Vec<Type>>;
```
developedby, line 12: Minor semantic thing, but this map doesn't hold the definitions, only their types, so a name like `DefinitionTypes` would be more appropriate

src/term/transform/flatten.rs
```diff
@@ -0,0 +1,215 @@
...
+  }
+  a_implies_b
+}
+
+/// Flattens a rule with nested pattern matching.
+/// This creates at least one new definition from the top layer of flattening.
+/// The new extracted definition might still have nested patterns,
+/// so we flatten those until all patterns have only 1 layer of pattern matching.
+/// Those recursive flattening steps return a vector of new subdefinitions which are also returned here.
+/// If we encounter
```
developedby, line 103: This comment is incomplete?

developedby (changes requested):

src/term/transform/flatten.rs
imaqtkatt: Yes, and actually this code is not mine

src/term/check/exhaustiveness.rs
imaqtkatt: I could not use `self.ctrs` because I need the constructor's arity

src/term/check/type_check.rs
```diff
@@ -0,0 +1,70 @@
...
+pub type DefinitionTypes = HashMap<DefId, Vec<Type>>;
+
+impl Book {
+  /// Returns a HashMap from the definition id to the inferred pattern types
+  /// and checks the rules arities based on the first rule arity.
+  pub fn type_check(&self) -> anyhow::Result<DefinitionTypes> {
+    let mut def_types = HashMap::new();
+
+    for def in self.defs.values() {
+      let current_arity = def.arity();
```
developedby, line 21: Can you separate this into another function? So we can get the type of just a single definition given the existing type definitions

src/term/check/type_check.rs
```diff
@@ -0,0 +1,70 @@
...
+  Any,
+  Adt(Name),
+}
+
+pub type DefinitionTypes = HashMap<DefId, Vec<Type>>;
+
+impl Book {
+  /// Returns a HashMap from the definition id to the inferred pattern types
+  /// and checks the rules arities based on the first rule arity.
+  pub fn type_check(&self) -> anyhow::Result<DefinitionTypes> {
```
developedby, line 17: This is more than just type checking. It checks the type and then returns it. Maybe `infer_types` or `infer_def_types` would be better, since what we do is infer the types from the used constructors and return an error in case the inference fails.

src/term/check/type_check.rs
developedby: `infer_def_type`

src/term/check/unbound_vars.rs
```diff
@@ -1,11 +1,13 @@
...
+use crate::term::{Book, LetPat, Name, Term};
 use hvmc::run::Val;
 use std::collections::HashMap;
 
-impl DefinitionBook {
+impl Book {
   pub fn check_unbound_vars(&self) -> anyhow::Result<()> {
     for def in self.defs.values() {
-      def.body.check_unbound_vars()?;
+      for rule in def.rules.iter() {
```
developedby, line 8: Either this is called after resolving pattern matching and we know the defs only have 1 rule (and we check that and only access the first rule), or this is called before compiling the pattern matching and then we need to consider the variables in the pattern.
I think the first option is a bit easier

src/term/net_to_term.rs
```diff
@@ -328,10 +327,12 @@ pub fn net_to_term_linear(net: &INet, book: &DefinitionBook) -> (Term, bool) {
         _ => unreachable!(),
       },
       Ref { def_id } => {
-        if book.is_generated_rule(def_id) {
-          let rule = book.defs.get(&def_id).unwrap();
-
-          let mut term = rule.body.clone();
+        if book.is_generated_def(def_id) {
+          let def = book.defs.get(&def_id).unwrap();
+          if def.rules.len() > 0 {
```
developedby, line 332: Either this is unreachable and we know the reason why, or we should be doing something here.

developedby, line 332: It should be `def.rules.len() > 1`

src/term/net_to_term.rs
```diff
@@ -549,9 +550,12 @@ impl Term {
       Term::Chn { nam: _, bod } => bod.fix_names(id_counter, book),
       Term::Lnk { .. } => {}
       Term::Ref { def_id } => {
-        if book.is_generated_rule(*def_id) {
-          let rule = book.defs.get(def_id).unwrap();
-          let mut term = rule.body.clone();
+        if book.is_generated_def(*def_id) {
+          let def = book.defs.get(def_id).unwrap();
+          if def.rules.len() > 1 {
+            panic!("What to do in this case?");
```
developedby, line 556: Same thing here

src/term/parser/parser.rs
```diff
@@ -205,47 +205,101 @@ where
...
-        emitter.emit(Rich::custom(span, format!("Repeated definition '{}'", name)));
+  let arity_0 = name().map(|nam| (nam, 0));
+  let arity_n = name()
+    .then(name().repeated().collect::<Vec<_>>())
+    .delimited_by(just(Token::LParen), just(Token::RParen))
+    .map(|(nam, args)| (nam, args.len()));
+  let ctr = arity_0.or(arity_n);
+
+  let data =
+    name().try_map(|Name(nam), span| if nam == "data" { Ok(()) } else { Err(Rich::custom(span, "")) });
```
developedby, line 252: Can you make a function for parsing a soft keyword?  So this whole line, but you receive `"data"` as an argument to the function. So here we just call `let data = soft_keyword("data");`. This could then just be inlined in the line below, like

```
  soft_keyword("data")
    .ignore_then(name())
    .then_ignore(just(Token::Equals))
    .then(ctr.separated_by(just(Token::Or)).collect::<Vec<(Name, usize)>>())
    .map(|(name, ctrs)| (name, Adt { ctrs: ctrs.into_iter().collect() }))
```

developedby (changes requested):

src/term/net_to_term.rs
```diff
@@ -90,10 +87,12 @@ pub fn net_to_term_non_linear(net: &INet, book: &DefinitionBook) -> (Term, bool)
...
       },
       Ref { def_id } => {
-        if book.is_generated_rule(def_id) {
-          let rule = book.defs.get(&def_id).unwrap();
-
-          let mut term = rule.body.clone();
+        if book.is_generated_def(def_id) {
+          let def = book.defs.get(&def_id).unwrap();
+          if def.rules.len() > 1 {
+            panic!("What to do in this case?");
```
developedby, line 93: Can you make this like the rest?

src/term/parser/parser.rs
```diff
@@ -205,47 +212,100 @@ where
...
+  })
+}
+
+fn rule<'a, I>() -> impl Parser<'a, I, (Name, Rule), extra::Err<Rich<'a, Token>>>
 where
   I: ValueInput<'a, Token = Token, Span = SimpleSpan>,
 {
-  let lhs = choice((name(), name().delimited_by(just(Token::LParen), just(Token::RParen))));
+  let def_name =
+    name().try_map(|Name(nam), span| if nam != "data" { Ok(Name(nam)) } else { Err(Rich::custom(span, "")) });
```
developedby, line 237: Do we need to check that it's not called `data`?
If it's not inside `()` it'll be parsed as a datatype declaration and we don't need to check it here.
If it's inside `()` then we are certain that this is the name of function and there is no ambiguity, although it might not be best practice to choose this name.

developedby (changes requested):

developedby (approved):

PR #26, opened 2023-11-08
[sc-220] Add scott-encoding transformation

imaqtkatt:


src/term/transform/scott_generation.rs
```diff
@@ -0,0 +1,84 @@
...
+        for ctr in curr_ctrs.keys() {
+          ctrs.push(ctr.clone());
+          if ctr == ctr_name {
+            make_args(arity, &mut ctr_args);
+          }
+        }
+
+        let lam = make_lam(&ctr_args, &mut ctrs, ctr_name);
+
+        ctr_args.clear();
```
developedby, line 21: Why not just declare these variables inside the `for` loop?

src/term/transform/scott_generation.rs
```diff
@@ -0,0 +1,84 @@
...
+
+  let app = Term::Var { nam: ctr_name.clone() };
+
+  let app = args_to_apply
+    .into_iter()
+    .fold(app, |acc, nam| Term::App { fun: Box::new(acc), arg: Box::new(Term::Var { nam }) });
+
+  let lam = app;
+
+  lam_args.into_iter().fold(lam, |acc, arg| Term::Lam { nam: Some(arg), bod: Box::new(acc) })
```
developedby, line 60: I think it would be a bit more legible if first folded over the different constructors and then the over the arguments in a different statement.

src/term/transform/scott_generation.rs
```diff
@@ -0,0 +1,84 @@
...
+    let nam = Name(format!("a{n}"));
+    ctr_args.push(nam);
+  }
+}
+
+fn make_lam(ctr_args: &Vec<Name>, ctrs: &mut Vec<Name>, ctr_name: &Name) -> Term {
+  let args_to_apply = ctr_args.clone();
+  ctrs.reverse();
+
+  let mut lam_args = ctrs.clone();
```
developedby, line 47: Instead of cloning into another variable, why not iterate, clone each element and then fold? I think the code would be a bit cleaner:
```
let term = Term::Var {...};
let term = ctr_args.iter().cloned().fold(term, ...);
let term = ctrs.iter().rev().cloned().fold(term, ...);
let term = ctr_args.iter().cloned().fold(term, ...);
```

src/term/transform/scott_generation.rs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        let lam = make_lam(&ctr_args, &mut ctrs, ctr_name);
+
+        ctr_args.clear();
+        ctrs.clear();
+
+        let def_id = self.def_names.insert(ctr_name.clone());
+
+        let rule = Rule { def_id, pats: vec![], body: lam };
```
developedby, line 26: Consider inlining these two variables `rule` and `rules`. They are pretty straight forward and I don't think we gain by separating these lines.

developedby: Maybe just one of them:
`let rules = vec![Rule{...}];`
or
```
let rule = Rule {...};
let def = Definition {def_id, rules: vec![rule]};
```

src/term/transform/scott_generation.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        self.defs.insert(def_id, def);
+      }
+    }
+  }
+}
+
+fn make_args(args_count: &usize, ctr_args: &mut Vec<Name>) {
+  for n in 0 .. *args_count {
+    let nam = Name(format!("a{n}"));
```
developedby, line 38: Maybe we could use the names given by the user in the program.
Would look a bit better, but we gotta change the ADT.

src/term/mod.rs
```diff
@@ -1,11 +1,13 @@
 use bimap::{BiHashMap, Overwritten};
 use derive_more::{Display, From, Into};
 use hvmc::run::Val;
+use indexmap::IndexMap;
 use itertools::Itertools;
 use shrinkwraprs::Shrinkwrap;
 use std::{
   collections::{BTreeMap, HashMap},
   fmt,
+  hash::Hash,
```
developedby, line 10: unused import?

developedby (changes requested):
Also, please add some regression tests for compiling programs with ADT declarations. Would be nice to have in the test an example of a sum type, a product type and a more complex ADT

src/term/transform/scott_generation.rs
imaqtkatt: I think it's easier to make the application first and then pass it as body

src/term/transform/scott_generation.rs
```diff
@@ -40,24 +36,21 @@ fn make_args(args_count: &usize, ctr_args: &mut Vec<Name>) {
...
-fn make_lam(ctr_args: &Vec<Name>, ctrs: &mut Vec<Name>, ctr_name: &Name) -> Term {
-  let args_to_apply = ctr_args.clone();
-  ctrs.reverse();
-
-  let mut lam_args = ctrs.clone();
-  let mut lam_args_ctrs = ctr_args.clone();
-  lam_args_ctrs.reverse();
-  lam_args.extend(lam_args_ctrs);
+fn make_lam(ctr_args: Vec<Name>, ctrs: Vec<Name>, ctr_name: &Name) -> Term {
+  fn fold_lam(acc: Term, arg: Name) -> Term {
```
developedby, line 40: Can this be just a closure before its used?

```
let fold_lam = |acc, arg|  Term::Lam { nam: Some(arg), bod: Box::new(acc) };
let lam = ctrs.iter().rev().cloned().fold(app, fold_lam);
ctr_args.iter().rev().cloned().fold(lam, fold_lam)
```

src/term/transform/scott_generation.rs
```diff
@@ -5,26 +5,22 @@ impl Book {
...
 
-      let mut ctr_args = Vec::new();
-      let mut ctrs = Vec::new();
-
       for (ctr_name, arity) in &curr_ctrs {
+        let mut ctr_args = Vec::new();
+        let mut ctrs = Vec::new();
+
         for ctr in curr_ctrs.keys() {
           ctrs.push(ctr.clone());
```
developedby, line 13: Isn't `ctrs` just `adt.ctrs.keys().collect()`?

src/term/transform/scott_generation.rs
```diff
@@ -5,26 +5,22 @@ impl Book {
...
           if ctr == ctr_name {
             make_args(arity, &mut ctr_args);
```
developedby, line 15: Why is this inside an `if` that will only match once? You can just move this line outside the `for`

developedby (changes requested):

src/term/transform/scott_generation.rs
```diff
@@ -0,0 +1,69 @@
...
+
+impl Book {
+  pub fn generate_scott_adts(&mut self) {
+    for adt in self.adts.values() {
+      let curr_ctrs = adt.ctrs.clone();
+
+      for (ctr_name, arity) in &curr_ctrs {
+        let mut ctr_args = Vec::new();
+        let ctrs = curr_ctrs.keys().cloned().collect();
+        make_args(arity, &mut ctr_args);
```
developedby, line 11: You could make this function return `ctr_args` instead for a cleaner interface.

`let ctr_args = make_args(arity);`

developedby (changes requested):

src/term/transform/scott_generation.rs
```diff
@@ -0,0 +1,48 @@
...
+
+        let lam = make_lam(ctr_args, ctrs, ctr_name);
+
+        let def_id = self.def_names.insert(ctr_name.clone());
+
+        let rules = vec![Rule { def_id, pats: vec![], body: lam }];
+        let def = Definition { def_id, rules };
+
+        self.defs.insert(def_id, def);
+        self.ctrs.remove(ctr_name);
```
developedby, line 20: Are you sure we wont' use this for other things kater? Seems unwise to remove it here

src/term/transform/scott_generation.rs
developedby: Actually, if the user pulls a funny move and declares something like this `data = (Ctr Ctr)` it would break, so we can leave it like this

developedby (changes requested):

src/term/transform/scott_generation.rs
imaqtkatt: I was probably testing something and I forgot to remove this line, my fault

developedby (approved):

PR #27, opened 2023-11-08
[sc-123] Add ref-to-ref simplification pass

developedby:


src/term/transform/simplify_ref_to_ref.rs
```diff
@@ -0,0 +1,95 @@
...
+    for def_id in self.defs.keys().copied().collect::<Vec<_>>() {
+      let body = &mut self.defs.get_mut(&def_id).unwrap().rules[0].body;
+      // Moving in and out so the borrow checker doesn't complain
+      let mut subst_body = std::mem::replace(body, Term::Era);
+      subst_ref_to_ref(&mut subst_body, &ref_map);
+      let body = &mut self.defs.get_mut(&def_id).unwrap().rules[0].body;
+      *body = subst_body;
+    }
+
+    // Remove the now unused defs
```
LunaAmora, line 46: This comment here is a TODO? the prune takes care of this, right?

src/lib.rs
```diff
@@ -43,7 +43,8 @@ pub fn run_compiled(book: &hvmc::ast::Book, mem_size: usize) -> (Net, RunStats)
   let start_time = Instant::now();
   root.normal(&runtime_book);
   let elapsed = start_time.elapsed().as_secs_f64();
-  let rewrites = Rewrites { anni: root.anni, comm: root.comm, eras: root.eras, dref: root.dref, oper: root.oper };
+  let rewrites =
+    Rewrites { anni: root.anni, comm: root.comm, eras: root.eras, dref: root.dref, oper: root.oper };
   let net = net_from_runtime(&root);
   //eprintln!("{}", show_net(&net));
```
imaqtkatt, line 49: Remove this commented line

src/term/transform/simplify_ref_to_ref.rs
developedby: I was going to do it here as a small optimization since we know these defs can be safely remove. In the end I decided to just leave it to pruning since it would inrease complexity and have almost no difference.

developedby: Forgot to remove the comment

src/lib.rs
developedby: This is some debug that I forgot and taelin commented out instead of removing

LunaAmora (approved):

PR #33, opened 2023-11-13
Add eta reduction pass

LunaAmora:


src/lib.rs
```diff
@@ -31,6 +31,7 @@ pub fn compile_book(book: &mut Book) -> Result<(hvmc::ast::Book, HashMap<Val, De
   book.make_var_names_unique();
   book.linearize_vars();
   book.detach_supercombinators();
+  book.eta_reduction();
```
developedby, line 34: Isn't it better to do eta reduction before separating the supercombinators? So we potentailly have fewer things to hoist.

developedby (approved):

PR #39, opened 2023-11-16
[sc-243] Hvm-lang bug with numeric operations

imaqtkatt:


src/term/net_to_term.rs
```diff
@@ -131,28 +131,31 @@ pub fn net_to_term_non_linear(net: &INet, book: &Book) -> (Term, bool) {
...
-                  valid,
-                )
+                  }
+                }
               }
+              Term::Opx { op, fst, snd } => match &*snd {
+                Term::Era => Term::Opx { op, fst, snd: Box::new(arg_term) },
+                _ => go(arg_term, Term::Opx { op, fst, snd }),
+              },
+              other => go(arg_term, other),
```
developedby, line 155: Add a comment explaining that this is actually an OP1 and we're flipping port 0 and 1 to undo the OP2 ~ NUM interaction.

src/term/net_to_term.rs
```diff
@@ -131,28 +131,31 @@ pub fn net_to_term_non_linear(net: &INet, book: &Book) -> (Term, bool) {
...
                     fst: Box::new(arg_term),
                     snd: Box::new(Term::Era),
-                  },
-                  valid,
-                )
+                  }
+                }
               }
+              Term::Opx { op, fst, snd } => match &*snd {
+                Term::Era => Term::Opx { op, fst, snd: Box::new(arg_term) },
```
developedby, line 152: Add a comment explaining that an Era means this came from the readback of the first OP2 node while anything else means a partially applied chain of OP2 nodes

developedby (changes requested):
Please add some tests

developedby (approved):

PR #40, opened 2023-11-17
[sc-156] Add explicit DUP/SUP labels to HVM-lang

imaqtkatt:


developedby (changes requested):
Please update the readme as well

tests/golden_tests/compile_file/tagged_dup.hvm
```diff
@@ -0,0 +1,5 @@
+main =
+  dup #i a b = @x x;
```
developedby, line 2: If I use the same tag across different functions, will they all generate the same dup label in the end?
And if I use separate tags in different functions, will they generate different dup labels?

src/term/term_to_net.rs
```diff
@@ -159,15 +169,21 @@ fn encode_term(
...
     // - 2: points to the occurrence of the second variable.
     // core: & val ~ {lab fst snd} (val not necessarily main port)
-    Term::Dup { fst, snd, val, nxt } => {
-      let dup = inet.new_node(Dup { lab: *dups });
-      *dups += 1;
-      let val = encode_term(inet, val, Port(dup, 0), scope, vars, global_vars, dups);
+    Term::Dup { fst, snd, val, nxt, tag } => {
+      let dup: u32;
+      if let Some(tag) = tag {
+        dup = tagged_dup(tag_storage, tag.clone(), inet);
```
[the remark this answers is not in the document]
imaqtkatt: In the moment that I use a tagged dup `dup #t a b = 0; ...`, every occurrence of `#t` in the program will have the same u8 value

tests/golden_tests/compile_file/tagged_dup.hvm
imaqtkatt: In the moment that I use a tagged dup dup #t a b = 0; ..., every occurrence of #t in the program will have the same u8 value

src/term/term_to_net.rs
```diff
@@ -159,15 +169,21 @@ fn encode_term(
     // - 1: points to the occurrence of the first variable.
     // - 2: points to the occurrence of the second variable.
     // core: & val ~ {lab fst snd} (val not necessarily main port)
-    Term::Dup { fst, snd, val, nxt } => {
-      let dup = inet.new_node(Dup { lab: *dups });
-      *dups += 1;
-      let val = encode_term(inet, val, Port(dup, 0), scope, vars, global_vars, dups);
+    Term::Dup { fst, snd, val, nxt, tag } => {
+      let dup: u32;
+      if let Some(tag) = tag {
```
developedby, line 174: I think it would be better to have this part only return the label.
If explicitly tagged and new tag, give a new label.
If explicitly tagged and repeated tag, use the label we decided on before.
If implicitly tagged, use the implicit label counter.

Then we create the dup node and insert it into the inet

src/term/term_to_net.rs
```diff
@@ -74,8 +74,17 @@ pub fn term_to_compat_net(term: &Term) -> (INet, u8) {
   // Encodes the main term.
   let mut global_vars = HashMap::new();
   let mut dups = 0;
-  let main =
-    encode_term(&mut inet, term, ROOT, &mut HashMap::new(), &mut vec![], &mut global_vars, &mut dups);
+  let mut tag_storage = HashMap::new();
```
developedby, line 77: What do you think about storing the used dup tags in the Book during parsing?
I don't know if it wouldn't make it hard to compile individual terms.

developedby (changes requested):

src/term/term_to_net.rs
```diff
@@ -100,6 +109,7 @@ fn encode_term(
   vars: &mut Vec<(Port, Option<Port>)>,
   global_vars: &mut HashMap<Name, (Port, Port)>,
   dups: &mut u8,
```
developedby, line 111: Could you join `dups` and `tag_storage` into a single "dup label generator" object?
Then you could have something like `LabelGenerator::generate(&mut self, tag: &Option<Name>)` and one argument fewer here.

This function is growing a lot in number of state variables and it becomes a bit hard to keep track of.

src/term/term_to_net.rs
developedby: If you do, please rename `dups` to something more explicit like `untagged_dups`, `num_untagged_dups` or `untagged_dup_counter`.

developedby (changes requested):

PR #51, opened 2023-11-24
Fix linearization bug with unused duplicated vars

LunaAmora:
this code was compiling wrong:

```
main = @x @y let b = (+ y y); x
```

resulting in `(a ({2 b c} a))` instead of `(a (* a))`


src/term/transform/linearize.rs
```diff
@@ -86,8 +87,14 @@ fn term_to_affine(term: &mut Term, var_uses: &mut HashMap<Name, Val>, let_bodies
     // Var-declaring terms
     Term::Lam { nam, bod } => {
       if let Some(nam_some) = nam {
-        if let Some(uses) = var_uses.get(nam_some).copied() {
+        if let Some(mut uses) = var_uses.get(nam_some).copied() {
           term_to_affine(bod, var_uses, let_bodies);
+
+          if uses != 0 {
```
developedby, line 93: Can you add a comment explaining why this is necessary?

src/term/transform/linearize.rs
```diff
@@ -86,8 +87,14 @@ fn term_to_affine(term: &mut Term, var_uses: &mut HashMap<Name, Val>, let_bodies
...
+            let name = Name(format!("{nam_some}$used"));
```
developedby, line 94: Can you refactor this into a function or a lambda so that there's only one place deciding what the name of this generated variable is?

src/term/transform/linearize.rs
```diff
@@ -133,7 +152,9 @@ fn term_to_affine(term: &mut Term, var_uses: &mut HashMap<Name, Val>, let_bodies
       if let Some(subst) = let_bodies.remove(nam) {
         *term = subst.clone();
       } else {
+        let used_counter = format!("{nam}$used");
```
developedby, line 155: here as well

src/term/transform/linearize.rs
```diff
@@ -154,8 +175,14 @@ fn term_to_affine(term: &mut Term, var_uses: &mut HashMap<Name, Val>, let_bodies
 
         if let MatchNum::Succ(nam) = num {
           if let Some(nam_some) = nam {
-            if let Some(uses) = var_uses.get(nam_some).copied() {
+            if let Some(mut uses) = var_uses.get(nam_some).copied() {
               term_to_affine(term, var_uses, let_bodies);
+
+              if uses != 0 {
+                let name = Name(format!("{nam_some}$used"));
```
developedby, line 182: also here

developedby (changes requested):

imaqtkatt (approved):

developedby (approved):

PR #56, opened 2023-11-25
fix match arm linearization procedure

tjjfvi:
Was poking at this code and got nerd-sniped 🙃 

The majority of the logic for linearization was already there; the order of the lambda args just needed to be changed from `@free @match_vars` to `@match_vars @free`.

Drafted as #54 needs to be merged first.


LunaAmora:
Looks good to me. When #54 is merged we'll review the fix. Thanks for the contribution!

tjjfvi:
incorporated into #60

PR #59, opened 2023-11-27
[sc-280] Add superposition

imaqtkatt:


src/term/mod.rs
```diff
@@ -291,7 +292,7 @@ impl Term {
         val.to_string(def_names),
         nxt.to_string(def_names)
       ),
-      Term::Sup { fst, snd } => format!("{{{} {}}}", fst.to_string(def_names), snd.to_string(def_names)),
+      Term::Sup { fst, snd, .. } => format!("{{{} {}}}", fst.to_string(def_names), snd.to_string(def_names)),
```
developedby, line 295: We should probably readback and print the label of sups, since they can appear in the result (in non-lambda programs)

src/term/net_to_term.rs
```diff
@@ -109,7 +109,7 @@ pub fn net_to_term_non_linear(net: &INet, book: &Book) -> (Term, bool) {
             let (fst, fst_valid) = reader(net, fst, namegen, dup_scope, tup_scope, book);
             let (snd, snd_valid) = reader(net, snd, namegen, dup_scope, tup_scope, book);
             let valid = fst_valid && snd_valid;
-            (Term::Sup { fst: Box::new(fst), snd: Box::new(snd) }, valid)
+            (Term::Sup { tag: None, fst: Box::new(fst), snd: Box::new(snd) }, valid)
```
developedby, line 112: Readback of dup tag here. We can check that this label is one of the ones decided by the user and for automatically decided labels we can use the number as the name (since numbers are not valid user-given names).

src/term/net_to_term.rs
```diff
@@ -432,7 +432,7 @@ pub fn net_to_term_linear(net: &INet, book: &Book) -> (Term, bool) {
           let snd_port = net.enter_port(Port(node, 2));
           let (snd, snd_valid) = reader(net, snd_port, namegen, dup_scope, tup_scope, seen, book);
           let valid = fst_valid && snd_valid;
-          (Term::Sup { fst: Box::new(fst), snd: Box::new(snd) }, valid)
+          (Term::Sup { tag: None, fst: Box::new(fst), snd: Box::new(snd) }, valid)
```
developedby, line 435: readback here as well

src/term/mod.rs
```diff
@@ -106,6 +106,7 @@ pub enum Term {
     nxt: Box<Term>,
   },
   Sup {
+    tag: Option<Name>,
```
developedby, line 109: Rethinking this, we should probably make sup tags obligatory.
* If each dup and sup is given an unique label, the user-provided sups will never be annihilated.
* If dup and sup labels can overlap and are generated independently, we get behaviour that depends on other functions and that the user has no control. Depending on arbitrary decisions of the compiler, a sup may interact with dups that have equal or different labels, leading to different results.

So, to me the only sensible option is to force sups to have explicit labels

src/term/term_to_net.rs
```diff
@@ -316,22 +327,45 @@ impl Op {
...
+      match tagged.entry(tag.clone()) {
         Entry::Occupied(e) => *e.get(),
         Entry::Vacant(e) => *e.insert(next_lab),
       }
     } else {
-      let lab = self.untagged_dups;
-      self.untagged_dups += 1;
+      let untagged = self.get_untagged(label_for);
+      let lab = *untagged;
+      *untagged += 1;
```
developedby, line 353: This is what I meant. If dups and sups can share the same labels depending on the order they appear in the program, we get unpredictable runtime results.

developedby (changes requested):

developedby (changes requested):

developedby (approved):

PR #60, opened 2023-11-27
[sc-278] Add linearize_matches pass

LunaAmora:
[Shortcut Story #278: Add linearization of variables used in match expressions](https://app.shortcut.com/hocomp/story/278/add-linearization-of-variables-used-in-match-expressions)


tjjfvi (approved):

tests/snapshots/run_single_files__match_mult_linearization.hvm.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: tests/golden_tests.rs
+input_file: tests/golden_tests/run_single_files/match_mult_linearization.hvm
+---
+λa match a { 0: λb λc λd (+ (+ b c) d); +: λf λg λh (+ (+ (+ h e) f) g) }
```
developedby, line 5: This test result is incorrect, `e` is a free variable

tests/snapshots/run_single_files__linearize_match.hvm.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: tests/golden_tests.rs
+input_file: tests/golden_tests/run_single_files/linearize_match.hvm
+---
+λa match a { 0: λb b; +: λd (+ d c) }
```
developedby, line 5: This test result is incorrect, `c` is a free variable.

tests/golden_tests/compile_file/scrutinee_reconstruction.hvm
```diff
@@ -0,0 +1,8 @@
+data Option = (Some val) | None
+
+Option.or = @a @b match a {
```
developedby, line 3: Could you add another test for reconstructing the matched variable, but in a scenario where both the variable itself and its argument are used? Something like
```
@a match a {
  Ctr: (foo a a.field)
}
```
The expected result should be that only the field is `dup`ed and `a` is reconstructed inside the match case body

developedby (changes requested):

tests/snapshots/run_single_files__match_mult_linearization.hvm.snap
LunaAmora: It's not incorrect, `e` is the variable `a-1` that got renamed by the unique_names pass. The succ arm contains the label information, but its not printed as it would be invalid syntax.

tests/snapshots/run_single_files__linearize_match.hvm.snap
LunaAmora: Same as the above

tests/snapshots/run_single_files__match_mult_linearization.hvm.snap
developedby: Could you fix the readback by renaming the variable to the correct name?

developedby (approved):

PR #61, opened 2023-11-28
Fix some tests, because of not being able to mem::take net.rdex

eduhenke:
Related to https://github.com/HigherOrderCO/hvm-core/pull/42

### DIFF
--- a/src/hvmc_net/pre_reduce.rs
+++ b/src/hvmc_net/pre_reduce.rs
@@ -33,8 +33,7 @@ pub fn pre_reduce_run_net(nam: &str, net: &mut hvmc::run::Net) -> Result<(), Str
 
   // Separate the derefing interactions, reduce all the others.
   // Also, always put the Ref on the left side (first element).
-  // The mem::replace thing implements a queue using 2 vecs.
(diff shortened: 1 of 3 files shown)


src/hvmc_net/pre_reduce.rs
```diff
@@ -33,8 +33,7 @@ pub fn pre_reduce_run_net(nam: &str, net: &mut hvmc::run::Net) -> Result<(), Str
 
   // Separate the derefing interactions, reduce all the others.
   // Also, always put the Ref on the left side (first element).
-  // The mem::replace thing implements a queue using 2 vecs.
-  let mut rdexes = std::mem::take(&mut net.rdex);
+  let mut rdexes: Vec<_> = net.rdex.drain(|n| n).map(Iterator::collect).unwrap_or(vec![]);
```
Boscop, line 36: It would be slightly more idiomatic to use `.unwrap_or_default()` :)
(Also below.)

developedby:
Not valid anymore after https://github.com/HigherOrderCO/hvm-lang/pull/194

PR #65, opened 2023-11-30
[sc-282] Add hvml syntax for binary and hexadecimal numbers

imaqtkatt:


tests/golden_tests/compile_term/nums.hvm
```diff
@@ -0,0 +1 @@
+(+ 0xFFFF_FFFF 0b101)
```
developedby, line 1: Can you also add a test for wrong number syntax? Like using decimal numbers when specifying binary and using hexa when specifying decimal

LunaAmora (approved):

src/term/parser/lexer.rs
```diff
@@ -29,6 +29,8 @@ pub enum Token {
   Equals,
 
   #[regex("[0-9]+", |lex| lex.slice().parse().ok())]
```
developedby, line 31: Could you change this to be like the other two?

developedby (changes requested):

developedby (approved):

PR #74, opened 2023-12-04
[sc-272] Add matching of pairs to pattern matching functions

imaqtkatt:
I'll open another pr to add the number pattern matching, I don't know how to do it yet

### DIFF
--- a/src/term/check/type_check.rs
+++ b/src/term/check/type_check.rs
@@ -6,6 +6,7 @@ use std::collections::HashMap;
 pub enum Type {
   Any,
   Tup,
+  Num,
   Adt(Name),
 }
 
@@ -53,9 +54,9 @@ pub fn infer_arg_type<'a>(
         }
       }
       Pattern::Tup(..) => Type::Tup,
-      Pattern::Num(..) => todo!(),
+      Pattern::Num(..) => Type::Num,
     };
-    unify(pat_type, &mut arg_type)?;
+    unify(pat_type, &mut arg_type)?
   }
   Ok(arg_type)
 }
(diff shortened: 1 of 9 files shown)


LunaAmora (approved):

developedby (changes requested):

src/term/transform/encode_pattern_matching.rs
```diff
@@ -286,7 +302,11 @@ fn get_pat_arg_count(match_path: &[Pattern]) -> (usize, usize) {
     Pattern::Var(_) => 1,
     Pattern::Ctr(_, vars) => vars.len(),
     Pattern::Num(_) => todo!(),
-    Pattern::Tup(..) => todo!(),
+    Pattern::Tup(fst, snd) => match (fst, snd) {
+      (None, None) => 0,
```
developedby, line 306: With the way the encoding is implemented right now, we always pass the unused arguments forward, so tuples should always count as 2 variables. We can improve this later to insert some Era terms at the end, but for now fix this block

src/term/check/type_check.rs
```diff
@@ -66,6 +67,8 @@ fn unify(new: Type, old: &mut Type) -> Result<(), String> {
     (Type::Adt(new), Type::Adt(old)) if &new != old => {
       return Err(format!("Type mismatch. Found '{}' expected {}.", new, old));
     }
+    (Type::Tup, Type::Any) => *old = Type::Tup,
```
developedby, line 70: Can you make it slightly simpler by changing these cases to just `(new, Type::Any) => *old = new`?

src/term/transform/encode_pattern_matching.rs
developedby: Also, we should have some tests for compiling unused pattern variables

developedby (changes requested):

developedby (approved):

PR #76, opened 2023-12-04
adt readback

tjjfvi:
(Based on #75)

Changes adt encoding from `@some @none (some x)` to `@#Option some @#Option none (some x)`, and adds a `resugar_adts` pass to `net_to_term` to readback ADTs and ADT matches.


tests/golden_tests.rs
```diff
@@ -102,6 +102,7 @@ fn compile_file() {
 #[test]
 fn run_single_files() {
   run_golden_test_dir(function_name!(), &|code| {
+    dbg!(code);
```
developedby, line 105: Is this leftover?

developedby (changes requested):

tests/golden_tests.rs
tjjfvi: Whoops, yes; removed.

developedby (approved):

PR #77, opened 2023-12-04
WIP let record destruction

LunaAmora:


src/term/mod.rs
```diff
@@ -131,6 +131,56 @@ pub enum Pattern {
...
+        ret
+      }
+      Pattern::Num(..) => false,
+      Pattern::Tup(fst, snd) => {
+        fst.as_ref().map_or(false, |fst| fst == name) || snd.as_ref().map_or(false, |snd| snd == name)
+      }
+    }
+  }
+
+  pub fn names(&self) -> Vec<&Option<Name>> {
```
developedby, line 153: Why not make a function that returns an iterator over the names? Then you don't need for_each_bind, you can just use `pat.names().for_each()`. To get them into a Vec you can just collect.

src/term/transform/simplify_matches.rs
```diff
@@ -193,26 +261,38 @@ fn match_adt_app(
...
-      let Pattern::Var(Some(ctr)) = rule else { unreachable!() };
-      if ctr == ctr_name {
-        let pat = Pattern::Ctr(ctr_name.clone(), vec_name_to_pat(&scrutinee, args));
-        let adt_binds: HashSet<_> = args.iter().map(|n| binded(&scrutinee, n)).collect();
-        let term_free_vars = term.free_vars().into_keys().filter(|k| !adt_binds.contains(k));
-        free_vars.extend(term_free_vars);
-
-        let rule = Rule { pats: vec![pat], body: term.clone() };
-        rules.push(rule);
+    let pats = match rule {
```
developedby, line 275: I think it would be better as `let (adt_binds, pats) = match ...`

developedby (changes requested):
Please add some tests

developedby (approved):

PR #78, opened 2023-12-05
[sc-274] Allow nested patterns in let expressions

imaqtkatt:


developedby (commented):
Can we hold this until this is merged? https://github.com/HigherOrderCO/hvm-lang/pull/77

Then we can have nested let with adt destructors

src/term/display.rs
```diff
@@ -94,21 +96,14 @@ impl Tag {
...
-impl Display for Pattern {
+impl fmt::Display for Pattern {
   fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
     match self {
       Pattern::Var(None) => write!(f, "*"),
       Pattern::Var(Some(nam)) => write!(f, "{nam}"),
-      Pattern::Ctr(nam, pats) => {
-        write!(f, "({}{})", nam, DisplayJoin(|| pats.iter().map(|p| display!(" {p}")), ""))
-      }
+      Pattern::Ctr(nam, pats) => write!(f, "({}{})", nam, pats.iter().map(|p| format!(" {p}")).join("")),
```
developedby, line 104: We have DisplayJoin for displaying joins of iterators, which avoids copying everything

src/term/transform/simplify_lets.rs
```diff
@@ -0,0 +1,68 @@
+use std::collections::BTreeMap;
+
+use crate::term::{Book, DefId, DefNames, Definition, Name, Pattern, Rule, Tag, Term};
+
+impl Term {
+  pub fn simplify_let(
```
developedby, line 6: Instead of converting `let`s directly to definitions, couldn't we convert them to `match` expressions which then get converted to function definitions? I don't see a reason why tuples should be treated differently from ADTs, and using the match machinery will also take care of linearizing the matched variable.
let to match conversion is currently done for ADT patterns in the simplify_matches file.

tests/golden_tests/run_single_files/def_tups.hvm
developedby, line 1: Could you also add these two files also to the desugaring tests? That way we can visualize the generated structure and it's easier for us to check if they are as expected

developedby (approved):

PR #79, opened 2023-12-06
[sc-287] Improve prereduction pass

developedby:


src/lib.rs
```diff
@@ -133,7 +148,9 @@ pub fn desugar_book(book: &mut Book) -> Result<DefId, String> {
   book.check_unbound_vars()?;
   book.make_var_names_unique();
   book.linearize_vars();
-  book.eta_reduction();
+  if opt_level >= OptimizationLevel::Heavy {
```
LunaAmora, line 151: I remember a case where a recursive rule with native number matching was not working properly without the eta_reduction active. I don't know if it was a bug that got fixed or the issue still persists. So, something to keep in mind if that happens again.

PR #80, opened 2023-12-07
[sc-272] Add matching of numbers

imaqtkatt:


src/term/transform/encode_pattern_matching.rs
```diff
@@ -180,6 +193,61 @@ fn make_leaf_pattern_matching_case(
...
+  use MatchNum::*;
+
+  fn filter_rules(def_rules: &[Rule], crnt_rules: &[usize], arg_idx: usize, Name(ctr): &Name) -> Vec<usize> {
+    crnt_rules
+      .iter()
+      .copied()
+      .filter(|&rule_idx| match &def_rules[rule_idx].pats[arg_idx] {
+        Pattern::Var(_) => false,
+        Pattern::Ctr(_nam, _) => false,
+        Pattern::Num(Succ(_)) => ctr == "+",
```
developedby, line 213: Could you order these?

developedby, line 213: Is there any case where the constructor name will not be `"+"`? If not, we can just have this return true always

src/term/transform/encode_pattern_matching.rs
```diff
@@ -180,6 +193,61 @@ fn make_leaf_pattern_matching_case(
...
+  }
+
+  let make_next_fn_name = |crnt_name, ctr_name: &Name| Name(format!("{crnt_name}$P{ctr_name}"));
+
+  for (next_ctr, next_ctr_args) in vec![(Name::new("0"), vec![]), (Name::new("+"), vec![Name::new("p")])] {
+    let def = &book.defs[&def_id];
+    let crnt_name = make_next_fn_name(crnt_name, &next_ctr);
+    let crnt_rules = filter_rules(&def.rules, &crnt_rules, match_path.len(), &next_ctr);
+    let new_vars =
+      Pattern::Ctr(next_ctr.clone(), vec![Pattern::Var(Some(Name::new(""))); next_ctr_args.len()]);
```
developedby, line 227: It should be `Pattern::Num` and create the correct number pattern

src/term/transform/encode_pattern_matching.rs
```diff
@@ -180,6 +193,61 @@ fn make_leaf_pattern_matching_case(
...
+
+  let term = Term::Var { nam: Name::new("x").clone() };
+  let zero_id = book.def_names.def_id(&make_next_fn_name(crnt_name, &Name::new("0"))).unwrap();
+  let succ_id = book.def_names.def_id(&make_next_fn_name(crnt_name, &Name::new("+"))).unwrap();
+
+  let term = Term::Match {
+    scrutinee: Box::new(term),
+    arms: vec![
+      (Pattern::Num(MatchNum::Zero), Term::Ref { def_id: zero_id }),
+      (Pattern::Num(MatchNum::Zero), Term::Ref { def_id: succ_id }),
```
developedby, line 241: MatchNum::Succ

src/term/transform/encode_pattern_matching.rs
LunaAmora: Isn't it better to leave the check in case it is different somehow, to help catch errors when maintaining the code?

src/term/transform/encode_pattern_matching.rs
developedby: The pattern for the second is not Zero, otherwise youre creating this
```
match x {
 0: Foo$P0
 0: Foo$P+
}
```

src/term/transform/encode_pattern_matching.rs
developedby: It won't catch the error, it will misteriously generate incorrect compilations because this expression will return `false`.

These names are created 10 lines below
`for (next_ctr, next_ctr_args) in vec![(Name::new("0"), vec![]), (Name::new("+"), vec![Name::new("p")])] {`, and they're not really relevant for this function. They're only used to generate the correct name for the next subfunctions.

LunaAmora: I see

src/term/transform/encode_pattern_matching.rs
developedby: I see, but this looks rather wrong, so maybe we should find a better way of doing this in a future PR.

LunaAmora (approved):

developedby (changes requested):
I think you missed this message:
> Please add some more tests.
> One using the matched variable from the Succ case, one with non exhaustive number patterns, one with non-exhaustive ctr pattern but with a number pattern first, and any others you can think of.

developedby (approved):

PR #82, opened 2023-12-08
[sc 296] Update hvml's readme with all the new pattern matching info

LunaAmora:


README.md
```diff
@@ -46,6 +46,10 @@ hvml compile <file>
 ```
 This will output the compiled file to stdout.
 
+## Machine Integer
```
developedby, line 49: It is a bit strange to have this section right after how to use and right before syntax. It's worded as a limitation of the language and generally we don't want to advertise in bold letters how limited we are.

I'd say this would be better placed  2 paragraphs below, when you're defining the syntax of a number.
"Here we define a function `two` as returning the number 2:
...
Currently, the only supported type of machine numbers are unsigned 60-bit integers.
..."

Also, it would be nice to have a paragraph explaining binary and hexa number syntax and mention the possibility of having digit separators (123_456 vs 0b101011_11010 vs 0x123_abc).

README.md
```diff
@@ -106,12 +110,62 @@ This term will reduce to:
 
 A match syntax for machine numbers.
 We match the case 0 and the case where the number is greater
-than 0, p binds the value of the matching number - 1:
+than 0, `n-1` binds the value of the matching number - 1:
```
developedby, line 113: Mention that n is the name of the matched variable.

We match ... greater than 0. If `n` is the matched variable, `n-1` binds the value..."

README.md
```diff
@@ -106,12 +110,62 @@ This term will reduce to:
...
+    +: (f (Number.to_church n-1 f x))
+  }
+```
+
+It is possible to define Data types using `data`:
+```rs
+data Option = (Some val) | None
+```
+
+If the data type has a single possible value, it can be destructured using `let`:
```
developedby, line 127: single possible constructor

README.md
```diff
@@ -106,12 +110,62 @@ This term will reduce to:
...
+Number.to_church = λn λf λx 
+  match n {
+    0: x
+    +: (f (Number.to_church n-1 f x))
+  }
+```
+
+It is possible to define Data types using `data`:
+```rs
+data Option = (Some val) | None
```
developedby, line 124: Mention that if the constructor has arguments, we need to put parentheses.

README.md
```diff
@@ -106,12 +110,62 @@ This term will reduce to:
...
+let (Box value) = boxed; value
+```
+
+Otherwise, there are two pattern syntaxes for matching on data types:
+
+```rs
+// Implicit bindings
+Option.map = λoption λf
+  match option {
+    Some: (Some (f option.val))
```
developedby, line 140: Explain where the names of the implicit binds come from (matched variable . constructor name)

README.md
```diff
@@ -106,12 +110,62 @@ This term will reduce to:
...
+But with the extra ability to match on multiple values at once:
+
+```rs
+data Boolean = True | False
+
+(Option.is_both_some (Some lft_val) (Some rgt_val)) = True
+(Option.is_both_some lft rgt) = False
 ```
 
 ## Terms to Nodes
```
developedby, line 171: `Compilation of Terms to hvm-core` or something similar would be more clear

tests/golden_tests/run_single_files/example.hvm
```diff
@@ -1,25 +1,82 @@
...
+// Use tuples to store two values together without needing to create a new data type
+(Tuple.new fst snd) =
+  let pair = (fst, snd)
+  pair
+
+// Then you can destructure it inside the definition or using `let`
+(Tuple.fst) = λpair
+  let (fst, snd) = pair
+  fst
+(Tuple.snd (fst, snd)) = snd
```
developedby, line 66: I would change the order and put the pattern matching function first, since it's slightly simpler to understand

developedby (changes requested):

developedby (approved):

PR #83, opened 2023-12-12
[sc 308] Update readme with information about tagged terms

LunaAmora:
https://github.com/HigherOrderCO/hvm-core/pull/48 should probably be merged first, as the sup example with numeric operations is currently bugged on hvm-core

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,19 +92,60 @@ And destructuring tuples with `let`:
 let (x, y) = tup; (+ x y)
 ```
 
-Term duplication is possible using `dup`:
+Term duplication is done automatically when a variable is used more than once. But it's possible to manually duplicate a term using `dup`:
 ```rs
 // the number 2 in church encoding using dup.
 ch2 = λf λx dup f1 f2 = f; (f1 (f2 x))
-
-/// a tagged '#i' dup
-id_id = dup #i id1 id2 = λx x; (id1 id2)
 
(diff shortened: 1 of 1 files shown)


imaqtkatt (approved):

README.md
```diff
@@ -92,19 +105,60 @@ And destructuring tuples with `let`:
...
+To avoid label collision, HVM-Lang automatically generates new dup labels for each dup in the code. But with cases like the example above, when the interacting dups comes from the same place, the result is an invalid reduction.  
+
+To fix the problem, its necessary to re-create the term so that a new label is assigned, or manually assign one:
+```rs
+c2  = λf λx dup        f1 f2 = f; (f1 (f2 x))
+c2_ = λf λx dup #label f1 f2 = f; (f1 (f2 x))
+main = (c2 c2_)
+```
+
+It is possible to use channels, the channel variable can occur outside it's body:
```
developedby, line 161: Can you add an explanation on what "channels" are (which I'm not sure if are the name we want to give these). Basically lambdas whose variable escapes their scope. Also, we should probably move this to the advanced section

developedby (changes requested):

README.md
LunaAmora: Should we just call then scopeless lambdas?

developedby (approved):

tjjfvi (approved):

PR #86, opened 2023-12-13
[sc-304] Refactor pattern matching code, change Num patterns

developedby:


tests/snapshots/readback_lnet__match.hvm.snap
```diff
@@ -2,4 +2,4 @@
 source: tests/golden_tests.rs
 input_file: tests/golden_tests/readback_lnet/match.hvm
 ---
-let b = 1; match b { 0: λa a; +: b-1 }
+let b = 1; match b { 0: λa a; +b-1: b-1 }
```
LunaAmora, line 5: Shouldn't this `b-1` be omitted as it is not a valid syntax?

developedby: Except for the invalid name, it is valid syntax. Do you think we should just leave it with a generic variable name?

src/term/transform/extract_matches.rs
```diff
@@ -0,0 +1,215 @@
...
+    for (def_id, def) in &mut self.defs {
+      let def_name = book.def_names.name(def_id).cloned().unwrap();
+      for rule in def.rules.iter_mut() {
+        rule.body.extract_matches(&def_name, book, &mut 0).map_err(|err| err.to_string())?;
+      }
+    }
+
+    if !book.new_defs.is_empty() {
+      self.defs.append(&mut book.new_defs);
+      self.flatten_rules();
```
LunaAmora, line 24: this can be removed if flatten_rules are used after the `extract_matches` pass (as it is now in the order on the `encode_pattern_matching`). 

We can keep in case that order could change, but adding a post-condition documentation that flatten_rules should be used after this could be enough.

developedby: Didn't even notice this line here. I think it's better to remove it and have all the passes be sort of complete in themselves and not call other passes, since that increases the complexity of the code.

src/term/transform/extract_matches.rs
```diff
@@ -0,0 +1,215 @@
...
+/// Transforms a match into a new definition with every arm of `arms` as a rule.
+/// The result is the new def applied to the scrutinee followed by the free vars of the arms.
+fn match_to_def(
+  scrutinee: Name,
+  arms: &[(Pattern, Term)],
+  def_name: &Name,
+  book: &mut MatchesBook,
+  match_count: &mut usize,
+) -> Result<Term, MatchError> {
+  let current_count = *match_count;
```
LunaAmora, line 176: We can remove this now, the function does not need to receive a `&mut usize` anymore but was left here in an earlier refactor.

LunaAmora: The function does not need to return a result anymore too

tests/snapshots/compile_file__linearize_match.hvm.snap
LunaAmora, line 1: Should main be extracted like this and being just a ref to another rule?

developedby: If main is just a match expression, it will be extracted. This is a problem of the pass that removes functions that are just refs to other functions, it doesn't work on `main`.
But that's outside the scope of this PR.

tests/snapshots/readback_lnet__match.hvm.snap
LunaAmora: What do you mean it is valid syntax? The `+` arm of a native match can only be `+: body`, if you put any name there the parser will give an error

LunaAmora: What i mean is that: Shouldn't the readback contain only valid syntax that just by copying it and pasting into a file it should be parseable without problems?

LunaAmora (approved):

tests/snapshots/readback_lnet__match.hvm.snap
developedby: It should accept both versions, like we do for adt constructors. If right now it doesn't then we should fix it

imaqtkatt (approved):

LunaAmora (approved):

LunaAmora (changes requested):
Some tests are broken now after the parsing change

developedby:
> Some tests are failing now after the parsing change

I messed up on the commit and didn't add some of the changes.

LunaAmora (approved):

PR #87, opened 2023-12-13
Fix tuple defs applications

imaqtkatt:
The desugar for String.concat was generating a wrong pattern definition:
```
(String.concat$P$P) = λy0 λx0 (String.concat$R0 x0 y0 x0 y0)
```

### DIFF
--- a/src/term/transform/encode_pattern_matching.rs
+++ b/src/term/transform/encode_pattern_matching.rs
@@ -167,6 +167,7 @@ fn make_leaf_pattern_matching_case(
         let ctr_term = ctr_args.fold(Term::Ref { def_id: ctr_ref_id }, Term::arg_call);
         Term::App { tag: Tag::Static, fun: Box::new(term), arg: Box::new(ctr_term) }
       }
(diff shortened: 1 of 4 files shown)


src/term/transform/encode_pattern_matching.rs
```diff
@@ -167,7 +167,7 @@ fn make_leaf_pattern_matching_case(
         let ctr_term = ctr_args.fold(Term::Ref { def_id: ctr_ref_id }, Term::arg_call);
         Term::App { tag: Tag::Static, fun: Box::new(term), arg: Box::new(ctr_term) }
       }
-      (_, Pattern::Tup(..)) => arg_use.clone().fold(term, Term::arg_call),
+      (_, Pattern::Tup(..)) => Term::arg_call(term, arg_use.next().unwrap()),
```
developedby, line 170: This is because pairs are only destructed at the leafs, right? Can you add a comment explaining this?

developedby (changes requested):

developedby (approved):

PR #88, opened 2023-12-14
Added brackets to scopeless lambda term

N0lim:


developedby:
It was purposefully written like that to show how hvm-lang syntax works, but I will improve the explanation for the application syntax to avoid further confusion.

PR #89, opened 2023-12-14
[sc-307] Add string syntax to hvml

imaqtkatt:


src/term/net_to_term.rs
```diff
@@ -207,6 +210,25 @@ impl<'a> Reader<'a> {
   }
 }
 
+fn decode_str(fst: Term, bod: Term) -> Term {
```
tjjfvi, line 213: These `unreachable!`s should instead push readback errors (as many of them are likely reachable in debug mode)

src/term/parser/lexer.rs
```diff
@@ -33,6 +33,14 @@ pub enum Token {
   #[regex("0[bB][0-1_]+", |lex| from_radix(2, lex).ok())]
   Num(u64),
 
+  #[regex(r#""([^"\\]|\\t|\\u|\\n|\\")*""#, |lex| {
+    let mut s = lex.slice().to_string();
+    s.pop();
+    s.remove(0);
+    s
```
tjjfvi, line 40: The escapes allowed in the regex are not normalized here

src/term/transform/encode_strs.rs
```diff
@@ -0,0 +1,61 @@
...
+        arg.encode_str();
+      }
+      Term::Tup { fst, snd } | Term::Sup { fst, snd, .. } => {
+        fst.encode_str();
+        snd.encode_str();
+      }
+      Term::Opx { fst, snd, .. } => {
+        fst.encode_str();
+        snd.encode_str();
+      }
```
tjjfvi, line 52: These branches can be merged

src/term/display.rs
```diff
@@ -67,6 +67,7 @@ impl Term {
       }
       Term::Era => write!(f, "*"),
       Term::Num { val } => write!(f, "{val}"),
+      Term::Str { val } => write!(f, "{val}"),
```
tjjfvi, line 70:
```suggestion
      Term::Str { val } => write!(f, "{val:?}"),
```
(to show it with quotes/escapes)

src/term/parser/lexer.rs
```diff
@@ -33,6 +33,12 @@ pub enum Token {
   #[regex("0[bB][0-1_]+", |lex| from_radix(2, lex).ok())]
   Num(u64),
 
+  #[regex(r#""([^"\\]|\\t|\\u|\\n|\\")*""#, |lex| normalized_string(lex).ok())]
+  Str(String),
+
+  #[regex(r#"'[\W|\w]'"#, num_of_char)]
```
tjjfvi, line 39:
```suggestion
  #[regex(r#"'.'"#, num_of_char)]
```

tjjfvi: Although perhaps this should have the same escape handling as strings

tjjfvi (approved):

developedby (approved):

PR #94, opened 2024-01-09
Improve documentation and write some tutorials.

FranchuFranchu:
Currently, documentation for `hvm-lang` is very poor. The only documentation is the `README.md` file which is too large and contains a confusing mixture of explanations and examples.

This PR splits the documentation part of `README.md` into many short files, each of which teaches a single `hvm-lang` concept. We can later make them longer if necessary.


README.md
```diff
@@ -115,225 +98,18 @@ And destructuring tuples with `let`:
...
-
-Also, nodes have labels, we use the label to store data in the node's memory and also differentiate them.
-
-The `Number` node uses the label to store it's number.
-An `Op2` node uses the label to store it's operation.
-And a `Constructor` node can have a label too! The label is used for `Dup` and `Tuple` nodes.
+### Advanced features
 
-Check [HVM-Core](https://github.com/HigherOrderCO/hvm-core/tree/main#language) to know more about.
+Advanced features are described in other files:
```
developedby, line 103: It's a bit strange to call numbers and data types advanced features. Maybe "other features" or similar.

README.md
```diff
@@ -115,225 +98,18 @@ And destructuring tuples with `let`:
...
 
-### Runtime and Compiler optimizations
+- Duplications and superpositions: [Dups and sups](docs/dups-and-sups.md)
+- Data types: [Defining data types](docs/defining-data-types.md)
+- Native numbers and operations: [Native numbers](docs/native-numbers.md)
+- Scopeless lambdas: [Using scopeless lambdas](docs/using-scopeless-lambdas.md)
+- Tagged lambdas and applications: [Automatic vectorization with tagged lambdas](docs/automatic-vectorization-with-tagged-lambdas.md)
+- Lazy definitions: [Making recursive definitions lazy](docs/lazy-definitions.md)
```
developedby, line 110: We should probably order this by complexity/"advancedness".

Data types, numbers and lazy definitions should come first, since they are important for writing even normal lambda programs.

examples/callcc.hvm
```diff
@@ -0,0 +1,10 @@
+// TODO: Doesn't yet work because of a bug in hvm-lang.
```
developedby, line 1: What is the bug that prevents this program from working?

developedby (changes requested):

examples/callcc.hvm
FranchuFranchu: The code `main = let * = λ$x *; $x` triggers a `thread 'main' panicked at 'internal error: entered unreachable code', src/net/net_to_hvmc.rs:77:22` panic. The callcc code is an example of this bug because it erases `$garbage` with a let statement.

LunaAmora: Should the let body not be pruned out when a channel is defined inside of it?

developedby: Yes, the let shouldn't be removed in this case.

FranchuFranchu: @developedby I figured out a workaround that makes the program work in the current version of HVM-lang. 

By adding a `Seq a b = a` rule, and passing `$garbage` as the second argument and `result` as the first one, it's possible to avoid the `let` body being deleted while at the same time discarding `$garbage`

developedby (approved):

developedby:
#99 and #85 have some fixes on the documentation, make sure to properly merge them into this pr

FranchuFranchu:
> #99 and #85 have some fixes on the documentation, make sure to properly merge them into this pr

Alright, that's done by the last two commits

PR #95, opened 2024-01-09
[sc-305] Simplify the generated encoding for pattern matching

LunaAmora:
This pr deals with the this part of the task:

- [x] If we know an argument will not be used, we can not pass it forward, which reduces the number of generated nodes.

Making the current program and compilation of:
```rs
data MyType = (A a) | (B b) | (C c) | (D d1 d2) | (E e1 e2)

Foo (A a) = R0
Foo * = R1
```
```rs
Foo = @x (x Foo$A Foo$B Foo$C Foo$D Foo$E)
Foo$R0 = @a R0
Foo$R1 = @* R1
Foo$A = @a (Foo$R0 a)
Foo$B = @b (Foo$R1 (B b))
Foo$C = @c (Foo$R1 (C c))
Foo$D = @d1 @d2 (Foo$R1 (D d1 d2))
Foo$E = @e1 @e2 (Foo$R1 (E e1 e2))
```

into this:
```rs
Foo = @x (x Foo$A Foo$B Foo$C Foo$D Foo$E)
Foo$R0 = @a R0
Foo$R1 = @* R1
Foo$A = @a (Foo$R0 *)
Foo$B = @b (Foo$R1 *)
Foo$C = @c (Foo$R1 *)
Foo$D = @d1 @d2 (Foo$R1 *)
Foo$E = @e1 @e2 (Foo$R1 *)
```

Instead of merging the now identical definitions of `Foo$B & Foo$C` and `Foo$D & Foo$E`, this can be done in a later pass to merge all identical definitions without further complicating the pattern matching encoding pass.

### DIFF
--- a/src/term/transform/encode_pattern_matching.rs
+++ b/src/term/transform/encode_pattern_matching.rs
@@ -146,8 +146,8 @@ fn make_leaf_pattern_matching_case(
   // The term we're building
   let term = Term::Ref { def_id: rule_def_id };
 
-  // The lambdas  of the rule we're calling
-  let lambdas_usage = &mut book.defs[&rule_def_id].rules[0].body.lambdas_usage().into_iter();
+  // Whether the args of the rule we're calling are used or discarded
+  let lambdas_usage = &mut book.defs[&rule_def_id].rules[0].body.arg_vars_are_used().into_iter();
 
   let (num_new_args, num_old_args) = get_pat_arg_count(&match_path);
   let old_args = (0 .. num_old_args).map(|x| Name(format!("x{x}")));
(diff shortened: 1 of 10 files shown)


src/term/transform/encode_pattern_matching.rs
```diff
@@ -354,6 +375,39 @@ impl Term {
...
     Term::App { tag: Tag::Static, fun: Box::new(term), arg: Box::new(Term::Var { nam: arg }) }
   }
+
+  fn optional_arg_call(term: Term, arg: Option<Name>) -> Term {
+    let arg = Box::new(arg.map_or(Term::Era, |nam| Term::Var { nam }));
+    Term::App { tag: Tag::Static, fun: Box::new(term), arg }
+  }
+
+  // Returns a sequence representing whether the lambda arg is binded or ignored
+  fn lambdas_usage(&self) -> Vec<Option<()>> {
```
developedby, line 385: Can't this just be a bool?

developedby, line 385: This function name is a bit hard to understand. Something like `arg_vars_are_used` maybe? Not sure, this one is a bit hard.

src/term/transform/encode_pattern_matching.rs
```diff
@@ -354,6 +375,39 @@ impl Term {
   fn arg_call(term: Term, arg: Name) -> Term {
     Term::App { tag: Tag::Static, fun: Box::new(term), arg: Box::new(Term::Var { nam: arg }) }
   }
+
+  fn optional_arg_call(term: Term, arg: Option<Name>) -> Term {
+    let arg = Box::new(arg.map_or(Term::Era, |nam| Term::Var { nam }));
+    Term::App { tag: Tag::Static, fun: Box::new(term), arg }
+  }
+
+  // Returns a sequence representing whether the lambda arg is binded or ignored
```
developedby, line 384: Can you add an example? Like `@a @* (a @x x) -> vec![true, false]`.

developedby (changes requested):

src/term/transform/encode_pattern_matching.rs
```diff
@@ -146,40 +146,61 @@ fn make_leaf_pattern_matching_case(
   // The term we're building
   let term = Term::Ref { def_id: rule_def_id };
 
+  // The lambdas  of the rule we're calling
```
developedby, line 149: This comment is not very descriptive.
I think "Whether the args of the rule we're calling are used" is a bit better.

developedby (changes requested):

imaqtkatt (approved):

developedby (approved):

PR #100, opened 2024-01-10
[sc-311] fix ref to ref removal pass

LunaAmora:
This pr implements the first option described in the task:

1. One would be to have main be a duplicate of the function it's referencing. This is the simplest option.
2. Another option would be to have the other function be merged with main, replacing all calls to it with calls to main.

### DIFF
--- a/src/term/transform/simplify_ref_to_ref.rs
+++ b/src/term/transform/simplify_ref_to_ref.rs
@@ -7,7 +7,7 @@ impl Book {
   // When we find a function that is simply directly calling another function,
   // substitutes all occurences of that function to the one being called, avoiding the unnecessary redirect.
   // In case there is a long chaing of ref-to-ref-to-ref, we substitute values by the last function in the chain.
-  pub fn simplify_ref_to_ref(&mut self, main: DefId) -> Result<(), String> {
+  pub fn simplify_ref_to_ref(&mut self) -> Result<(), String> {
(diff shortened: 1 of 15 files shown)


src/term/transform/simplify_ref_to_ref.rs
```diff
@@ -39,6 +39,12 @@ impl Book {
       *body = subst_body;
     }
 
+    if let Term::Ref { def_id } = &self.defs.get(&main).unwrap().rules[0].body {
```
developedby, line 42: Can you just add a comment explaining why we do this `if`?

developedby (changes requested):

developedby:
I think we can do without this pass at all. With compile-time reduction and later inlining, we will achieve the same results.

But this is not super high priority, so I'll come back later to decide what to do with this PR.

LunaAmora:
> I think we can do without this pass at all. With compile-time reduction and later inlining, we will achieve the same results.
> 
> But this is not super high priority, so I'll come back later to decide what to do with this PR.

As can be observed on this [str task](https://app.shortcut.com/hocomp/story/338/function-pattern-matching-fails-on-num-and-var), i don't think the other passes are achieving the desired results, as a main that is only a ref to another function causes hvm-core to have non-deterministic results. 

Or did you mean we could have this fixed when we improve the other passes and remove this one?

imaqtkatt (approved):

developedby (approved):

PR #103, opened 2024-01-10
[sc-327] Fix Num pattern matching encoding

LunaAmora:
- [x] Problem 1: Number match enconding miscompilation
```rs
(StrInc (len, buf)) = (len, @#str x (StrGo len (#str buf x)))

(StrGo 0 str) = str
(StrGo +x (head, tail)) = ((+ 1 head), (StrGo x tail))

main = (StrInc "hello world")
```

- [x] Problem 2: Readback of an eta-equivalent code was wrong
```rs
(StrInc (len, buf)) = (len, @#str x (StrGo len (#str buf x)))

(StrGo 0  (head, tail)) = (head, tail)
(StrGo +x (head, tail)) = ((+ 1 head), (StrGo x tail))

main = (StrInc "hello world")
```


src/term/transform/encode_pattern_matching.rs
```diff
@@ -475,7 +464,7 @@ fn get_pat_arg_count(match_path: &[Pattern]) -> (usize, usize) {
     Pattern::Ctr(_, vars) => vars.len(),
     Pattern::Num(MatchNum::Zero) => 0,
     Pattern::Num(MatchNum::Succ { .. }) => 1,
-    Pattern::Tup(..) => 2,
+    Pattern::Tup(..) => unreachable!(), // TODO: Why is this unreachable? Can this be refactored?
```
developedby, line 467: This function is called only when adding the fields of a destructured argument as a lambda. Since we don't do this for tuples, it never gets called, but if it did, the correct value would be indeed 2.
I think you can keep it `2` but add a comment that for tuples this isn't actually called because we only destructure them at the end.

developedby (changes requested):

imaqtkatt (approved):

src/term/net_to_term.rs
```diff
@@ -211,6 +214,42 @@ impl<'a> Reader<'a> {
...
+  /// (a a)
+  /// ```
+  ///
+  fn decay_or_get_ports(&mut self, node: NodeId) -> Result<Term, (Term, Term)> {
+    let fst_port = self.net.enter_port(Port(node, 1));
+    let snd_port = self.net.enter_port(Port(node, 2));
+
+    if fst_port.node() == snd_port.node() && fst_port.slot() == 1 && snd_port.slot() == 2 {
+      let first_node = fst_port.node();
+      if self.net.node(node).kind == self.net.node(first_node).kind {
```
developedby, line 243: This is not valid for all kinds of nodes, only CON/TUP/DUP, due to  their interaction rules.

src/term/net_to_term.rs
```diff
@@ -211,6 +214,42 @@ impl<'a> Reader<'a> {
 
     term
   }
+
+  /// Enters both ports 1 and 2 of a node  
+  /// If both ports are connected to the same node, and that node Kind is the same as the given node Kind,  
+  /// Reads the port 0 of the connected node, and returns that term,  
+  /// Otherwise, returns the terms on ports 1 and 2 of the given node.
```
developedby, line 221: Maybe add that these two inets (before and after the transformation) always have equivalent outcomes, so this is why we're able to do it.

developedby (changes requested):

developedby (approved):

PR #107, opened 2024-01-12
[sc-317] Support era pattern with nested constructor rules

imaqtkatt:


src/term/transform/flatten.rs
```diff
@@ -177,6 +177,7 @@ fn make_split_rule(old_rule: &Rule, other_rule: &Rule, def_names: &DefNames) ->
           new_pats.push(other_field.clone());
         }
       }
+      (Pattern::Ctr(..), Pattern::Var(None)) => new_pats.push(Pattern::Var(None)),
```
developedby, line 180: We need to also implement `(Pattern::Tup(..), Pattern::Var(None))`, `(Pattern::Num(..), Pattern::Var(None))`, etc, right?

And remove the unreachable on the bottom of the match

developedby: And add test examples of those other cases

developedby (changes requested):

src/term/transform/flatten.rs
```diff
@@ -189,10 +189,7 @@ fn make_split_rule(old_rule: &Rule, other_rule: &Rule, def_names: &DefNames) ->
         new_body.subst(other_arg, &new_ctr);
       }
       (Pattern::Num(..), Pattern::Num(..)) => new_pats.push(other_arg.clone()),
-      (Pattern::Num(..), Pattern::Var(..)) => {
-        // How to do this with this kind of number pattern? Subst with a match?
-        todo!();
-      }
+      (Pattern::Num(..), _) => unreachable!(),
```
developedby, line 192: All `(Pattern::Num(..), _) => unreachable!(),` should be unreachable I think, because there can't be patterns nested in num (it's always final, a number). You can remove the line above.

src/term/transform/flatten.rs
```diff
@@ -207,17 +204,19 @@ fn make_split_rule(old_rule: &Rule, other_rule: &Rule, def_names: &DefNames) ->
...
       // Unreachable cases, we only call this function if we know the two patterns match together
       (
-        Pattern::Ctr(..) | Pattern::Num(..) | Pattern::Tup(..),
+        Pattern::Ctr(..) | Pattern::Tup(..),
         Pattern::Ctr(..) | Pattern::Num(..) | Pattern::Tup(..),
       ) => {
         unreachable!()
       }
-      (_, Pattern::Var(None)) => unreachable!(),
+      (Pattern::Ctr(..), Pattern::Var(None)) => new_pats.push(Pattern::Var(None)),
```
developedby, line 215: Shouldn't you push one * pattern for each field that is being extracted from the nested pattern?
Try testing this:
```
data Tuple = (Pair a b)
(Foo (Pair (Pair a b) c)) = ...
(Foo *) = ...
```

developedby (changes requested):

main.hvm
developedby, line 1: Remove this

src/term/transform/flatten.rs
```diff
@@ -188,11 +188,7 @@ fn make_split_rule(old_rule: &Rule, other_rule: &Rule, def_names: &DefNames) ->
         let new_ctr = Term::call(Term::Ref { def_id: rule_arg_def_id }, new_ctr_args);
         new_body.subst(other_arg, &new_ctr);
       }
-      (Pattern::Num(..), Pattern::Num(..)) => new_pats.push(other_arg.clone()),
-      (Pattern::Num(..), Pattern::Var(..)) => {
-        // How to do this with this kind of number pattern? Subst with a match?
-        todo!();
-      }
+      (Pattern::Num(..), _) => unreachable!(),
```
developedby, line 191: Can you add a comment explaining that this is unreacheable because `Num` can't have nested patterns?

developedby (changes requested):

developedby (approved):

PR #108, opened 2024-01-12
[sc-333] Fix pattern matching encoding for functions with only var args

developedby:


tests/golden_tests/encode_pattern_match/var_only.hvm
```diff
@@ -0,0 +1,9 @@
+data Bool = False | True
+
+(Foo a b) = λf (f a)
+(Foo a b) = b
```
LunaAmora, line 4: Shouldn't this be an error or warning on the future? the second rule is completely ignored

LunaAmora (approved):

tests/golden_tests/encode_pattern_match/var_only.hvm
developedby: We probably want redundant rules to give a warning, as well as match expression with only variable patterns, but that is beyond the scope of this PR

imaqtkatt (approved):

PR #110, opened 2024-01-15
[sc-341] Read special characters in char

imaqtkatt:


src/term/parser/lexer.rs
```diff
@@ -34,7 +34,7 @@ pub enum Token {
   Str(String),
 
   #[regex(r#"'\\U[0-9a-fA-F]{1,8}'"#, normalized_char, priority = 2)]
-  #[regex(r#"'(.|\\t|\\u[0-9a-fA-F]{1,4}|\\n|\\')'"#, normalized_char)]
+  #[regex(r#"'(\p{L}|\p{M}|\p{N}|\p{P}|\p{S}|\p{Z}|\p{C}|\p{Emoji}|\\t|\\u[0-9a-fA-F]{1,4}|\\n|\\')'"#, normalized_char)]
```
developedby, line 37: Can you add a comment explaining that this covers every possible unicode character?.

Also, you could merge the different escape codes into a single regex rule for simplicity: `\\[tn']`. Then if we want to add more in the future it becomes easier.

developedby (changes requested):

developedby (approved):

PR #117, opened 2024-01-16
[sc-339] Improve eta-reduction to not apply on all functions

LunaAmora:


tests/golden_tests/compile_file/nested_eta.hvm
```diff
@@ -1,3 +0,0 @@
-X = λa a
-
-main = λb0 λb1 (X b0 b1)
```
developedby, line 3: We still want to have tests for eta reductions somewhere

developedby (changes requested):

tests/golden_tests/compile_file/nested_eta.hvm
LunaAmora: So, just leave the two eta tests there(to know that the eta pass is doing nothing to it)? Or you mean to test the eta on a generate function?

developedby: Actually, both would be good. You can keep one of them and rename to show the new purpose.

But what I meant mainly was the second, to test that eta reduction works as expected

developedby (approved):

PR #118, opened 2024-01-16
[sc-196] Add n-tuples to hvm-lang

imaqtkatt:


src/term/parser/parser.rs
```diff
@@ -266,6 +275,7 @@ where
       str,
       chr,
       sup,
+      ntup,
```
developedby, line 278: Couldn't we have just a single rule for any tuple length?

src/term/parser/parser.rs
```diff
@@ -294,6 +316,15 @@ where
...
+    let ntup = pattern
+      .clone()
+      .separated_by(just(Token::Comma))
+      .at_least(2)
+      .collect::<Vec<Pattern>>()
+      .delimited_by(just(Token::LParen), just(Token::RParen))
+      .map(|xs| make_tup_tree(&xs, |a, b| Pattern::Tup(Box::new(a), Box::new(b))))
+      .boxed();
+
     let tup = pattern
```
developedby, line 328: Same here

developedby (changes requested):

developedby (approved):

PR #123, opened 2024-01-17
[sc-356] Add tests for eta reduction

LunaAmora:


src/lib.rs
```diff
@@ -48,7 +48,7 @@ pub fn desugar_book(book: &mut Book, opt_level: OptimizationLevel) -> Result<Def
   book.make_var_names_unique();
   book.linearize_vars();
   if opt_level >= OptimizationLevel::Heavy {
-    book.eta_reduction();
+    book.eta_reduction(false);
```
developedby, line 51: Actually, we probably want this true or false to be gated by the optimization level. `bool.eta_reduction(opt_level >= OptimizationLevel::Heavy)`.

developedby: I don't see any reason not to always eta reduce generated functions

developedby (changes requested):

src/lib.rs
LunaAmora: This bool is the argument `reduce_all`. It's meant to be used as true when you want to reduce all the functions, even the non generated. (I pass true on the eta_reduce tests)

developedby (changes requested):

developedby (approved):

developedby (approved):

developedby (approved):

PR #124, opened 2024-01-18
[sc-334] Dont compile-time reduce functions extracted from main

imaqtkatt:


src/hvmc_net/pre_reduce.rs
```diff
@@ -17,7 +17,7 @@ pub fn pre_reduce_book(book: &mut BTreeMap<String, hvmc::ast::Net>, cross_refs:
   let mut rt_book = hvmc::ast::book_to_runtime(book);
   for (nam, net) in book.iter() {
     // Skip unnecessary work
-    if net.rdex.is_empty() {
+    if net.rdex.is_empty() || nam == DefNames::ENTRY_POINT {
```
developedby, line 20: Main still needs to be reduced to only have `REF` and `NUM` active pairs to run hvmc in compile mode.

We need to figure out a flag/options scheme to choose appropriately when to do no pre-reduction of main or minimal pre-reduction of main, but we definitely can't always skip it.

developedby (changes requested):

developedby (approved):

PR #128, opened 2024-01-18
[sc-345] Warn users when a match expression only uses Var patterns

imaqtkatt:


tests/snapshots/run_file__match_vars.hvm.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: tests/golden_tests.rs
+input_file: tests/golden_tests/run_file/match_vars.hvm
+---
+Could not run the code because of the previous warnings
```
developedby, line 5: Which warnings? They're not shown in the test result

developedby, line 5: We probably want for users to be able to bypass warnings if they really want to

developedby: And maybe indicate in this error message how they can do that.

developedby (changes requested):
Let's wait for this PR https://github.com/HigherOrderCO/hvm-lang/pull/122 and then rebase

developedby (approved):

PR #130, opened 2024-01-19
[sc-322] Fix unused let vars with unscoped lam/vars

LunaAmora:


src/term/transform/linearize.rs
```diff
@@ -204,6 +219,38 @@ fn term_to_affine(term: &mut Term, var_uses: &mut HashMap<Name, Val>, let_bodies
   };
 }
 
+fn check_unscoped(term: &Term) -> bool {
```
developedby, line 222:
```suggestion
fn has_unscoped(term: &Term) -> bool {
```

developedby: Also, could be a method of Term

developedby (changes requested):

PR #131, opened 2024-01-19
[sc-321] Add cli toggle for each hvml compilation pass

imaqtkatt:


src/lib.rs
```diff
@@ -47,13 +50,15 @@ pub fn desugar_book(book: &mut Book, opt_level: OptimizationLevel) -> Result<Def
...
-  book.detach_supercombinators(main);
-  if opt_level >= OptimizationLevel::Heavy {
+  book.eta_reduction(eta);
+  if supercombinators {
+    book.detach_supercombinators(main);
+  }
+  if ref_to_ref {
     book.simplify_ref_to_ref()?;
   }
   book.simplify_main_ref(main);
```
developedby, line 60: This could also be an option. There isn't any reason to disable it really, but I also don't see why not give the option

src/main.rs
```diff
@@ -29,8 +28,17 @@ struct Args {
...
   #[arg(short = 'l', help = "Linear readback (show explicit dups)")]
   pub linear: bool,
-  #[arg(short = 'O', default_value = "1", help = "Optimization level (0 or 1)", value_parser = opt_level_parser)]
-  pub opt_level: OptimizationLevel,
+
+  #[arg(
+    short = 'O',
+    value_delimiter = ' ',
+    action = clap::ArgAction::Append,
+    long_help = r#"Enables or disables the given optimizations
```
developedby, line 36: Isn't there a field to specify possible allowed values for this option? I think there was something like that, but maybe not if your option is type String

tests/golden_tests.rs
```diff
@@ -95,15 +95,15 @@ fn compile_term() {
 fn compile_file() {
```
developedby, line 95: Can we swap the names of these tests? Make this one `compile_file_o_all` and the other one just `compile_file`

developedby (changes requested):

src/lib.rs
imaqtkatt: Should it be under the ref-to-ref option or a new one?

LunaAmora: Isn't it necessary for correct output? For example it fixed this [task](https://app.shortcut.com/hocomp/story/350/fix-wrong-invalid-readback-from-string).
Or are we okay with having some wrong/inconsistent results with the user wants to disable a flag? @developedby

developedby: Results will still be correct, but readback will be strange.
Also, we shouldn't give an error when a string is not properly terminated since that could be on purpose. Instead it should just readback the term without converting into a string term

developedby: Also, we are already ok with results being wrong by disabling flags since we have one for ref_to_ref and one for supercombinators

developedby (changes requested):

src/main.rs
imaqtkatt: With the derive I don't think it is possible

developedby (changes requested):
cargo-insta doesn't keep track of which snapshots are valid, so remember to remove the old ones.

developedby (approved):

PR #133, opened 2024-01-19
[sc-359] Don't allow main to be called by any functions

LunaAmora:


tests/snapshots/compile_file__self_ref.hvm.snap
```diff
@@ -2,4 +2,4 @@
 source: tests/golden_tests.rs
 input_file: tests/golden_tests/compile_file/self_ref.hvm
 ---
-Definition main is a reference to itself
```
developedby, line 5: Can you make this still tests a self reference and have a separate test for main references?

developedby (changes requested):

developedby (approved):

PR #140, opened 2024-01-23
[sc-340] Fix unused dup var linearization

LunaAmora:


src/term/transform/linearize.rs
```diff
@@ -111,29 +111,17 @@ fn count_var_uses_in_term(term: &Term, uses: &mut HashMap<Name, Val>) {
...
 fn term_with_bind_to_affine(
   term: &mut Term,
   nam: &mut Option<Name>,
   var_uses: &mut HashMap<Name, Val>,
   let_bodies: &mut HashMap<Name, Term>,
 ) {
-  if let Some(nam_some) = nam {
-    if let Some(mut uses) = var_uses.get(nam_some).copied() {
+  if let Some(name) = nam {
+    if let Some(_) = var_uses.get(name).copied() {
```
developedby, line 122: If we don't check the number of uses anymore here, you can change it to `var_uses.contains_key(name)`

src/term/transform/linearize.rs
```diff
@@ -255,25 +252,56 @@ fn get_var_uses(nam: Option<&Name>, var_uses: &HashMap<Name, Val>) -> Val {
...
+// Can't it be unified in the same function? (given that the actual var uses in a term is already dynamic)
+//
+/// When a term is removed, the number of uses of a variable can change inside the term_to_affine call.
+/// This returns the updated count instead of using the one we calculated initially
+fn actual_var_use(var_uses: &HashMap<Name, Val>, name: &Name) -> Val {
+  var_uses.get(&used_var_counter(name)).copied().unwrap_or(0)
+}
+
+/// We use a special named variable to only count the number of actual var uses on the resulting term
+fn used_var_counter(var_name: &Name) -> Name {
```
developedby, line 265: Can you change the name of this function to reflect what it actually does (what is explained in the doc comment)?

tests/snapshots/compile_file_o_all__exp.hvm.snap
```diff
@@ -2,5 +2,5 @@
 source: tests/golden_tests.rs
 input_file: tests/golden_tests/compile_file_o_all/exp.hvm
 ---
-@main = ({5 ({5 a b} c) (c {5 b d})} (a d))
+@main = ({5 (a {5 b c}) ({5 c d} a)} (d b))
```
developedby, line 5: This should be only `({(a b) a} b)`, aka `@a let {a1 a2} = a; (a1 a2)`

src/term/transform/linearize.rs
```diff
@@ -181,13 +169,23 @@ fn term_to_affine(term: &mut Term, var_uses: &mut HashMap<Name, Val>, let_bodies
         uses => {
           term_to_affine(val, var_uses, let_bodies);
           term_to_affine(nxt, var_uses, let_bodies);
-          duplicate_let(nam, nxt, uses, val);
+
+          let mut new_uses = actual_var_use(var_uses, nam);
+
+          // If the number of uses changed (because a term was linearized out):
+          //  We could redo the whole pass on the term.
+          //  Or we can just create extra half-erased-dups `let {nam_n *} = nam_m_dup; nxt` to match the correct labels
```
developedby, line 177: We don't want to create unnecessary dups since they decrease performance.

src/term/transform/linearize.rs
```diff
@@ -255,25 +252,56 @@ fn get_var_uses(nam: Option<&Name>, var_uses: &HashMap<Name, Val>) -> Val {
...
-fn make_dup_tree(nam: &Name, nxt: &mut Term, uses: Val, dup_body: Option<&mut Term>) {
-  // TODO: Is there a difference between a list of dups and a complete binary tree of dups
-  // Creates this: "dup x1 x1_dup = body; dup x2 x2_dup = x1_dup; dup x3 x4 = x2_dup; nxt"
+// TODO: Is it really necessary to `count_var_uses_in_term` before `term_to_affine`?
+// Can't it be unified in the same function? (given that the actual var uses in a term is already dynamic)
+//
+/// When a term is removed, the number of uses of a variable can change inside the term_to_affine call.
+/// This returns the updated count instead of using the one we calculated initially
+fn actual_var_use(var_uses: &HashMap<Name, Val>, name: &Name) -> Val {
+  var_uses.get(&used_var_counter(name)).copied().unwrap_or(0)
```
developedby, line 261: Wouldn't it be better to use the `entry().or_default()` pattern we used in the other passes and in some other places here?

developedby (changes requested):

src/term/transform/linearize.rs
LunaAmora: Won't the pre_reduce pass remove them anyway? Or should find a better way to not generate on the first place?

developedby: pre_reduce will not remove them since they are not in active position. If we know that an inet package is normalized, it's safe to remove any half-erased dups, but that would have to be done in a separate pass.

Overall, it'd be better to not generate them in the first place, then we don't have to worry about removing them. I think this is the only place that generates them, so we haven't needed this additional pass so far.

LunaAmora: `pre_reduce will not remove them since they are not in active position.`
wdym?
image

tests/snapshots/compile_file_o_all__exp.hvm.snap
LunaAmora: I don't understand why it should?
the file is just `main = ((λfλx (f (f x))) (λfλx (f (f x))))`, we don't do anything that could transform to that

developedby: Sorry, the change above confused me and I thought that was a new file and this was the output

developedby (changes requested):

src/term/transform/linearize.rs
```diff
@@ -255,25 +250,56 @@ fn get_var_uses(nam: Option<&Name>, var_uses: &HashMap<Name, Val>) -> Val {
...
+// Can't it be unified in the same function? (given that the actual var uses in a term is already dynamic)
+//
+/// When a term is removed, the number of uses of a variable can change inside the term_to_affine call.
+/// This returns the updated count instead of using the one we calculated initially
+fn actual_var_use<'a>(var_uses: &'a mut HashMap<Name, Val>, name: &Name) -> &'a mut u64 {
+  var_uses.entry(actual_var_use_counter(name)).or_default()
+}
+
+/// We use a special named variable to only count the number of actual var uses on the resulting term
+fn actual_var_use_counter(var_name: &Name) -> Name {
```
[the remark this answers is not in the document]
developedby: https://github.com/HigherOrderCO/hvm-lang/pull/140
But I still ask you to improve this function's name and description now and later we'll deal with refactoring this.

developedby (approved):

PR #141, opened 2024-01-23
Clarify fusing functions documentation [sc-394] and add explanation on removing duplications

FranchuFranchu:
…n removing duplication.

### DIFF
--- a/docs/writing-fusing-functions.md
+++ b/docs/writing-fusing-functions.md
@@ -8,7 +8,7 @@ false = λt λf f
 Matching on values of this representation is simply calling the boolean value with what the function should return if the boolean is true and if the boolean is false.
 ```rs
 if boolean case_true case_false = (boolean case_true case_false)
-main = (if true 42 37) 
+main = (if true 42 37)
 // Outputs 42
 
 // Or alternatively:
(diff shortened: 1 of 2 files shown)


docs/writing-fusing-functions.md
```diff
@@ -67,9 +67,58 @@ main = λxλt1λf1 (λtλf (x f t) f1 t1)
 main = λxλt1λf1 (λf (x f f1) t1))
 main = λxλt1λf1 (x t1 f1)
 ```
-Wow! Simply by replacing lambda arguments with the values applied to them, we were able to make `(not (not x))` not grow in size. This is what fusing means, and it's a really powerful tool to make programs faster.
+Wow! Simply by replacing lambda arguments with the values applied to them, we were able to make `(not (not x))` not grow in size. This is what fusing means, and it's a really powerful tool to make programs faster.
```
developedby, line 70:
```suggestion
Wow! Simply by replacing lambda arguments with the values applied to them, we were able to make `(not (not x))` not grow in size. This is what fusing means, and it's a really powerful tool to make programs faster.
```

docs/writing-fusing-functions.md
```diff
@@ -67,9 +67,58 @@ main = λxλt1λf1 (λtλf (x f t) f1 t1)
...
+
+```rs
+add = λa λb
+	let {b1 b2} = b;
+	let case_succ = λa_pred (succ (add a_pred b0))
+	let case_zero = b1
+	(a case_succ case_zero)
+```
+
+However, only one of the two duplicated values is ever used. The other one is discarded. This means that this duplication is completely useless! It makes this function suboptimal. In fact, as of commit `ebd469`, it will crash HVM, because it's recursive and HVM2 is eager.
```
developedby, line 106: It won't crash, it will get stuck in an infinite loop (maybe eventually crash with OoM but that will take a long time)

developedby (changes requested):

docs/writing-fusing-functions.md
FranchuFranchu: Replaced "crash" by "hang" in 26f4e65

docs/writing-fusing-functions.md
FranchuFranchu: I removed this instance of trailing whitespace and other trailing whitespace in 26f4e65

developedby (approved):

PR #142, opened 2024-01-23
[sc-387] String readback bug

imaqtkatt:


tests/snapshots/run_file__nested_str.hvm.snap
```diff
@@ -0,0 +1,6 @@
+---
+source: tests/golden_tests.rs
+input_file: tests/golden_tests/run_file/nested_str.hvm
+---
+Invalid readback: [InvalidStrTerm]
```
developedby, line 5: Shouldn't we output malformed strings in their constructor form? Since we can expect them to happen for example in partially applied functions.

LunaAmora: Just fixed this, will add a commit soon

src/term/net_to_term.rs
```diff
@@ -327,6 +352,51 @@ impl<'a> Reader<'a> {
...
+      let snil = Term::Var { nam: Name::new(SNIL) };
+      let last_term = string_cons(cons_term, snil);
+      *term = string_app(val, last_term)
+    }
+    app @ Term::App { .. } => insert_string_cons(app, cons_term),
+    _ => unreachable!(),
+  }
+}
+
+// TODO: `λa (SCons a (SCons 'b' (SCons 'c' SNil)))` could be read back as `λa (SCons a "bc")`
```
developedby, line 368: Why not simply try twice?

In `resugar_adts`:
```
match term {
  ...
  Term::Lam {...} if tag == Tag::string => {
    // The second one recreates the constructor and calls resugar_adts recursively on the ctr fields. Not necessarily a separate function, just for the sake of keeping this comment short
    resugar_str(term).or_else(|| rebuild_constructor(term)) 
  }
  ...
}
```
This would allow you to still use the old very simple resugar_str function and have the readback of substrings in a malformed string like mentioned in this todo.

developedby: Also, I feel like we're polluting this file too much and it's getting a bit out of hand (not due to this PR specifically but it's growing and getting messier over time).

src/term/net_to_term.rs
```diff
@@ -327,6 +352,51 @@ impl<'a> Reader<'a> {
...
+// but the order in witch decode_str is run complicates this as the `SNil` could be another variable.
+// If decode_str is changed to the reverse order,
+// the hability to just push new chars to the string would be lost,
+// but the readback could be improved.
+//
+/// Inserts a term as an application with the last argument of an app chain
```
developedby, line 374: This comment doesn't seem to reflect what the function does

src/term/net_to_term.rs
developedby: Another alternative would be to first resugar all constructors ignoring the existence of lists and strings, and then right after try to construct the list and string sugars from the constructors. That would also make the code a bit cleaner i think
```
// in resugar_adts
match term {
  ...
  Term::Lam {tag, ...} => {
    if book.adt.contains(tag) {
      // Tries to recover this constructor and calls resugar_adt recursively
      *term = rebuild_constructors(body);
      match tag {
        // Since this function is recursive, if the constructor fields were lists or strings, they will already have been resugared, making these 2 resugar_functions simpler.
        Tag::string => resugar_string(term)
        Tag::list => resugar_list(term)
      }
    }
  }
  ...
}
```

tests/snapshots/run_file__nested_str.hvm.snap
```diff
@@ -0,0 +1,6 @@
+---
+source: tests/golden_tests.rs
+input_file: tests/golden_tests/run_file/nested_str.hvm
+---
+Invalid readback: [InvalidStrTerm, InvalidStrTerm, InvalidStrTerm, InvalidStrTerm, InvalidStrTerm]
```
developedby, line 5: The invalid readback message is very hard to understand (not due to this PR, i know)

developedby (changes requested):
Your PR is good, but I think we should refactor the resugaring code instead of adding this functionality/fix on top, since the complexity is starting to grow too much

src/term/net_to_term.rs
developedby: For now I'll accept it, but I'll open a separate story for us to refactor it

tests/snapshots/run_file__nested_str.hvm.snap
developedby: https://app.shortcut.com/hocomp/story/401/improve

developedby (approved):

PR #143, opened 2024-01-23
[sc-385] Resugar of lists on the readback

imaqtkatt:


LunaAmora (approved):

src/term/net_to_term.rs
```diff
@@ -297,6 +297,34 @@ impl<'a> Reader<'a> {
     go(term, &mut s, self);
     Term::Str { val: s }
   }
+
+  fn decode_list(&mut self, term: &mut Term) -> Term {
```
developedby, line 301: can you call it `resugar_list` and `resugar_str` to be like `resugar_adts` (which maybe should be `resugar_constructors`)

developedby (changes requested):

developedby (approved):

PR #146, opened 2024-01-24
[sc-401] Improve invalid readback warnings

imaqtkatt:


src/term/net_to_term.rs
```diff
@@ -722,3 +724,81 @@ impl<'a> Reader<'a> {
     T::default()
   }
 }
+
+/// A structure that implements display logic for Readback Errors.
+pub struct ReadbackErrors(Vec<ReadbackError>, DefNames);
```
developedby, line 729: It doesn't need to own a `DefNames`, right? Isn't a reference good enough?

developedby (approved):

tests/snapshots/run_file__wrong_string.hvm.snap
```diff
@@ -0,0 +1,9 @@
+---
+source: tests/golden_tests.rs
+input_file: tests/golden_tests/run_file/wrong_string.hvm
+---
+Invalid readback:
+Invalid String Character value '(*, 4)'
+Invalid String Character value '*'
+
+""
```
LunaAmora, line 9: Shouldn't the result still be readback as `(SCons (*, 4) (SCons * SNil))`?

Oh, this is happening on master too. I thought #146 had fixed this. Should be fixed later then

LunaAmora (approved):

developedby (changes requested):
Selected approve by accident

developedby (approved):

PR #147, opened 2024-01-24
[sc-323] Make compilation passes not as conservative

LunaAmora:


src/term/check/unbound_vars.rs
```diff
@@ -25,8 +25,11 @@ impl Term {
 
     // Check global vars
     for (nam, (declared, used)) in globals.into_iter() {
-      if used && !declared {
-        return Err(format!("Unbound unscoped variable '${nam}'"));
+      match (used, declared) {
+        (1, 1) => {}
+        (0, _) => return Err(format!("Missing unscoped variable use for 'λ${nam}'")),
```
developedby, line 30:
```suggestion
        (0, _) => return Err(format!("Unscoped variable from lambda 'λ${nam}' is never used")),
```

src/term/check/unbound_vars.rs
```diff
@@ -25,8 +25,11 @@ impl Term {
...
+        (_, 0) => return Err(format!("Unbound unscoped variable '${nam}'")),
+        (_, _) => return Err(format!("Unscoped variable '${nam}' used more than once")),
```
developedby, line 32: A bit confusing that you're mixing 1 declaration multiple use and multiple declarations.

developedby (changes requested):

developedby (approved):

PR #151, opened 2024-01-25
[sc-400] Refactor adt resugaring in readback

LunaAmora:


src/term/net_to_term.rs
```diff
@@ -275,52 +275,31 @@ impl<'a> Reader<'a> {
   fn resugar_string(&mut self, term: &mut Term) -> Term {
```
developedby, line 275: Could you refactor these `impl` blocks so that they're more organized? Right now `resugar_string` and `resugar_list` are in a separate `impl` from `resugar_adts` and we have 3 impl blocks for Reader and 2 impl blocks for Term.

Maybe separate this file into sections, each doing one thing, or extract adt resugaring to a separate pass in a different file. 
I think overall this file is a bit disorganized.

developedby (changes requested):
Could you also try to refactor the `resugar_adts` a bit? Right now it's very hard to understand with the large blocks of code inside the matches. It could help adding some comments explaining what each function should do and what is the logic/reasoning behind the way we are implementing this resugaring transformation.

imaqtkatt (approved):

src/term/resugar.rs
```diff
@@ -0,0 +1,224 @@
...
+      Term::Lnk { .. }
+      | Term::Num { .. }
+      | Term::Var { .. }
+      | Term::Str { .. }
+      | Term::Ref { .. }
+      | Term::Era => {}
+    }
+  }
+
+  fn resugar_adt_cons(&mut self, term: &mut Term, adt: &Adt, adt_name: &Name) {
```
developedby, line 62: Can you add a short description of this?

src/term/resugar.rs
```diff
@@ -0,0 +1,224 @@
...
+    match cur {
+      Term::Var { nam } if nam == &arm_name => {}
+      _ => return self.error(ReadbackError::InvalidAdt),
+    }
+
+    *cur = self.book.def_names.get_ref(ctr);
+    *term = std::mem::take(app);
+  }
+
+  fn resugar_adt_match(&mut self, term: &mut Term, adt_name: &Name, adt: &Adt) {
```
developedby, line 110: Here as well

src/term/mod.rs
```diff
@@ -258,6 +267,10 @@ impl Book {
     let name = self.def_names.remove(def_id);
     name.zip(def)
   }
+
+  pub fn is_generated_def(&self, def_id: DefId) -> bool {
```
developedby, line 271: This causes confusion with the Definition.generated field.
Here we're talking about a generated name. There it's a generated body.
Could you add this distinction in both places?

developedby (changes requested):
Good job

developedby (approved):

PR #155, opened 2024-01-26
[sc-409] Unused unscoped lambda in match arm shows wrong error

imaqtkatt:


src/term/check/unbound_vars.rs
```diff
@@ -19,9 +25,9 @@ impl Book {
 impl Term {
   /// Checks that all variables are bound.
   /// Precondition: References have been resolved.
```
developedby, line 27: References have been resolved, implicit binds have been resolved

src/term/check/unbounds.rs
```diff
@@ -0,0 +1,9 @@
+use crate::term::Book;
+
+impl Book {
```
developedby, line 3: Unless we plan to do something else with this, I don't think this is a good idea.
This is simple enough that it seems better to me to just always have it inlined

src/lib.rs
```diff
@@ -52,11 +52,13 @@ pub fn desugar_book(
   book.generate_scott_adts();
   book.resolve_refs()?;
   encode_pattern_matching(book, &mut warnings)?;
+  book.check_unbound_vars()?;
```
developedby, line 55: You can add a comment that this is just a sanity check

src/lib.rs
```diff
@@ -52,11 +52,13 @@ pub fn desugar_book(
...
   book.normalize_native_matches()?;
   book.check_unbound_vars()?;
   book.make_var_names_unique();
   book.linearize_vars();
   book.eta_reduction(eta);
+  book.check_unbounds()?;
```
developedby, line 61: Same as above

developedby: Also checking for unbound pats here doesn't make sense since they have already been encoded.
A sanity check of unbound patterns would be better after the intermediate steps where we're messing with the definitions (after extracting matches and after flatten, although i think it's not needed)

developedby (changes requested):

developedby (approved):

PR #157, opened 2024-01-26
[sc-414] Builtins not generated if no literals

imaqtkatt:


src/term/transform/encode_lists.rs
```diff
@@ -82,6 +82,7 @@ impl Term {
 impl Pattern {
   pub fn encode_lists(&mut self) -> bool {
     match self {
+      Pattern::Var(Some(Name(nam))) => nam == LNIL,
```
developedby, line 85: Can you add a TODO here that this should be removed once we separate adding the constructors to the book from desugaring list terms?
This is necessary to work right now, but it's not the right way we should be doing it.

src/term/transform/encode_lists.rs
```diff
@@ -92,8 +93,8 @@ impl Pattern {
 
         true
       }
-      Pattern::Ctr(_, pats) => {
-        let mut uses = false;
+      Pattern::Ctr(Name(nam), pats) => {
+        let mut uses = nam == LCONS;
```
developedby, line 97: nam == LCONS || nam == LNIL

If we write the pattern like this `(Head (LNil)) = LNil` then it will be understood as a constructor by the parser without needing the delayed name resolving

src/term/transform/encode_strs.rs
```diff
@@ -80,3 +81,19 @@ impl Term {
     }
   }
 }
+
+impl Pattern {
+  pub fn encode_str(&mut self) -> bool {
+    match self {
+      Pattern::Var(Some(Name(nam))) => nam == SNIL,
```
developedby, line 88: Same as I've commented for list

src/term/transform/encode_strs.rs
```diff
@@ -80,3 +81,19 @@ impl Term {
...
+      Pattern::Ctr(Name(nam), pats) => {
+        let mut uses = nam == SCONS;
```
developedby, line 90: Same here

developedby (changes requested):

developedby (approved):

src/term/transform/encode_strs.rs
```diff
@@ -80,3 +81,19 @@ impl Term {
...
+    match self {
+      Pattern::Var(Some(Name(nam))) => nam == SCONS || nam == SNIL,
+      Pattern::Ctr(Name(nam), pats) => {
+        let mut uses = nam == SCONS || nam == SNIL;
+        for pat in pats {
+          uses |= pat.encode_str();
+        }
+        uses
+      }
+      _ => false,
```
LunaAmora, line 96: missing Pattern::Tup & Pattern::List here.

also, can use this function instead of this line on the two files
```rs
    pat.names().chain(pat.ctrs()).any(|Name(n)| matches!(n.as_str(), ...));
```

LunaAmora (changes requested):

LunaAmora (approved):

PR #160, opened 2024-01-29
[sc-411] Add Cli opt flag to disable the pre-reduce pass

imaqtkatt:


src/lib.rs
```diff
@@ -186,6 +188,9 @@ pub struct Opts {
 
   /// Enables [term::transform::simplify_main_ref].
   pub simplify_main: bool,
+
+  /// TODO
```
LunaAmora, line 192: Missing this todo

docs/toggleable-passes.md
```diff
@@ -146,3 +147,29 @@ main = (a a)
 @id = (a a)
 @main = @id
 ```
+
+## Cross-refs
+
+Is triggered only if the [pre-reduce](#pre-reduce) option is enabled.
+
+If enabled, will dereferentiate and try to find the smallest net.
```
LunaAmora, line 155: I think the word here is `dereference`

LunaAmora (approved):

docs/toggleable-passes.md
```diff
@@ -10,6 +10,7 @@
 | `-Opre-reduce` `-Ono-pre-reduce` | Disabled | [pre-reduce](#pre-reduce) |
 | `-Osupercombinators` `-Ono-supercombinators` | Enabled  | [supercombinators](#supercombinators) |
 | `-Osimplify-main` `-Ono-simplify-main` | Disabled | [simplify-main](#simplify-main) |
+| `-Ocross-refs` `-Ono-cross-refs` | Disabled | [cross-refs](#cross-refs) |
```
developedby, line 13: I suggest `pre-reduce` and `pre-reduce-refs`.
Both are combined so that `-Opre-reduce -Opre-reduce-refs` is full compile-timme reduction of functions, `-Opre-reduce -Ono-pre-reduce-refs` is limited compile-time reduction and `-Ono-pre-reduce` is no compile-time reduction.

docs/toggleable-passes.md
```diff
@@ -146,3 +147,29 @@ main = (a a)
 @id = (a a)
 @main = @id
 ```
+
+## Cross-refs
+
+Is triggered only if the [pre-reduce](#pre-reduce) option is enabled.
+
+If enabled, will dereference and try to find the smallest net.
```
developedby, line 155: "When enabled, the compile-time reduction pass will also reduce reference applications, leading to full normalization of the compiled net of each function."
Something like this, explaining with a little bit more details

docs/toggleable-passes.md
```diff
@@ -146,3 +147,29 @@ main = (a a)
...
+
+Example:
+```rs
+foo = (+ 2 2)
+
+bar = (+ foo foo)
+
+main = (bar)
+
+// -Ocross-refs, compilation output
```
developedby, line 165: Would be interesting to also add the -Ono-pre-reduce case here, for contrast.

docs/toggleable-passes.md
developedby: Also it would be nice if `-Opre-reduce-refs` could imply `-Opre-reduce` if `-Ono-pre-reduce` was not already specified.
Maybe this would complicate too much the logic of which flags are enabled, both for the user and ion the codebase, but consider this option and choose whatever you think is best.

developedby (changes requested):

developedby (approved):

PR #161, opened 2024-01-29
[sc-306] merge identical functions

LunaAmora:
Is this pass on the definitions satisfactory or would the pass on the INet/Net level be better?


developedby (commented):
Are you sure that all alpha-equivalent terms will have the same hash. It looks to be true after giving variable unique names and linearizing them, but I'm wondering if there is some catch.

The reason I suggested doing it on the inet level is that we're sure that equal terms become equal inets, regardless of variable names.

imaqtkatt (approved):
I think if it works, it is fine, although I don't like this idea of hashing an ast

developedby (approved):

PR #162, opened 2024-01-29
[sc-415] Improve skip_warnings cli flag

imaqtkatt:


src/main.rs
```diff
@@ -14,8 +15,17 @@ struct Args {
...
+    short = 'W',
+    value_delimiter = ' ',
+    action = clap::ArgAction::Append,
+    long_help = r#"Skip compilation warnings
+    all, unused-defs, match-only-vars"#
+  )]
+  pub skip_warnings: Vec<String>,
+
+  #[arg(long, help = "Stops if any warning was produced")]
+  pub errors: bool,
```
LunaAmora, line 28: I really don't like this as a cli arg, i would prefer a more descriptive name

imaqtkatt: I don't like it too. But what name?

src/lib.rs
```diff
@@ -240,6 +248,52 @@ impl Opts {
...
+impl WarningOpts {
+  pub fn all() -> Self {
+    Self { all: true, match_only_vars: true, unused_defs: true }
+  }
+
+  // TODO(refactor): this kind of code does not look good
+  pub fn from_vec(&mut self, values: Vec<String>) -> Result<(), String> {
+    for value in values {
+      match value.as_ref() {
+        "all" => self.all = true,
```
developedby, line 267: Can we follow the same naming scheme as for selecting compiler passes? `-Wunused-defs` for activating, `-Wno-unused-defs` for disabling the warning.
With default for now being all of them activated.

src/lib.rs
```diff
@@ -240,6 +248,52 @@ impl Opts {
   }
 }
 
+#[derive(Default)]
+pub struct WarningOpts {
+  pub all: bool,
```
developedby, line 253: We shouldn't have a `all` option, it should instead activate all the individual options.

src/lib.rs
```diff
@@ -240,6 +248,52 @@ impl Opts {
...
+        "unused-defs" => self.unused_defs = true,
+        "match-only-vars" => self.match_only_vars = true,
+        other => return Err(format!("Unknown option '{other}'.")),
+      }
+    }
+    Ok(())
+  }
+
+  /// Removes warnings based on the enabled flags.
+  pub fn filter(&self, wrns: Vec<Warning>) -> Vec<Warning> {
```
developedby, line 277: I think it would be better to have this be: `pub fn is_warning_enabled(&self, warn: &Warning) -> bool`.
Then you can do `wrns.iter().filter(|w| opt.is_warning_enabled(w))`.

src/lib.rs
```diff
@@ -98,17 +98,25 @@ pub fn run_book(
...
-    } else {
-      eprintln!("{warnings}");
+  // if both enabled, ignore the all option
+  if warning_opts.all && errors {
+    warning_opts.all = false;
+  }
+  if !warning_opts.all {
+    if !warnings.is_empty() {
+      let warnings = warnings.iter().join("\n");
+      if errors {
```
developedby, line 115: We should deny on individual warnings. Just as suggested below a `is_warning_enabled` function, we could have a `is_warning_denied` and filter those that are true.

I also suggest having `-D<warning>` and `-Dno-<warning>` just like for warnings and optimizations.

src/main.rs
```diff
@@ -14,8 +15,17 @@ struct Args {
   #[arg(short, long)]
   pub verbose: bool,
 
-  #[arg(short = 'w', help = "Skip compilation warnings")]
-  pub skip_warnings: bool,
+  #[arg(
+    short = 'W',
+    value_delimiter = ' ',
+    action = clap::ArgAction::Append,
+    long_help = r#"Skip compilation warnings
```
developedby, line 22: `-W` skipping warnings is unintuitive. Better to have -Wno-<warning> for skipping.

src/main.rs
```diff
@@ -14,8 +15,17 @@ struct Args {
   #[arg(short, long)]
   pub verbose: bool,
 
-  #[arg(short = 'w', help = "Skip compilation warnings")]
-  pub skip_warnings: bool,
+  #[arg(
+    short = 'W',
```
developedby, line 19: Can we also have a long option? --warning or whatever

developedby (changes requested):

src/lib.rs
```diff
@@ -240,6 +242,56 @@ impl Opts {
...
+  pub fn deny(&mut self, values: Vec<String>) -> Result<(), String> {
+    self.iter_vals(values, Self::all(), true)
+  }
+
+  pub fn allow(&mut self, values: Vec<String>) -> Result<(), String> {
+    self.iter_vals(values, Self::default(), false)
+  }
+
+  /// Filters warnings based on the enabled flags.
+  pub fn filter(&self, wrns: Vec<Warning>) -> Vec<Warning> {
```
developedby, line 279: `filter_enabled` is a better name

developedby (changes requested):

developedby (approved):

LunaAmora (approved):

PR #166, opened 2024-01-30
[sc-389][sc-403] Refactor and complete pattern matching function compilation

developedby:


developedby:
This undoes the optimization of not passing forward Era patterns on flattening, but that was bugged.

At some point we can reintroduce this optimization, but implemented correctly. I'd rather we more to a more robust algorithm for pattern matching that works directly without the flattening step.

LunaAmora (approved):

src/lib.rs
```diff
@@ -84,6 +86,7 @@ pub fn encode_pattern_matching(book: &mut Book, warnings: &mut Vec<Warning>) ->
   book.check_unbound_vars()?;
   book.extract_adt_matches(warnings)?;
   book.flatten_rules();
+  eprintln!("{book}\n========================\n");
```
imaqtkatt, line 89: Remove this print

developedby, line 89:
```suggestion
```

imaqtkatt (approved):

PR #169, opened 2024-02-01
[sc-418] Improve hvml CLI options

imaqtkatt:
Couldn't add a compile to rust option since those functions are not present in the hvmc lib

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -253,6 +253,12 @@ impl Opts {
     }
     Ok(())
   }
+
+  pub fn check(&self) {
+    if !self.supercombinators {
+      println!("Warning: Running in strict mode can lead some functions expand infinitely.");
+    }
+  }
 }
 
 #[derive(Default, Clone, Copy)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -101,17 +101,7 @@ pub fn run_book(
 ) -> Result<(Term, DefNames, RunInfo), String> {
   let CompileResult { core_book, hvmc_names, labels, warnings } = compile_book(&mut book, opts)?;
 
-  let warns = warning_opts.filter(&warnings, WarnState::Warn);
-  if !warns.is_empty() {
-    let warns = warns.iter().join("\n");
-    eprintln!("Warnings:\n{warns}");
-  }
-
-  let denies = warning_opts.filter(&warnings, WarnState::Deny);
-  if !denies.is_empty() {
-    let denies = denies.iter().join("\n");
-    return Err(format!("{denies}\nCould not run the code because of the previous warnings"));
-  }
+  display_warnings(warning_opts, &warnings)?;
 
   fn debug_hook(net: &Net, book: &Book, hvmc_names: &HvmcNames, labels: &Labels, linear: bool) {
     let net = hvmc_to_net(net, &|id| hvmc_names.hvmc_name_to_id[&id]);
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -214,31 +204,57 @@ impl Opts {
   }
 }
 
+#[derive(clap::ValueEnum, Clone, Debug)]
+pub enum OptArgs {
```
developedby, line 208: All this CLI interface code should be moved to `main.rs`.
hvml as a library shouldn't be aware of CLI

src/main.rs
```diff
@@ -8,57 +8,86 @@ use std::path::PathBuf;
...
+      supercombinators is enabled by default."#,
+    )]
+    cli_opts: Vec<hvml::OptArgs>,
 
-  #[arg(help = "Path to the input file")]
-  pub path: PathBuf,
+    #[command(flatten)]
+    wopts: WOpts,
+  },
+  /// Runs the lambda-term level optimization passes.
```
developedby, line 75:
```suggestion
  /// Runs the lambda-term level desugaring passes.
```
Since we don't perform any optimizations here anymore

src/main.rs
```diff
@@ -8,57 +8,86 @@ use std::path::PathBuf;
...
 #[group(multiple = true)]
 struct WOpts {
   #[arg(
     short = 'W',
     long = "warn",
     value_delimiter = ' ',
     action = clap::ArgAction::Append,
-    long_help = r#"Show compilation warnings
-    all, unused-defs, match-only-vars"#
+    long_help = "Show compilation warnings",
```
developedby, line 90: From just the help this is a bit ambiguous and seems more like it will show/allow/deny all warnings.

It would be better along the lines of "Show the specified compilation warning".

It would also be nice if the help command could show all the possible values somewhere and group these three options together somehow, eg:
"-W/D/A --warn/deny/allow : Show/Deny/Allow compilation warnings".
But I don't know how feasible that would be with clap

src/main.rs
```diff
@@ -107,78 +122,100 @@ fn mem_parser(arg: &str) -> Result<usize, String> {
...
-      }
-    }
     Ok(())
   }
   if let Err(e) = run() {
     eprintln!("{e}");
   }
 }
+
+fn run_mode(cli: Cli, verbose: &dyn Fn(&hvml::term::Book), mut opts: Opts) -> Result<(), String> {
```
developedby, line 149: This name is a bit strange because we have a mode called "run".
Maybe something like `execute_cli_mode` to avoid the confusion.

developedby (changes requested):

src/lib.rs
imaqtkatt: Can't move it to `main.rs`, the only option I see is to crate a public `cli` crate

LunaAmora: You can move it, just need to not have a impl block for Opts there

src/main.rs
```diff
@@ -107,78 +122,100 @@ fn mem_parser(arg: &str) -> Result<usize, String> {
...
-    let mut opts = Opts::light();
-    Opts::from_vec(&mut opts, args.opts)?;
-    let warning_opts = args.wopts.get_warning_opts();
-
-    let mut book = load_file_to_book(&args.path)?;
-    if args.verbose {
-      println!("{book:?}");
-    }
+    let cli = Cli::parse();
+    let opts = Opts::light();
```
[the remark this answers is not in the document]
developedby: It could instead be called inside `from_cli_opts`, like this:
```
fn from_cli_opts(cli_opts) -> Self {
  let mut opts = Opts::light();
  for cli_opt in cli_opts {
     match cli_opt {...}
  }
}
```

src/lib.rs
developedby: You could instead have a `OptArgs::modify_opts(&self, &mut Opts) -> ()` if you want to keep it a method, and then `cli_opts.for_each(|o| o.modify_opts(opts))`

developedby (changes requested):
See the unresolved comment.

src/lib.rs
```diff
@@ -219,31 +209,10 @@ impl Opts {
...
-        "no-simplify-main" => self.simplify_main = false,
-        "pre-reduce-refs" => self.pre_reduce_refs = true,
-        "no-pre-reduce-refs" => self.pre_reduce_refs = false,
-        "merge-definitions" => self.merge_definitions = true,
-        "no-merge-definitions" => self.merge_definitions = false,
-        other => return Err(format!("Unknown option '{other}'.")),
-      }
+  pub fn check(&self) {
+    if !self.supercombinators {
+      println!("Warning: Running in strict mode can lead some functions expand infinitely.");
```
developedby, line 214:
```suggestion
      println!("Warning: Running in strict mode without enabling the supercombinators pass can lead to some functions expanding infinitely.");
```

developedby (changes requested):

developedby (approved):

PR #170, opened 2024-02-01
Improves some functions, apply clippy adjustments and fmt to the codebase

LunaAmora:


src/net/net_to_hvmc.rs
```diff
@@ -54,7 +54,7 @@ fn net_tree_to_hvmc_tree(
       rgt: Box::new(var_or_subtree(inet, Port(tree_root, 2), port_to_var_id, id_to_hvmc_name)),
     },
     NodeKind::Con { lab: Some(lab) } => Tree::Dup {
-      lab: (lab + 1) << 1 | 0,
+      lab: (lab + 1) << 1,
```
developedby, line 57: Can you keep this one? So that it pairs correctly with the dup node below.
Also, it's like "label plus bit 0 is set to 0" so I think how it was before conveys that better.

src/term/check/unbound_pats.rs
```diff
@@ -39,10 +40,12 @@ impl Pattern {
...
           check.extend(args.iter());
         }
-        Pattern::Var(_) => (),
-        Pattern::Num(_) => (),
-        Pattern::Tup(_, _) => (),
+        Pattern::Tup(fst, snd) => {
+          unbounds.extend(fst.unbound_pats(is_ctr));
+          unbounds.extend(snd.unbound_pats(is_ctr));
+        }
         Pattern::List(..) => unreachable!(),
```
developedby, line 47: Can you implement this for Pattern::List?
I feel like this pass is so basic that at some point in the future we might change the order of things.

src/term/term_to_net.rs
```diff
@@ -260,7 +260,7 @@ impl<'a> EncodeTermState<'a> {
       }
       // core: #val
       Term::Num { val } => {
-        // debug_assert!(*val <= LABEL_MASK); // Uneeded?
+        // debug_assert!(*val <= LABEL_MASK); // Unneeded?
```
developedby, line 263: Are we checking somewhere that the number fits in an hvmc node?

Also I think we can remove this line

LunaAmora: Should we add a check now? i don't think we are

imaqtkatt (approved):

developedby (approved):

PR #175, opened 2024-02-05
[sc-430] Constructor readback not working in some cases

imaqtkatt:


developedby (approved):

src/term/mod.rs
```diff
@@ -252,6 +252,10 @@ impl Tag {
   pub fn list_lcons_head() -> Self {
     Self::Named(Name::new("List.LCons.head"))
   }
+
+  pub fn list_lcons_tail() -> Self {
+    Self::Named(Name::new("List.LCons.tail"))
```
LunaAmora, line 257: I think we could improve this functions to use the constants instead of magic strings, like Name(format!("{LIST}.{LCONS}.{HEAD}")). Could you update this now?

LunaAmora (approved):

PR #176, opened 2024-02-05
[sc-433] Add Term::Invalid for invalid readbacks

LunaAmora:


imaqtkatt (approved):

tests/snapshots/readback_lnet__invalid_op2_op2.hvm.snap
```diff
@@ -0,0 +1,8 @@
+---
+source: tests/golden_tests.rs
+input_file: tests/golden_tests/readback_lnet/invalid_op2_op2.hvm
+---
+Readback Warning:
+Invalid Op2 with 2 occurrences
```
developedby, line 6: I think you forgot to update this test

src/term/mod.rs
```diff
@@ -153,6 +153,7 @@ pub enum Term {
   },
   #[default]
   Era,
+  Invalid,
```
developedby, line 156: This should probably be default

developedby (changes requested):

developedby (approved):

PR #177, opened 2024-02-05
[sc-432] Refactor list and string decoding

imaqtkatt:
Since the readback warnings are not good with this refactor, we should discuss about what to do or how to implement it properly

### DIFF
--- a/tests/snapshots/run_lazy__nested_str.hvm.snap
+++ b/tests/snapshots/run_lazy__nested_str.hvm.snap
@@ -2,4 +2,4 @@
 source: tests/golden_tests.rs
 input_file: tests/golden_tests/run_lazy/nested_str.hvm
 ---
-(((SCons "a" ""), (SCons 97 (SCons "bc" ""))), ((SCons "ab" "c"), (SCons "ab" (SCons "cd" ""))))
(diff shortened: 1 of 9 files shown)


developedby (approved):

tests/snapshots/run_lazy__nested_str.hvm.snap
```diff
@@ -2,11 +2,4 @@
...
 ---
-Readback Warning:
-Invalid String Character value '"a"'
-Invalid String Character value '"bc"'
-Invalid String Character value '"ab"'
-Invalid String Character value '"cd"'
-Invalid String Character value '"ab"'
-
-(((SCons "a" SNil), (SCons 97 (SCons "bc" SNil))), ((SCons "ab" "c"), (SCons "ab" (SCons "cd" SNil))))
+(((SCons "a" ""), (SCons 97 (SCons "bc" ""))), ((SCons "ab" "c"), (SCons "ab" (SCons "cd" ""))))
```
LunaAmora, line 5: I would prefer SNil instead of `""`

LunaAmora (approved):
I prefer having the flexibility of the old code for better errors, but this way is good too

PR #181, opened 2024-02-08
[sc-410] Add untagged scott encoding option for ADTs

developedby:


src/term/resugar.rs
```diff
@@ -236,16 +244,19 @@ impl<'a> Reader<'a> {
...
-        if let Term::List { els: tail } = tail {
+        if ctr.as_str() == LCONS && let Term::List { els: tail } = tail {
           let mut els = vec![*el.clone()];
           els.extend(tail);
           Term::List { els }
         } else {
           // Make the constructor again
-          let fun = Term::App { tag: Tag::Static, fun: ctr.clone(), arg: el.clone() };
-          Term::App { tag: Tag::Static, fun: Box::new(fun), arg: Box::new(tail) }
+          Term::call(Term::Ref { def_name: ctr.clone() }, [*el.clone(), tail])
```
LunaAmora, line 259: We can just std::mem::take(term) here instead of cloning

LunaAmora (approved):

src/term/resugar.rs
developedby: I guess we can, but I'm refactoring this right now, so I'll leave it like that in this PR.

imaqtkatt (approved):

PR #182, opened 2024-02-08
[sc-435] Add support for custom entrypoints

imaqtkatt:


src/term/mod.rs
```diff
@@ -786,3 +789,9 @@ impl Deref for Name {
     self.0.deref()
   }
 }
+
+impl Book {
+  pub fn entrypoint(&self) -> String {
```
developedby, line 794: This is the hvmc entrypoint right? Also it's a bit confusing to call it the same name as one of the fields of `Book`, so let's change it to `Book::hvmc_entrypoint`.

developedby (changes requested):
Please add this new option to the docs

developedby:
> Please add this new option to the docs

Also I think we could change `toggleable passes` in the docs to `compiler options` or something of that sort.

docs/compiler-options.md
```diff
@@ -13,6 +13,7 @@
 | `-Osimplify-main` `-Ono-simplify-main` | Disabled | [simplify-main](#simplify-main) |
 | `-Omerge` `-Ono-merge` | Disabled | [definition-merging](#definition-merging) |
 | `-Oinline` `-Ono-inline` | Disabled | [inline](#inline) |
+| `-e` `--entrypoint` | `Main \| main` | Use another definition as entrypoint, rather than `main` or `Main`. |
```
developedby, line 16: I think it's best to expand this into it's own section so that we can explain the compiled output difference between having a custom entry point and not having one.

developedby (changes requested):

docs/compiler-options.md
```diff
@@ -230,3 +231,27 @@ main = (id foo)
...
+If given the option, use another definition as entrypoint rather than `main` or `Main`.
+
+> HVM-Lang finds by default for a `main` or `Main` as entrypoint to a program, but it is possible to use a different entrypoint with the `-e --entrypoint` option.
+
+Example:
+```
+// program
+Main = (λx x λx x)
+
+// compilation output, Main or main becomes a @main definition
```
developedby, line 246:
```suggestion
// compilation output, both `Main` and `main` compile to a `@main` definition
```

docs/compiler-options.md
```diff
@@ -230,3 +231,27 @@ main = (id foo)
...
+@main = a
+& (b b) ~ ((c c) a)
+
+// program
+// should fail as the program does not have a `main` definition.
```
developedby, line 251:
```suggestion
// Normally would fail as the program does not have a `main` definition.
```

docs/compiler-options.md
```diff
@@ -230,3 +231,27 @@ main = (id foo)
...
+run = (λx x λx x)
+
+// compilation output, --entrypoint run.
```
developedby, line 254:
```suggestion
// compilation output using `--entrypoint run`.
// No `@main` definition is created.
```

developedby (changes requested):

developedby (approved):

LunaAmora (approved):

PR #195, opened 2024-02-16
[sc-402] Refactor var-use counting when linearizing

LunaAmora:


imaqtkatt (approved):

src/term/transform/linearize.rs
```diff
@@ -277,6 +173,7 @@ fn duplicate_lam(nam: &mut Option<Name>, nxt: &mut Term, uses: Val) {
   }
 }
 
+#[allow(dead_code)]
```
developedby, line 176: Why not remove this?

src/term/transform/linearize.rs
```diff
@@ -289,6 +186,7 @@ fn internal_dup_name(nam: &Name, uses: Val) -> Name {
   format!("{}_dup", dup_name(nam, uses)).into()
 }
 
+#[allow(dead_code)]
```
developedby, line 189: also this

developedby (changes requested):

developedby (approved):

PR #197, opened 2024-02-20
[sc-450] Refactor error handling

LunaAmora:
Adds the option to: 
- handle multiple errors
- emit non-fatal errors (continues until a fatal error is found, or the end of `desugar_book`)

Bike-shedding and opinions on what passes can be fatal/non-fatal wanted


src/lib.rs
```diff
@@ -25,40 +27,31 @@ pub use term::load_book::load_file_to_book;
...
-pub fn desugar_book(
-  book: &mut Book,
-  opts: CompileOpts,
-  entrypoint: Option<Name>,
-) -> Result<(Name, Vec<Warning>), String> {
-  let mut warnings = Vec::new();
-  let main = book.check_has_entrypoint(entrypoint)?;
-  book.check_shared_names()?;
+pub fn desugar_book(book: &mut Book, opts: CompileOpts) -> Result<Name, String> {
+  let main = book.check_has_entrypoint();
```
imaqtkatt, line 50: This function should mutate the `Book.entrypoint` field instead of returning the entrypoint name, the other passes that need the `book.hvmc_entrypoint()` won't work because it will be None

LunaAmora: but this is working with custom entrypoints already. Wdym?

src/diagnostics.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(Debug, Clone)]
+pub enum Error {
+  Exhaustiveness(Name, ExhaustivenessErr),
+  MainRef(Name, ReferencedMain),
+  AdtMatch(Name, MatchError),
+  UnboundVar(Name, UnboundVar),
+  UnboundCtr(Name, UnboundCtr),
+  Infer(Name, InferErr),
+  Arity(Name, ArityErr),
+  Clyclic(ClyclicDef),
```
developedby, line 53: Cyclic

src/lib.rs
```diff
@@ -25,40 +27,31 @@ pub use term::load_book::load_file_to_book;
...
 
-pub fn desugar_book(
-  book: &mut Book,
-  opts: CompileOpts,
-  entrypoint: Option<Name>,
-) -> Result<(Name, Vec<Warning>), String> {
-  let mut warnings = Vec::new();
-  let main = book.check_has_entrypoint(entrypoint)?;
-  book.check_shared_names()?;
+pub fn desugar_book(book: &mut Book, opts: CompileOpts) -> Result<Name, String> {
```
developedby, line 49: Wouldn't it make more sense to return the errors here and call the stringification in the CLI?
Since we're already introducing the error types we might as well expose them for whoever wants to use them

developedby: Or rather the set of diagnostics `Info`

src/term/mod.rs
```diff
@@ -31,6 +31,8 @@ pub struct Book {
 
   /// A custom or default "main" entrypoint.
   pub entrypoint: Option<Name>,
+
+  pub info: Info,
```
developedby, line 35: Although this is part of the compilation context, i don't think it's part of the internal representation of the program.

I understand that this makes all the passes simpler to write and call, but maybe there's a better alternative.

src/term/check/has_entrypoint.rs
```diff
@@ -1,32 +1,89 @@
...
+}
 
 impl Book {
-  pub fn check_has_entrypoint(&mut self, nam: Option<Name>) -> Result<Name, String> {
-    if let Some(nam) = nam {
-      match self.defs.get(&nam) {
-        Some(_) => {
-          self.entrypoint = Some(nam.clone());
-          return Ok(nam);
+  pub fn check_has_entrypoint(&mut self) -> Option<Name> {
```
developedby, line 32: Adding to what Kate said, I think this would be better as a transformation pass `set_entrypoint`, where we set `Book.entrypoint` to either `main`, `Main` or some custom entrypoint depending on the options and the function definitions (or None in case of error i guess).

src/lib.rs
imaqtkatt: Oh ok, didn't notice that the entrypoint is being setted directly in main

developedby (changes requested):

src/term/mod.rs
LunaAmora: Maybe having an struct `Ctx { Book, Info }`?

developedby: Could be. You'll have to try it out and see how it turns outs.

src/term/check/has_entrypoint.rs
LunaAmora: What about now? or would it still be better to not set the `book.entrypoint` field on `main.rs` at all and just pass around as an arg until the `set_entrypoint` is called?

src/main.rs
```diff
@@ -146,29 +149,33 @@ fn main() {
     #[cfg(not(feature = "cli"))]
     compile_error!("The 'cli' feature is needed for the hvm-lang cli");
 
-    let cli = Cli::parse();
+    let mut cli = Cli::parse();
     let arg_verbose = cli.verbose;
+    let entrypoint = cli.entrypoint.take();
+
+    let book = |path: &Path| {
```
developedby, line 156: If this is the only place we do this, why not just put it inside `execute_cli_mode` instead of passing a function?

developedby (changes requested):

src/term/check/has_entrypoint.rs
developedby: They way you did it is good

src/diagnostics.rs
```diff
@@ -0,0 +1,118 @@
...
+  fn from(value: &str) -> Self {
+    Info::from(value.to_string())
+  }
+}
+
+// TODO: Merge errors that reference the same definition
+#[derive(Debug, Clone)]
+pub enum Error {
+  Exhaustiveness(Name, ExhaustivenessErr),
+  MainRef(Name, ReferencedMain),
```
developedby, line 61: Can you add `Err` to all these error type names?
`ReferencedMainErr`, `UnboundVarErr`, etc.
When a function returns `ReferencedMain` it's a bit confusing without context

developedby (changes requested):

imaqtkatt (approved):

src/term/check/set_entrypoint.rs
```diff
@@ -0,0 +1,87 @@
...
+      EntryErr::Arguments => write!(f, "Main definition can't have any arguments"),
+    }
+  }
+}
+
+impl Ctx {
+  pub fn set_entrypoint(&mut self) {
+    let mut main = None;
+
+    match self.book.get_possible_entry_points() {
```
developedby, line 35: Can you change this into
```
let (custom, main, hvm1_main) = self.book.get_possible_entry_points();
match (custom, main, hvm1_main) {
```
It's a bit confusing without seeing the order

developedby (changes requested):

tests/snapshots/desugar_file__non_exaustive_limit.hvm.snap
```diff
@@ -0,0 +1,12 @@
...
+source: tests/golden_tests.rs
+input_file: tests/golden_tests/desugar_file/non_exaustive_limit.hvm
+---
+In definition 'Bar':
+  Non-exhaustive pattern. Hint:
+    (Bar A A B) not covered.
+    (Bar A A C) not covered.
+    (Bar A A D) not covered.
+    (Bar A A E) not covered.
+    (Bar A A F) not covered. ...
```
developedby, line 11: Would be nice if it said something like "and others. Use the --verbose option to see all cases"

developedby (changes requested):

src/term/check/exhaustiveness.rs
```diff
@@ -1,19 +1,55 @@
...
+    let ident = ERR_INDENT_SIZE * 2;
+    let hints =
+      self.1.iter().take(limit).map(|pat| format!("{:ident$}({} {pat}) not covered.", "", self.0)).join("\n");
+
+    let mut str = format!("Non-exhaustive pattern. Hint:\n{}", hints);
 
-impl Book {
+    let len = self.1.len();
+    if len > limit {
+      str.push_str(&format!(" ... and {} others\n{:ident$}{}", len - limit, "", ERROR_LIMIT_HINT))
```
developedby, line 26:
```suggestion
      str.push_str(&format!(" ... and {} others.\n{:ident$}{}", len - limit, "", ERROR_LIMIT_HINT))
```

developedby (changes requested):

src/term/check/unbound_vars.rs
```diff
@@ -1,114 +1,143 @@
...
+  Local(Name),
+  Global { var: Name, declared: usize, used: usize },
+}
+
+impl Display for UnboundVarErr {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    match self {
+      UnboundVarErr::Local(var) => write!(f, "Unbound variable '{var}'"),
+      UnboundVarErr::Global { var, declared, used } => match (declared, used) {
+        (0, _) => write!(f, "Unbound unscoped variable '${var}'"),
```
developedby, line 22: Add `.` at the end of these

developedby (changes requested):

developedby (approved):

imaqtkatt (approved):

PR #203, opened 2024-02-22
Update cargo.lock,  remove unused snap

LunaAmora:


developedby (approved):

developedby:
cargo.lock is still there in master

LunaAmora:
> cargo.lock is still there in master

I wanted to say it was not updated

PR #205, opened 2024-02-23
[sc-468] pre-reduction with hvmc builtins is broken

imaqtkatt:


tjjfvi (approved):

FranchuFranchu (approved):

developedby (approved):
I'll trust the test, I don't understand the difference

tjjfvi:
The old code was casting an unknown `Def<T>` into a `Def<&Def<InertDef>>` (which it would never be); this now casts it to a `Def<InertDef>` (which is correct).

PR #207, opened 2024-02-26
[sc-470] Add missing escape sequences

imaqtkatt:


src/term/parser/lexer.rs
```diff
@@ -31,15 +31,15 @@ pub enum Token {
   #[regex("[0-9][0-9a-zA-Z_]*", |lex| from_radix(10, lex))]
   Num(u64),
 
-  #[regex(r#""([^"\\]|\\[tun"\\])*""#, |lex| normalized_string(lex).ok())]
-  #[regex(r#"`([^`\\]|\\[tun`\\])*`"#, |lex| normalized_string(lex).ok())]
+  #[regex(r#""([^"\\]|\\[0tunr'"\\])*""#, |lex| normalized_string(lex).ok())]
+  #[regex(r#"`([^`\\]|\\[0tunr`"\\])*`"#, |lex| normalized_string(lex).ok())]
```
developedby, line 35: Here you define the \` escape sequence, but you don't process it in the function below and don't test it

developedby (changes requested):

developedby (approved):

PR #212, opened 2024-02-27
[sc-475] Add documentation for new match options

imaqtkatt:


README.md
```diff
@@ -130,6 +130,53 @@ data List = (List.cons head tail) | (List.nil)
...
+  4:  "four"
+  _:  "other"
+}
+```
+
+Which is the equivalent of nesting match terms:
+```rs
+match 4 {
+  0: "zero"
+  1+a: match (- a 4) {
```
developedby, line 165:
```suggestion
  1+a: match (- (+ a (+ 0 1)) 5) {
```
Or something similar to show what is the logic behind this match

README.md
```diff
@@ -130,6 +130,53 @@ data List = (List.cons head tail) | (List.nil)
 ListEx2 = (List.cons 1 (List.cons 2 (List.cons 3 List.nil)))
 ```
 
+Match different kinds of terms, both matches are equivalent:
```
developedby, line 133: "It's possible to pattern match on different kinds of terms. These three forms are equivalent" or similar

developedby (changes requested):
Can you also fix the cspell warnings?
Either by adding `nums` to the dictionary or by changing it to `numbers`

README.md
```diff
@@ -130,6 +130,53 @@ data List = (List.cons head tail) | (List.nil)
...
+  1+a: match (- a 4) {
+    0: "five"
+    _:  ...
+  }
+}
+```
+
+Match multiple terms:
+```rs
+match True, True {
```
developedby, line 174: Can you make this example also be nested?
```
@a @b match a, b {
  (Some True) (x, y): (Some (x, y))
  (Some False) (x, y): (Some (y, x))
  None *: None
}
```

developedby (changes requested):
Could you make

developedby (approved):

docs/native-numbers.md
```diff
@@ -60,14 +60,10 @@ fibonacci = λn // n is the argument
...
-    +a:
-    match a {
-      // If the predecessor is 0, then return 1
-      0: 1
-      // Otherwise, bind n-2 to `b` and return the sum of (fib n-1) and (fib n-2)
-      +b: (+ (fibonacci a) (fibonacci b))
-    }
+    // If the number is 1, then return 1
+    1: 1
+    // Otherwise, and return the sum of (fib (n-2 + 1)) and (fib n-2)
```
LunaAmora, line 65: `Otherwise, return ...`

docs/native-numbers.md
```diff
@@ -60,14 +60,10 @@ fibonacci = λn // n is the argument
...
+    2+: (+ (fibonacci (+ n-2 1)) (fibonacci n-2))
```
LunaAmora, line 66: Are the names like `n-2` still binded automatically? If so, should probably still say it here. (unless another doc is saying that and i didn't find)

LunaAmora (changes requested):

docs/native-numbers.md
imaqtkatt: If the successor pattern is `3+` then the bind will be `var-3` and so on

README.md
```diff
@@ -130,6 +130,54 @@ data List = (List.cons head tail) | (List.nil)
...
 ```
 
+It's possible to match different kinds of terms. These three forms are equivalent:
+```rs
+match list {
+  (List.cons hd tl):  (Some hd)
+  List.nil:  None
+}
+
+match list {
```
developedby, line 140: can you add a comment here explaining that if we don't provide binds to the fields, then it uses implicit binding using `(Ctr arg.field1 arg.field2)`?

developedby (changes requested):

developedby (approved):

docs/native-numbers.md
LunaAmora: Yes, i know, but a doc should be saying that. The name `n-2` is just appearing magically without explanation

LunaAmora (approved):

PR #214, opened 2024-02-28
[sc-482] Add documentation on how pattern matching is encoded

imaqtkatt:


docs/pattern-matching.md
```diff
@@ -0,0 +1,152 @@
...
+Foo = @arg1 @arg2 @arg3 match arg1 arg2 arg3 {
+  (Foo 0 false (Cons h1 (Cons h2 t))): (A h1 h2 t)
+  (Foo 0 * *): B
+  (Foo 1+n false *): n
+  (Foo 1+n true *): 0
+}
+```
+
+Pattern matching on numbers has two forms.
+With the successor pattern it will expect a sequence of numbers up to the `n+var` pattern:
```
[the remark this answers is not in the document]
developedby: If the match is of NumSucc(n) type, we need to cover every case up to n, and can't cover a case greater or equal to n with a Num pattern.

Some examples to show what's ok and what's not:
```
match n { 0: ; 2+: ; } // Err, doesn't cover 1
match n { 0: ; 1+: ; 2+: ; } // Err, more than one value of + pattern
match n { 0: ; 1+: ; 1+: ; } // Ok, only one vaue of + pattern
match n { 0: ; 10+: ; n: ; } // Ok, 1 through 9 will go to the default
match n { n: ; 10+: ; } // Ok, all cases will go to 'n', compile to a Var match
match n { 10+: ; n: ; } // Ok, 0 through 9 go to 'n'. Very inneficient.
match n { 0: ; 1: ; 1+: ; } // Err, nums >= 1 can't be covered by a Num pattern
```

We should encourage users to avoid doing strange things with this type of match since it's not very intuitive.

docs/pattern-matching.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+Match on tuples become let tuple destructors, which are compiled to a native tuple destructor in hvm-core.
+```rust
+match x {
+  (f, s): s
+}
+
+// Becomes:
+let (f, s) = (1, 2); s
```
developedby, line 68:
```suggestion
let (f, s) = x; s
```

docs/pattern-matching.md
```diff
@@ -0,0 +1,152 @@
...
+```rust
+match x {
+  c: (Some c)
+}
+
+// Becomes:
+let c = x; (Some c)
+```
+
+Pattern matching on strings and lists desugars to a list of matches on List/String.cons and List.String.nil
```
developedby, line 81:
```suggestion
Pattern matching on strings and lists desugars to a list of matches on List/String.cons and List/String.nil
```

docs/pattern-matching.md
```diff
@@ -0,0 +1,152 @@
...
+// Becomes:
+Hi (String.cons 'h' (String.cons 'i' String.nil)) = 2
+Hi _ = 0
+
+Foo List.nil = 0
+Foo (List.cons x List.nil) = x
+Foo _ = 3
+```
+
+Match on ADT constructors can change based on the current encoding.
```
developedby, line 99:
```suggestion
Matches on ADT constructors are compiled to different expressions depending on the chosen encoding.
```

docs/pattern-matching.md
```diff
@@ -0,0 +1,152 @@
...
+```rust
+data Tree = (Leaf value) | (Node left right)
+
+match tree {
+  (Node (Leaf a) (Leaf b)): (Combine a b)
+  (Node left right): (Merge left right)
+  Leaf: (Single tree.value)
+}
+
+// Becomes:
```
developedby, line 141:
```suggestion
// Becomes:
match tree {
  (Node left right): match left {
    (Node left.left left.right):
      let left = (Node left.left left.right);
      (Merge left right)
    (Leaf a): match right {
      (Node right.left right.right):
        let left = (Leaf a);
        let right = (Node right.left right.right); 
        (Merge left right)
      (Leaf b): (Combine a b)
    }
  }
  (Leaf value): (Single value)
}
  
// Which is roughly equivalent to
```

developedby (changes requested):

docs/pattern-matching.md
```diff
@@ -0,0 +1,182 @@
...
+  (Foo 1+n false *): n
+  (Foo 1+n true *): 0
+}
+```
+
+Pattern matching on numbers has two forms.
+With the successor pattern `n+var` it will expect to cover every case up to `n`:
+
+```rust
+// Error: Case '2' not covered.
```
developedby, line 26: I think you should put the correct example first

developedby (changes requested):

LunaAmora (approved):

developedby (approved):

PR #217, opened 2024-02-29
[sc-483] Add an option to pass arguments to hvm-lang programs

imaqtkatt:


src/lib.rs
```diff
@@ -95,13 +95,18 @@ pub fn create_host(book: Arc<Book>, labels: Arc<Labels>, compile_opts: CompileOp
...
-  compile_book(book, CompileOpts::light())?;
+  compile_book(book, None, CompileOpts::light())?;
   Ok(())
 }
 
-pub fn compile_book(book: &mut Book, opts: CompileOpts) -> Result<CompileResult, Info> {
-  let warns = desugar_book(book, opts)?;
+pub fn compile_book(
+  book: &mut Book,
+  args: Option<Vec<Term>>,
```
developedby, line 104: For compiling programs that receive arguments, we want to compile down to an hvm-core program that receives arguments instead of passing the arguments as compile-time constants like this.

For running programs with hvm-lang the current approach is good

src/lib.rs
```diff
@@ -112,11 +117,17 @@ pub fn compile_book(book: &mut Book, opts: CompileOpts) -> Result<CompileResult,
   Ok(CompileResult { core_book, labels, warns })
 }
 
-pub fn desugar_book(book: &mut Book, opts: CompileOpts) -> Result<Vec<Warning>, Info> {
+pub fn desugar_book(
+  book: &mut Book,
+  args: Option<Vec<Term>>,
```
developedby, line 122: Same here, desugar shouldn't receive arguments but instead just desugar main to a match or something

src/lib.rs
```diff
@@ -184,8 +195,9 @@ pub fn run_book(
   run_opts: RunOpts,
   warning_opts: WarningOpts,
   compile_opts: CompileOpts,
+  args: Option<Vec<Term>>,
```
developedby, line 198: Can you put this function argument in the same position as you used for the others?

developedby (changes requested):

src/term/check/set_entrypoint.rs
```diff
@@ -23,13 +22,17 @@ impl Display for EntryErr {
...
       }
       EntryErr::MultipleRules => write!(f, "Main definition can't have more than one rule."),
-      EntryErr::Arguments => write!(f, "Main definition can't have any arguments."),
     }
   }
 }
 
 impl Ctx<'_> {
   pub fn set_entrypoint(&mut self) {
+    // already setted
```
developedby, line 31:
```suggestion
    // already set
```

src/term/transform/apply_args.rs
```diff
@@ -0,0 +1,27 @@
+use crate::term::{Book, Pattern, Term};
+
+impl Book {
+  pub fn apply_args(&mut self, args: Option<Vec<Term>>) -> Result<(), String> {
```
developedby, line 4: We don't want to return a string, but rather an Info on errors. That way we don't lose previous errors.

Check how other passes emit fatal errors.

developedby (changes requested):

src/lib.rs
```diff
@@ -184,7 +185,11 @@ pub fn run_book(
...
 ) -> Result<(Term, RunInfo), Info> {
+  let mut ctx = Ctx::new(&mut book);
+  ctx.set_entrypoint();
```
LunaAmora, line 191: Why set the entrypoint here and them again inside desugar_book?

imaqtkatt: If I don't set the entrypoint earlier in the run mode it won't be able to apply the cli arguments

LunaAmora: I see, idk how to fix that is_some() problem then.

maybe the args could be passed to be compiled too? That way set_entrypoint and apply_args would only be called one time inside the compile_book/desugar

LunaAmora: Idk if accepting args on compile mode would be something we want, but is there a problem with that?

src/term/check/set_entrypoint.rs
```diff
@@ -23,13 +22,17 @@ impl Display for EntryErr {
...
+    if self.book.entrypoint.is_some() {
```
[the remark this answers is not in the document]
imaqtkatt: I only added this line because it will fail in the run mode.
Also, I don't see any reason to check for Main or main when there is an custom entrypoint.

src/lib.rs
imaqtkatt: If we are compiling we just desugar the patterns, the args are be passed on hvm-core side

src/term/check/set_entrypoint.rs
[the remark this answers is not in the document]
LunaAmora: Why would it fail?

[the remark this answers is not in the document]
imaqtkatt: The logic is wrong, it shows me 'File has both 'Main' and 'Main' definitions.'

src/lib.rs
developedby: I think she meant having the compile/desugar function accept arguments, but the cli then just passes None.

run_book -> args = Some(vec![])
compile/desugar_book -> args = None

developedby (changes requested):

developedby (approved):

LunaAmora (approved):

PR #218, opened 2024-03-01
[sc-485] Error printing not correctly breaking lines

imaqtkatt:
Because `join` method does not insert a newline at the end

### DIFF
--- a/src/term/transform/resolve_refs.rs
+++ b/src/term/transform/resolve_refs.rs
@@ -124,7 +124,6 @@ impl Term {
         }
         Term::Lst { .. } => unreachable!("Should have been desugared already"),
         Term::Lnk { .. } | Term::Ref { .. } | Term::Num { .. } | Term::Str { .. } | Term::Era | Term::Err => {
-          ()
         }
       }
       Ok(())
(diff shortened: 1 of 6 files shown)


src/term/transform/resolve_refs.rs
```diff
@@ -124,7 +124,6 @@ impl Term {
         }
         Term::Lst { .. } => unreachable!("Should have been desugared already"),
         Term::Lnk { .. } | Term::Ref { .. } | Term::Num { .. } | Term::Str { .. } | Term::Era | Term::Err => {
-          ()
```
developedby, line 127: You should run cargo fmt and cargo clippy

developedby (changes requested):

src/term/transform/resolve_refs.rs
imaqtkatt: I already did it

developedby: Should cargo format be changing this into
```
erm::Lnk { .. } | Term::Ref { .. } | Term::Num { .. } | Term::Str { .. } | Term::Era | Term::Err => {}
```
Without the line break?

imaqtkatt: I think it should, but it does not format this way because we are using `max_width = 110`

developedby (approved):

PR #219, opened 2024-03-01
[sc-479] Update encodings to make use of hvmc n-ary nodes. Add n-ary tups/dups

developedby:


imaqtkatt (approved):

src/term/transform/simplify_matches.rs
```diff
@@ -43,21 +43,20 @@ impl Term {
       match self {
         Term::Mat { args, rules } => {
           let (new_args, extracted) = extract_args(args);
+          let rules = std::mem::take(rules);
 
-          let term = simplify_match_expression(new_args, std::mem::take(rules), ctrs, adts)?;
-          *self = bind_extracted_args(extracted, term);
+          *self = simplify_match_expression(new_args, rules, ctrs, adts)?;
```
LunaAmora, line 48: Why do this extra assignment just to mem::take it out at the next line?

LunaAmora (approved):

src/term/transform/simplify_matches.rs
developedby: I'm not sure what you mean by this.

`simplify_match_expression` sometimes creates match expressions with non-Var arguments, so I have to extract the arguments and then bind them to generated variables, just like we do in that other pass.
This is the reason why I have these two lines.

Also, I'm not mem::take-ing anything here other than `self.rules` which must be a mut ref since we can't destructure self because it manually implements Drop;

LunaAmora:
```rs
*self = simplify_match_expression(new_args, rules, ctrs, adts)?;
*self = bind_extracted_args(extracted, std::mem::take(self));
```
the second line

developedby: Oh i just got confused by the way github was highlighting the changes, now I see it.
This is just a mistake from rebasing

PR #220, opened 2024-03-04
[sc-490] Remove restriction of types not sharing names with constructors

imaqtkatt:


src/term/check/shared_names.rs
```diff
@@ -14,26 +17,70 @@ impl Display for TopLevelErr {
...
+impl NameInfo {
+  fn new(kind: NameKind) -> Self {
+    Self { kind, count: 0 }
+  }
+}
+
+impl NameInfo {
+  fn with_ctr(&mut self, current_name: &Name, info: &mut diagnostics::Info) {
+    match self.kind {
+      NameKind::Adt => {} // Catch by the parser
```
LunaAmora, line 65: *Caught

i think "// Error caught by the parser" would be a little better

LunaAmora (approved):

tests/golden_tests/compile_file/error_messages.hvm
```diff
@@ -1,5 +1,6 @@
 data A = (A)
 data B = (B)
+data C = (B)
```
developedby, line 3: Can you also test that this fails? (maybe in another file if the message doesn't show here.
```
data A = A
A = 0

main = A
```

developedby (changes requested):

PR #221, opened 2024-03-05
[sc-489] Simplify passes by using iterators over the AST

developedby:
If you have any suggestion on how to implement the mutable iterators, I'll be happy to hear them.

### DIFF
--- a/src/term/mod.rs
+++ b/src/term/mod.rs
@@ -1,4 +1,4 @@
-use self::{check::type_check::infer_match_arg_type, parser::lexer::STRINGS};
+use self::{check::type_check::infer_match_arg_type, parser::lexer::STRINGS, util::ChildrenIter};
 use crate::{diagnostics::Info, term::builtins::*, ENTRY_POINT};
 use indexmap::{IndexMap, IndexSet};
 use interner::global::GlobalString;
@@ -17,6 +17,7 @@ pub mod net_to_term;
 pub mod parser;
 pub mod term_to_net;
 pub mod transform;
+mod util;
 
 pub use net_to_term::{net_to_term, ReadbackError};
 pub use term_to_net::{book_to_nets, term_to_compat_net};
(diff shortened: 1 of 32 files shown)


LunaAmora (approved):

imaqtkatt (approved):

LunaAmora (approved):

src/term/mod.rs
```diff
@@ -398,150 +399,275 @@ impl Term {
...
-      // Only substitute scoped variables.
-      Term::Chn { bod, .. } => bod.subst(from, to),
-      Term::Lnk { .. } => (),
-      Term::Let { pat, val, nxt } => {
-        val.subst(from, to);
-        if !pat.binds().contains(from) {
-          nxt.subst(from, to);
-        }
+      Term::Mat { args, rules } => {
+        ChildrenIter::Many(Box::new(args.iter().chain(rules.iter().map(|r| &r.body))))
```
tjjfvi, line 406: Boxing these iterators is unnecessary overhead, and the `dyn` will prevent rust from optimizing the iteration. Is there a reason you can't use the macro I wrote, with the generics? It would let you avoid boxing these iterators.

tjjfvi (changes requested):

imaqtkatt (approved):

src/term/mod.rs
developedby: I don't like how the macro approach created a bunch of opaque structures.
It was hard to work with as I tried to extend them to do more things (which I ultimately failed to do in a simple way) and I got errors from things created in the macro.

I figured since performance is not a huge concern for hvm-lang right now I'd leave it like this which will probably be easier for me to try to improve in the future.

src/term/util.rs
```diff
@@ -0,0 +1,56 @@
+// TODO: Find a way to implement Clone for ChildrenIter
+
+/// Iterator for tree children
+pub enum ChildrenIter<'a, T: 'a> {
+  Zero(std::array::IntoIter<T, 0>),
+  One(std::array::IntoIter<T, 1>),
+  Two(std::array::IntoIter<T, 2>),
+  Many(Box<dyn DoubleEndedIterator<Item = T> + 'a>),
+}
```
tjjfvi, line 9: This reminds me of: https://xkcd.com/764/

Primitive cultures develop ~~sesame street~~ iterators 😆

LunaAmora (approved):

PR #223, opened 2024-03-06
[sc 492] Re-add repeated bind in match-arm/rule patterns warning

LunaAmora:
This depends on #221, leaving as a draft until that is merged

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -1,7 +1,7 @@
 use crate::term::{
   check::{
-    set_entrypoint::EntryErr, shared_names::TopLevelErr, unbound_pats::UnboundCtrErr,
-    unbound_vars::UnboundVarErr,
+    repeated_bind::RepeatedBindWarn, set_entrypoint::EntryErr, shared_names::TopLevelErr,
+    unbound_pats::UnboundCtrErr, unbound_vars::UnboundVarErr,
   },
   display::DisplayFn,
   transform::{
@@ -23,7 +23,7 @@ pub struct Info {
   err_counter: usize,
   errs: Vec<Error>,
   errs_with_def: BTreeMap<Name, Vec<Error>>,
-  pub warns: Vec<Warning>,
+  pub warns: Warnings,
 }
 
 impl Info {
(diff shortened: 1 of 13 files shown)


src/diagnostics.rs
```diff
@@ -23,7 +23,7 @@ pub struct Info {
   err_counter: usize,
   errs: Vec<Error>,
```
developedby, line 24: When I read this it's a bit confusing if errs_with_def copies errs or not.
I don't know what better name we could give these 2 fields, but I think it could be improved

developedby (changes requested):

src/diagnostics.rs
LunaAmora: maybe program_errors & rule_errors?

imaqtkatt (approved):

src/term/check/repeated_bind.rs
```diff
@@ -0,0 +1,70 @@
...
+pub enum RepeatedBindWarn {
+  Rule(Name),
+  Match(Name),
+}
+
+impl Display for RepeatedBindWarn {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    match self {
+      RepeatedBindWarn::Rule(bind) => write!(f, "Repeated bind inside rule pattern: '{bind}'."),
+      RepeatedBindWarn::Match(bind) => write!(f, "Repeated bind inside match arm: '{bind}'."),
```
developedby, line 14: Actually, can we add an explanation of what this means here? Something like "The last variable will shadow all the others" or "Only the last bind will be available"

developedby (changes requested):
Looks good overall

src/term/check/repeated_bind.rs
LunaAmora: I don't know why, but actually its only the first one that is binded

LunaAmora: ex:
```
Foo a a a = a
main = (Foo 1 2 3)
```
returns `1`

developedby: Maybe we can leave it like this here and fix and improve the message in another PR, when it's correct.

LunaAmora: > Yeah, I'm pretty sure it's a bug, since `(Foo a b c) = ...` should desugar to `Foo = @a @b @c ...` and the same for more complex matches

but it does desugar to that, the problem is only that the var binds to the first:
```
(Foo) = λa λb λc a
```

developedby: But if we have `(Foo a a a) = a` it'll desugar to `Foo = @a @a @a a`, the last lambda sugars the previous ones

LunaAmora: Ah, I see what you mean

developedby: The problem has to do with the matches, since it actually desugars to `(Foo) = λ%x0 λ%x1 λ%x2 match %x0, %x1, %x2 { a a a: a }`

developedby (approved):

PR #227, opened 2024-03-07
Fix pr merge error

LunaAmora:


tjjfvi:
might make sense to use [merge queues](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue); they test the merge commit before pushing it to main, preventing situations like this

(we're using them on hvm-core)

PR #229, opened 2024-03-08
[sc-502] Allow number of CLI arguments to be different than declared in main

imaqtkatt:


developedby (changes requested):
Can you also update the documentation?

developedby (approved):

PR #232, opened 2024-03-12
[sc-500] Add mutual recursion cycle check

imaqtkatt:
Infinite expansion check will be added in a later moment


docs/lazy-definitions.md
```diff
@@ -36,17 +34,23 @@ This code will work as expected, because `Nat.add` is unrolled lazily only when
 
 ### Automatic optimization
 
-HVM-lang carries out an optimization automatically:
+HVM-lang carries out a [float-combinators](compiler-options#float-combinators) optimization through the CLI, which is active by default in strict mode.
```
developedby, line 37:
```suggestion
HVM-lang carries out a [combinator floating](compiler-options#float-combinators) optimization, enabled through the CLI, which is active by default in strict mode.
```

developedby (changes requested):

src/hvmc_net/mutual_recursion.rs
```diff
@@ -0,0 +1,130 @@
...
+  let cycles = graph.cycles();
+
+  if cycles.is_empty() { Ok(()) } else { Err(pretty_print_cycles(&cycles)) }
+}
+
+fn pretty_print_cycles(cycles: &[Vec<Ref>]) -> String {
+  cycles
+    .iter()
+    .enumerate()
+    .map(|(i, cycles)| format!("Cycle {}: {}", 1 + i, cycles.join(" -> ")))
```
developedby, line 23: We need to improve this message.
It should be something like `Mutual recursion cycle detected in compiled functions. Cycle 1: A -> B -> ... -> A. This program will expand infinitely in strict evaluation mode. Read <link to docs> for more information.`

developedby (changes requested):

developedby (changes requested):
Can you add a test with multiple recursion cycles? A test where the cycles don't share any functions and a test where the cycles partially overlap

tests/snapshots/mutual_recursion__a_b_c.hvm.snap
```diff
@@ -0,0 +1,8 @@
+---
+source: tests/golden_tests.rs
+input_file: tests/golden_tests/mutual_recursion/a_b_c.hvm
+---
+Mutual recursion cycle detected in compiled functions:
+Cycle 1: A -> B -> C
```
developedby, line 6: Can you make it so that we get the first element at the end of the list as well? In this case `A -> B -> C -> A`

tests/golden_tests/mutual_recursion/multiple.hvm
```diff
@@ -0,0 +1,11 @@
+(A) = (B)
+(B) = (C)
+(C) = (A)
+
+(H) = (I)
+(I) = (H)
+
+(M) = (N)
```
developedby, line 8: Can you change this to have a cycle of length one? Like `(M) = (M)`. Might need to be something more complex not to get an error earlier, like `N = @x (x (N x))`

docs/lazy-definitions.md
```diff
@@ -19,34 +19,44 @@ main = (Nat.add (S (S (S Z))) (S Z))
 
 Because of that, is recommended to use [supercombinator](https://en.wikipedia.org/wiki/Supercombinator) formulation to make terms be unrolled lazily, preventing infinite expansion in recursive function bodies.
```
developedby, line 20:
```suggestion
Because of that, it's recommended to use a [supercombinator](https://en.wikipedia.org/wiki/Supercombinator) formulation to make terms be unrolled lazily, preventing infinite expansion in recursive function bodies.
```

developedby (changes requested):

developedby (approved):

PR #234, opened 2024-03-12
Remove expand in debug hook

imaqtkatt:


developedby (changes requested):
Can you add a test where this would fail?

developedby (approved):

PR #236, opened 2024-03-13
Implement remaining `hvm1` builtins.

FranchuFranchu:
I wrote implementations for `HVM.query`, `HVM.store`, and `HVM.load` using a new API for defs that exists in a hvm-core branch. I'm creating a PR mostly to get feedback on whether the API is high-level and practical enough for `hvm-lang`.

If it isn't, we can postpone this until we have an API in `hvm-core` that's higher-level. (it'd likely end up looking like `rlua`'s API, with lifetimed `Port`s which would allow us to make it completely safe).


developedby (changes requested):
Remove the local test files

developedby:
We got read/write from files and read/write from stdio. I think that's enough for most simple uses.
Maybe a panic/exit function in case something goes wrong?

developedby (approved):

PR #240, opened 2024-03-18
[sc-522] Update hvm-lang to make it compatible with hvm-core n-ary nodes.

FranchuFranchu:


tests/snapshots/readback_lnet__invalid_mat_mat.hvm.snap
```diff
@@ -2,8 +2,5 @@
 source: tests/golden_tests.rs
 input_file: tests/golden_tests/readback_lnet/invalid_mat_mat.hvm
 ---
-Warnings:
-During readback:
-  Invalid Numeric Match. (4 occurrences)
-
-λa match match <Invalid> { 0: 3; 1+*: 4 } { 0: <Invalid>; 1+: <Invalid> }
+Errors:
+invalid legacy match
```
developedby, line 6: Where is this error message comming from?

tjjfvi: https://github.com/HigherOrderCO/hvm-core/blob/main/src/ast.rs#L325

tjjfvi: `?<a b>` is no longer valid syntax: https://github.com/HigherOrderCO/hvm-lang/blob/be2dcf4203237f6ff2d22aa7c22a720fc07541f9/tests/golden_tests/readback_lnet/invalid_mat_mat.hvm#L3C17-L3C23

developedby (approved):

tests/snapshots/readback_lnet__invalid_mat_mat.hvm.snap
tjjfvi: correct; that test case no longer makes sense and can/should be removed

PR #241, opened 2024-03-18
[sc-519] Add syntax sugar for Nats

imaqtkatt:


developedby (changes requested):

tests/snapshots/encode_pattern_match__common.hvm.snap
```diff
@@ -15,6 +15,10 @@ TaggedScott:
 
 (Yellow) = #Light λ* #Light λb #Light λ* b
 
+(Red) = #Light λa #Light λ* #Light λ* a
```
developedby, line 18: Did we change something to cause these to have a different order? Might have something causing the output to not be deterministic across supposedly equivalent versions of the compiler.

tests/golden_tests/run_file/nat_add.hvm
```diff
@@ -0,0 +1,4 @@
+(Nat.add (Nat.zero)   x) = x
+(Nat.add (Nat.succ p) x) = (Nat.succ (Nat.add p x))
```
developedby, line 2: Can you add a test to resugar an invalid Nat.succ like `(Nat.succ 0)` and a Nat.zero in an unexpected position like `(Nat.zero foo)`

developedby (changes requested):

tests/snapshots/encode_pattern_match__common.hvm.snap
imaqtkatt: It is related with the `data Result` in the builtins.hvm file, without `Result` it outputs in correct order

developedby: I don't understand why it changes the order of other unrelated things. It should be in insertion order here, right? I would understand if it were after the merge pass, since it swaps things around.

developedby (approved):

PR #242, opened 2024-03-20
[sc-526] Improve deref loop warning in strict evaluation

imaqtkatt:


src/hvmc_net/mutual_recursion.message
```diff
@@ -0,0 +1,14 @@
+Seems like you're trying to run some recursive function(s) on strict-mode:
```
developedby, line 1: It's better than before, but I think it needs to be worded better.

First of all, in these cases we're 100% sure that this will result in an infinite loop and it is also very likely to allocate more and more memory.

If the float combinators optimization is not on, we should recommend activating it.

"Use combinators" is not great advice since a program can even be based only on supercombinators and still have an infinite deref loop.
The more correct version would be to "refactor the program to use lazy references", but this doesn't explain how to do it.
More concretely, the best advice is to write programs where we don't have a recursive reference in the head position of an application or in a duplication body.
In the case of applications, this rewrite can usually be achieved by lifting at least one of the  terms that contain one of the recursive calls in the chain (or making them into combinators which get automatically floated).
In the case of duplications (or pair decomposition for some cursed program using them to duplicate), we can usually fix this by manually duplicating the term with the recursive call or using a `use` expression.

Message 2 needs to containg clear and conscise version of this, and the documentation needs to be updated to clearly convey this information.

developedby (changes requested):

src/hvmc_net/mutual_recursion.message
```diff
@@ -0,0 +1,18 @@
...
+You have 2 options:
+
+1. Easy Fix: use lazy-mode.
+
+Just append the `-L` option to `HVM-Lang`, and it will run in lazy-mode. It has the advantage of not doing wasteful work, having an improved complexity class, and being compatible with unrestricted recursion! It has a small overhead though, and isn't compatible with GPU.
+
+2. Hard Fix: refactor the program to use lazy references.
+
+When a function reference is in head position of an application or is duplicated by the non-linear use of a `let` expression it will be greedly expanded, leading to an infinite loop.
+If the application is used written as a combinator, it will automatically lifted to a lazy reference, which usually breaks the recursion cycle. Alternatively, by using the built-in `data` and `match` syntax, hvm-lang will try to do this automatically.
```
developedby, line 16: Maybe we should put a very small example here? Something like
> (e.g. If pattern matching on scott-encoded ADTs, write `@a @x(x @a (Foo a) a)` instead of `@a @x(x (Foo a))`

src/hvmc_net/mutual_recursion.message
```diff
@@ -0,0 +1,18 @@
...
+For let expressions where the variable is non-linear (used more than once), you can instead employ `use` expressions to statically duplicate the offending recursive term.
```
developedby, line 17: Same here
> (e.g. write `Foo = @f use x = Foo; (f x x)` instead of `Foo = @f let x = Foo; (f x x)`

Although i'm still struggling to find a more meaningful example here.
`@x @f @a let a = (f a); (x a a)` is more realistic, but we need two steps of transformation.
First `@x @f @a use a = (f a); (x a a)`
Then `@x @f @a use b = @f @a (f a); (x b b f a)`

developedby (changes requested):

docs/lazy-definitions.md
```diff
@@ -1,42 +1,59 @@
 # Making recursive definitions lazy
 
-The HVM-Core is an eager runtime, for both CPU and parallel GPU implementations. Terms that use recursive terms will unroll indefinitely.
+In strict-mode, terms that use recursive terms will unroll indefinitely.
```
developedby, line 3:
```suggestion
In strict-mode, some types of recursive terms will unroll indefinitely.
```

src/hvmc_net/mutual_recursion.message
```diff
@@ -0,0 +1,22 @@
...
+
+1. Easy Fix: use lazy-mode.
+
+Just append the `-L` option to `HVM-Lang`, and it will run in lazy-mode. It has the advantage of not doing wasteful work, having an improved complexity class, and being compatible with unrestricted recursion! It has a small overhead though, and isn't compatible with GPU.
+
+2. Hard Fix: refactor the program to use lazy references.
+
+When a function reference is in head position of an application or is duplicated by the non-linear use of a `let` expression it will be greedly expanded, leading to an infinite loop.
+If the application is used written as a combinator, it will automatically lifted to a lazy reference, which usually breaks the recursion cycle. Alternatively, by using the built-in `data` and `match` syntax, hvm-lang will try to do this automatically.
+(e.g. If pattern matching on scott-encoded ADTs, write @a @x(x @a (Foo a) a) instead of @a @x(x (Foo a)))
```
developedby, line 17:
```suggestion
(e.g. If pattern matching on scott-encoded ADTs, write '@a @x(x @a (Foo a) a)' instead of '@a @x(x (Foo a))')
```

src/hvmc_net/mutual_recursion.message
```diff
@@ -0,0 +1,22 @@
...
+
+For let expressions where the variable is non-linear (used more than once), you can instead employ `use` expressions to statically duplicate the offending recursive term.
+(e.g. write Foo = @f use x = Foo; (f x x) instead of Foo = @f let x = Foo; (f x x))
```
developedby, line 20:
```suggestion
(e.g. write 'Foo = @f use x = Foo; (f x x)' instead of 'Foo = @f let x = Foo; (f x x)')
```

tests/snapshots/mutual_recursion__multiple.hvm.snap
```diff
@@ -3,10 +3,30 @@ source: tests/golden_tests.rs
...
 Errors:
-Mutual recursion cycle detected in compiled functions:
-  Cycle 1: A -> B -> C -> A
-  Cycle 2: H -> I -> H
-  Cycle 3: M -> M
-  Cycle 4: N -> N
-  This program will expand infinitely in strict evaluation mode.
-  Read https://github.com/HigherOrderCO/hvm-lang/blob/main/docs/lazy-definitions.md for more information.
+Seems like you're trying to run some recursive function(s) on strict-mode:
+A
```
developedby, line 7: This list of functions is a bit confusing.
How about
```
Seems like you're trying to run some recursive function(s) on strict-mode.
The following recursive cycles are not compatible with strict-mode the way these functions were written:
 * A -> B -> C -> A
 * ...

Due to the ultra-greedy nature of strict-mode, these cases will result in infinite loops.
...
```

docs/lazy-definitions.md
```diff
@@ -1,42 +1,59 @@
...
+Map = λf λlist
+  let cons = λx λxs λf (Cons (f x) (Map f xs))
+  let nil = λf Nil
+  (list cons nil f)
 ```
 
-This code will work as expected, because `Nat.add` is unrolled lazily only when it is used as an argument to a lambda.
+This code will work as expected, since `cons` and `nil` are lambdas without free variables, they will be automatically floated to new definitions if the [float-combinators](compiler-options#float-combinators) option is active, allowing them to be unrolled lazily by hvm.
+
+It's recommended to use a [supercombinator](https://en.wikipedia.org/wiki/Supercombinator) formulation to make terms be unrolled lazily, preventing infinite expansion in recursive function bodies.
```
developedby, line 57: We could add at the end something like 
```
If you have a set of mutually recursive functions, you only need to make one of the steps lazy. This might be useful when doing micro-optimizations, since it's possible to avoid part of the small performance cost of linearizing lambdas.
```

docs/lazy-definitions.md
```diff
@@ -1,42 +1,59 @@
 # Making recursive definitions lazy
 
-The HVM-Core is an eager runtime, for both CPU and parallel GPU implementations. Terms that use recursive terms will unroll indefinitely.
+In strict-mode, terms that use recursive terms will unroll indefinitely.
 
-This means that for example, the following code will hang, despite being technically correct and working on Haskell:
+This is a simple piece of code that works on many other functional programming languages, but hangs on HVM:
```
developedby, line 5:
```suggestion
This is a simple piece of code that works on many other functional programming languages, including hvm's lazy-mode, but hangs on strict-mode:
```

developedby (changes requested):

tests/snapshots/compile_file_o_all__repeated_name_trucation.hvm.snap
```diff
@@ -3,10 +3,30 @@ source: tests/golden_tests.rs
 input_file: tests/golden_tests/compile_file_o_all/repeated_name_trucation.hvm
 ---
 Warnings:
-Mutual recursion cycle detected in compiled functions:
-  Cycle 1: long_name_that_truncates -> long_name_that_truncates$C0 -> long_name_that_truncates
-  This program will expand infinitely in strict evaluation mode.
-  Read https://github.com/HigherOrderCO/hvm-lang/blob/main/docs/lazy-definitions.md for more information.
+Seems like you're trying to run some recursive function(s) on strict-mode.
+The following recursive cycles are not compatible with strict-mode the way these functions were written:
+  * long_name_that_truncates -> long_name_that_truncates$C0 -> long_name_that_truncates
```
developedby, line 8: Can we skip displaying generated functions? I think this will be confusing for users.

src/hvmc_net/mutual_recursion.message
```diff
@@ -0,0 +1,23 @@
...
+{cycles}
+
+Due to the ultra-greedy nature of strict-mode, that might result in infinite loops.
+If the float-combinators optimization is not on, we recommend activating it.
+
+You have 2 options:
+
+1. Easy Fix: use lazy-mode.
+
+Just append the `-L` option to `HVM-Lang`, and it will run in lazy-mode. It has the advantage of not doing wasteful work, having an improved complexity class, and being compatible with unrestricted recursion! It has a small overhead though, and isn't compatible with GPU.
```
developedby, line 12: The "!" bothers me a bit. I don't like when my terminal shouts at me

developedby (changes requested):

src/hvmc_net/mutual_recursion.rs
```diff
@@ -17,32 +17,36 @@ pub fn check_cycles(book: &Book, diagnostics: &mut Diagnostics) -> Result<(), Di
...
     .iter()
-    .enumerate()
-    .map(|(i, cycle)| {
-      let cycle_str = cycle.iter().chain(cycle.first()).cloned().collect::<Vec<_>>().join(" -> ");
-      format!("{:ERR_INDENT_SIZE$}Cycle {}: {}", "", 1 + i, cycle_str)
+    .take(5)
+    .map(|cycle| {
+      let cycle_str = cycle
+        .iter()
+        .filter(|nam| !nam.contains("$"))
```
developedby, line 37: I think this will break if a non-generated definition gets merged by the merge pass.
We could have `Foo_$_Bar -> Baz -> Foo_$_Bar`. And we should display it either as `Foo -> Baz -> Foo` or both cycles separately (i think showing just one is ok).
This is also an extremely rare case, so we could leave it to another task if this is hard to implement correctly

developedby (changes requested):

developedby (approved):

PR #244, opened 2024-03-25
[sc-527] Recursion check incorrectly displays some generated functions

imaqtkatt:


src/hvmc_net/mutual_recursion.rs
```diff
@@ -151,3 +153,29 @@ impl Debug for Graph {
     write!(f, "Graph{:?}", self.0)
   }
 }
+
+fn combinations_from_merges(cycle: Vec<Ref>) -> Vec<Vec<Ref>> {
+  let mut combinations: Vec<Vec<Ref>> = vec![vec![]];
+  for r#ref in cycle {
+    if let Some(index) = r#ref.find("_$_") {
```
developedby, line 160: Can you put _$_ in a constant that you load here from the merge module? I feel like we might change this in the future and it's best to avoid magic numbers

developedby (changes requested):

developedby (approved):

PR #245, opened 2024-03-27
[sc-525] Float combinators not extracting some terms

imaqtkatt:


src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,149 @@
...
-
-    if unscoped {
-      detach = detach & Detach::unscoped_lam(nam.cloned().unwrap());
-    } else {
-      term_info.provide(nam);
-    }
-
-    if depth != 0 && detach.can_detach() && term_info.has_no_free_vars() {
-      term_info.detach_term(self);
+  pub fn is_constant(&self) -> bool {
```
developedby, line 112: isn't `is_constant` already covered by `is_safe`?

developedby (changes requested):
Can you add a test for recursiveness/combinators in other positions? Like in different application positions, in adt matches and num matches, both in arms and as argument. In lists, tuples, sups, lets, dups, etc.

Also some positions where it should not get extracted, like the body of a lambda (eg `foo = @a @b @c (foo a)`)

Also, update the documentation for this pass to explain the new logic, including the exceptions where we don't float.

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,160 @@
...
-  /// Can be detached freely
-  Combinator,
-  /// Can not be detached
-  Unscoped { lams: BTreeSet<Name>, vars: BTreeSet<Name> },
-  /// Should be detached to make the program not hang
-  Recursive,
-}
+        match term {
+          // If it is an Application without free variables like '(bar 0)', float into a new definition.
+          Term::App { .. } => {
```
developedby, line 61: We don't need to handle app as an exception. All terms with children could have this logic of first trying to float the children and then trrying to float the term itself.

As an optimization we can consider an application with many arguments to be a single term with many children, as well as considering lambdas to have as children the children of the lambda body.
That would be only to avoid fragmenting the combinators too much, but the overall logic is the same.

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,160 @@
...
+            if term.free_vars().is_empty() && !term.has_unscoped_diff() {
+              float_combinator(def_name, name_gen, term, builtin, combinators);
+            }
+          }
 
-impl Detach {
-  fn can_detach(&self) -> bool {
-    !matches!(self, Detach::Unscoped { .. })
-  }
+          // If it is a Superposition, float every child element.
```
developedby, line 67: We also don't need Sup as an expection. (by the way, this case is already covered by the chidren loop above)

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,160 @@
...
-  Combinator,
-  /// Can not be detached
-  Unscoped { lams: BTreeSet<Name>, vars: BTreeSet<Name> },
-  /// Should be detached to make the program not hang
-  Recursive,
-}
+        match term {
+          // If it is an Application without free variables like '(bar 0)', float into a new definition.
+          Term::App { .. } => {
+            if term.free_vars().is_empty() && !term.has_unscoped_diff() {
```
developedby, line 62: You can call these two conditions `is_combinator`

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,160 @@
...
+      Term::Var { nam } if scope.contains(&nam) => true,
+      Term::Ref { .. } => true,
+      term => term.is_safe(book, seen),
     }
+  }
 
-    term_info.merge_scope(parent_scope);
-    detach
+  /// A term is a combinator if it is a lambda abstraction without free variables.
+  pub fn is_combinator(&self) -> bool {
```
developedby, line 152: A combinator is any term without free variables (+ free unscoped lambda abstractions in case of our modified lambda logic).

You shouldn't be floating only combinators that are abstractions or applications, but any combinator.

e.g.
```
@x (x match (Foo 0) {...})
```
The whole match expression could be extracted since it's a combinator.
I'm not sure how useful that is, but technically a combinator if all its children are combinators.

developedby (changes requested):

src/term/transform/float_combinators.rs
imaqtkatt: The match expressions are converted to lambdas before this pass, so it will probably be extracted

developedby: Native matches won't

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,151 @@
...
-        for (child, binds) in self.children_mut_with_binds() {
-          detach = detach & child.go_float(depth + 1, term_info);
-          for bind in binds {
-            term_info.provide(bind.as_ref());
-          }
+        seen.insert(nam.clone());
+        if let Some(definition) = book.defs.get(nam) {
+          definition.rule().body.is_safe(book, seen)
+        } else {
+          false
```
developedby, line 107: I think we'll never find a reference to one of the newly created definitions, so even throwing an error would be fine. But this is also ok

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,151 @@
...
-      }
-
-      _ => fun_detach & arg_detach,
-    };
-
-    term_info.merge_scope(fun_scope);
-    term_info.merge_scope(arg_scope);
-
-    detach
+  /// A term is a supercombinator if it is a lambda abstraction without free variables.
```
developedby, line 134: This is not quite the definition of a supercombinator. This is just a lambda abstraction that is also a combinator, which is all we need here in this case. To be a supercombinator we require that it is of the form
```
@x1 @x2 ... @xN (Arg1 Arg2 ... ArgM)
```
Where Arg1-M are also supercombinators, constants or one of the x1-N.

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,151 @@
...
-      detach = detach & arm_detach;
-    }
+  pub fn has_unscoped_diff(&self) -> bool {
+    let (declared, used) = self.unscoped_vars();
+    declared.difference(&used).count() != 0
+  }
 
-    term_info.merge_scope(parent_scope);
-    detach
+  fn is_combinator(&self) -> bool {
```
developedby, line 144: I think it's fine to consider as combinators other non-lambda terms, like matches, tups, sups, etc. Particularly, it would be useful to extract dups and lets, which can become redexes and so cause loop.
Concretely, changing `is_combinator` to just not having free variables or not having free variables and not having unmatched unscoped vars/binds.

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,151 @@
...
+    builtin: bool,
+    seen: &mut IndexSet<Name>,
+  ) {
+    for term in self.children_mut() {
+      // Recursively float the children terms.
+      term.float_combinators(combinators, name_gen, book, def_name, builtin, seen);
+
+      // Don't float if it has unscoped variables.
+      if term.has_unscoped_diff() {
+        return;
```
developedby, line 55: Don't return here but rather `continue`, since the siblings might be fine to float.

developedby: Alternatively, add this condition to the `if` below or to the `is_combinator` function (since it kinda captures the idea of an unscoped combinator)

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,151 @@
...
-      Term::Lam { nam, bod, .. } => (nam.as_ref(), bod, false),
-      Term::Chn { nam, bod, .. } => (nam.as_ref(), bod, true),
-      _ => unreachable!(),
-    };
-
-    let parent_scope = term_info.replace_scope(BTreeSet::new());
-
-    let mut detach = bod.go_float(depth + 1, term_info);
+  /// Checks if the term is a lambda sequence with the body being a variable in the scope or a reference.
+  fn is_constant_lambda(&self, book: &Book, seen: &mut IndexSet<Name>) -> bool {
```
developedby, line 116: is_safe_lambda rather than is_constant_lambda, since we're accepting some non-constant non-nullary ctr lambdas

developedby (changes requested):

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,180 @@
...
+  builtin: bool,
+  combinators: &mut BTreeMap<Name, Definition>,
+) {
+  let comb_name = Name::new(format!("{}$C{}", def_name, *name_gen));
+  *name_gen += 1;
 
-      (Detach::Combinator, Detach::Recursive) => Detach::Recursive,
-      (Detach::Recursive, Detach::Combinator) => Detach::Recursive,
-      (Detach::Recursive, Detach::Recursive) => Detach::Recursive,
+  let comb_var = Term::Ref { nam: comb_name.clone() };
```
developedby, line 74:
```suggestion
  let comb_ref = Term::Ref { nam: comb_name.clone() };
```

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,180 @@
...
+  let extracted_term = std::mem::replace(term, comb_var);
```
developedby, line 75:
```suggestion
  let extracted_term = std::mem::replace(term, comb_ref);
```

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,180 @@
...
-      let arm_detach = match rule.body.go_float(depth + 1, term_info) {
-        // If the recursive ref reached here, it is not in a active position
-        Detach::Recursive => Detach::Combinator,
-        detach => detach,
-      };
-
-      detach = detach & arm_detach;
-    }
+  fn is_combinator(&self) -> bool {
+    self.is_closed()
```
developedby, line 139: we don't need this double indirection

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,180 @@
...
+      }
+      Term::Mat { args, rules } => {
+        FloatIter::Mat(args.iter_mut().chain(rules.iter_mut().map(|r| &mut r.body)))
+      }
+      Term::Tup { els } | Term::Sup { els, .. } | Term::Lst { els } => FloatIter::Vec(els),
+      Term::Let { val: fst, nxt: snd, .. }
+      | Term::Use { val: fst, nxt: snd, .. }
+      | Term::Dup { val: fst, nxt: snd, .. }
+      | Term::Opx { fst, snd, .. } => FloatIter::Two([fst.as_mut(), snd.as_mut()]),
+      Term::Lam { bod, .. } | Term::Chn { bod, .. } => FloatIter::One([bod.as_mut()]),
```
developedby, line 169: If you change the Lam case to be `bod.float_children_mut()` I think it'll always keep sequences of lambdas together which would avoid fragmenting some combinators

src/term/transform/float_combinators.rs
```diff
@@ -1,277 +1,180 @@
...
-/// Replaces closed Terms (i.e. without free variables) with a Ref to the extracted term
-/// Precondition: Vars must have been sanitized
 impl Book {
+  /// Extracts unsafe terms into new definitions.
+  ///
+  /// Precondition: Variables must have been sanitized.
+  ///
+  /// The floating algorithm follows these rules:
+  /// - Don't extract safe terms or terms that contains unscoped variables.
+  /// - Extract if it is a closed Application
```
developedby, line 18: Can you update the description?

developedby (changes requested):
Remember to update the documentation

tests/golden_tests/desugar_file/combinators.hvm
```diff
@@ -0,0 +1,23 @@
+(foo) = λa λb λc (foo a)
+
```
developedby, line 2: Can you add this example: `(clax) = (@x x @a @b @c @d (clax d))`
I expect it to be converted to this
```
(clax) = (@x x clax$C0)
(clax$C0) = @a @b @c @d (clax d)
```

developedby (changes requested):

developedby (approved):
Great job

PR #246, opened 2024-03-29
[sc-517] Use match and switch terms like Kind, remove complex pat-match expressions

developedby:


src/term/transform/fix_match_terms.rs
```diff
@@ -0,0 +1,278 @@
...
+        self.fix_match(&mut errs, ctrs, adts);
+      } else if let Term::Swt { .. } = self {
+        self.fix_switch(&mut errs);
+      }
+
+      errs
+    })
+  }
+
+  fn fix_match(&mut self, errs: &mut Vec<FixMatchErr>, ctrs: &Constructors, adts: &Adts) {
```
imaqtkatt, line 97: Can you break this into smaller functions?

PR #249, opened 2024-04-02
[sc-538] Remove simplify_main_ref and simplify_ref_to_ref passes

imaqtkatt:


src/term/mod.rs
```diff
@@ -701,6 +705,28 @@ impl Term {
     })
   }
 
+  /// Returns whether any substitution happened within the term or not
+  pub fn subst_ref_to_ref(term: &mut Term, ref_map: &BTreeMap<Name, Name>) -> bool {
```
developedby, line 709: If this is used only in definition_merge, I think we can move it there

src/term/mod.rs
```diff
@@ -701,6 +705,28 @@ impl Term {
     })
   }
 
+  /// Returns whether any substitution happened within the term or not
```
developedby, line 708: Without the previous context, this description is not sufficient, since it only talks about the return and not what it does to the term

tests/golden_tests/cli/compile_inline.hvm
developedby, line 1: Since hvm-core's inline doesn't inline size 1 nets, this test actually does nothing. Can you change it to test the 3 types of inlining that it actually does (numbers, eras and refs), and add it also to compile_file or compile_file_o_all

tests/snapshots/compile_file_o_all__inline_app.hvm.snap
```diff
@@ -2,7 +2,9 @@
 source: tests/golden_tests.rs
 input_file: tests/golden_tests/compile_file_o_all/inline_app.hvm
 ---
-@Rul3 = (a a)
+@Rul1_$_Rul2 = (a a)
+
+@Rul3 = @Rul1_$_Rul2
 
 @main = a
   & @Rul3 ~ (@Rul3 a)
```
developedby, line 10: Does the inlining not work on the entrypoint?

src/lib.rs
```diff
@@ -49,6 +49,9 @@ pub fn compile_book(
   if opts.eta {
     core_book.values_mut().for_each(Net::eta_reduce);
   }
+  if opts.inline {
+    core_book.inline().map_err(|e| format!("Inline:\n{:ERR_INDENT_SIZE$}{e}", ""))?;
```
developedby, line 53: Could you instead add this to the diagnostics so that we also print the warnings from desugaring?
Something like 
```
diagnostics.start_pass();
let res = core_book.inline();
diagnostics.take_book_err(res);
diagnostics.fatal()?;
```

tests/snapshots/compile_file_o_all__self_ref.hvm.snap
```diff
@@ -3,7 +3,5 @@ source: tests/golden_tests.rs
 input_file: tests/golden_tests/compile_file_o_all/self_ref.hvm
 ---
 Errors:
-In definition 'Foo':
-  Definition is a reference to itself.
-In definition 'main':
-  Definition is a reference to itself.
+Inline:
```
developedby, line 6: "While inlining the compiled program:" or "During inlining:".
Ideally we check this before so that we get actual function names, but it probably won't catch all cases that can happen here.

developedby (changes requested):

src/lib.rs
```diff
@@ -49,6 +49,9 @@ pub fn compile_book(
   if opts.eta {
     core_book.values_mut().for_each(Net::eta_reduce);
   }
+  if opts.inline {
```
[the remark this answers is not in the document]
imaqtkatt: It does make difference in some references

tests/snapshots/compile_file_o_all__inline_app.hvm.snap
imaqtkatt: I believe in this test it does not work because of the constructor node or because Rul3 is pointing to another reference

developedby: I expected @Rul3 to inline to @Rul1_$_Rul2 here. It should be inlining Ref nodes since they have size 0.

developedby (approved):

PR #251, opened 2024-04-02
[sc-512] Add tests for file builtin functions

imaqtkatt:


developedby (changes requested):
Can you also add a test where these functions fail. For read it could be trying to read a file that doesn't exist.
For write i think writing a file to a non-existant directory would work, but if not the trying to write somewhere we don't have permission (hopefully the first because this second option seems hard to setup)

developedby (approved):

PR #257, opened 2024-04-05
[sc-536] Add debug mode for lazy evaluation in hvml

imaqtkatt:


developedby:
An alternative would be for hvm-core to implement partial lazy reduction which both hvm-core and hvm-lang could use for debug mode

src/lib.rs
```diff
@@ -213,10 +213,25 @@ pub fn run_compiled(
     let start_time = Instant::now();
 
     if let Some(mut hook) = hook {
-      while !root.redexes.is_empty() {
-        let readback = host.lock().unwrap().readback(root);
-        hook(&readback);
-        root.reduce(1);
+      if run_opts.lazy_mode {
```
developedby, line 216: Can you separate each of these cases into separate functions?
So that our "run" function is simpler
```
match (hook, run_opts.lazy_mode, max_rwts) {
 (Some(hook), true, max_rwts) => normal_lazy_debug(),
 (...) => normal_strict_debug(),
 ...
}
```

developedby (approved):

PR #261, opened 2024-04-09
[sc-543] Add hvml test on example programs

imaqtkatt:


developedby (commented):
Maybe instead we could have these files in the examples directory and the test function reads from there?
This way we test the examples and are forced to keep them up-to-date, give users more examples and don't duplicate these programs.

developedby:
That's what hvm-core is doing I think, if you want to have a look there.

tests/golden_tests/examples/all_tree.hvm
developedby, line 1: The files in this directory are not being used, right?

imaqtkatt: Yes, I thought I moved then but it was copied instead.

developedby (approved):

PR #267, opened 2024-04-12
[sc-537] Add pretty printing of hvml defs and terms

imaqtkatt:


developedby (commented):
Maybe we could move it to display.rs, and `impl Display for Term` as `term.pretty(0)`.

That way we'd only have one implementation for displaying terms.

src/term/pretty.rs
```diff
@@ -0,0 +1,118 @@
+use std::fmt::Display;
+
+use crate::term::display::DisplayJoin;
+
+use super::{display::DisplayFn, Name, Term};
+
+impl Term {
+  pub fn pretty(&self, tab: usize) -> impl Display + '_ {
```
developedby, line 8: display_pretty is more descriptive

developedby (changes requested):

developedby (approved):

src/term/display.rs
```diff
@@ -203,3 +203,117 @@ impl Tag {
 fn var_as_str(nam: &Option<Name>) -> &str {
   nam.as_ref().map_or("*", Name::deref)
 }
+
+impl Term {
+  pub fn display_pretty(&self, tab: usize) -> impl fmt::Display + '_ {
+    DisplayFn(move |f| match self {
```
developedby, line 209: You have to add a `maybe_grow` here i think

developedby (changes requested):

developedby (approved):

PR #269, opened 2024-04-16
[sc-251] Mark scopeless lambdas on readback in hvm-lang

imaqtkatt:


tests/snapshots/readback_lnet__complicated_dup.hvm.snap
```diff
@@ -2,4 +2,4 @@
 source: tests/golden_tests.rs
 input_file: tests/golden_tests/readback_lnet/complicated_dup.hvm
 ---
-λa (λb (a (b c)) λc (a (b c)))
+λa (λb (a (b $c)) λ$c (a (b $c)))
```
developedby, line 5: Here `b` should also be unscoped

src/term/net_to_term.rs
```diff
@@ -88,11 +105,24 @@ impl Reader<'_> {
           // If we're visiting a port 0, then it is a lambda.
           0 => {
             let nam = self.namegen.decl_name(self.net, Port(node, 1));
+            if let Some(nam) = &nam {
+              if !self.unscoped_vars.contains(nam) {
+                self.vars.insert(nam.clone());
+              }
+            }
             let bod = self.read_term(self.net.enter_port(Port(node, 2)));
```
developedby, line 113: You have to remove `vars` from the scope after reading the lambda body

developedby (changes requested):

src/term/net_to_term.rs
```diff
@@ -65,6 +80,8 @@ pub struct Reader<'a> {
   seen_fans: Scope,
   seen: HashSet<Port>,
   errors: Vec<ReadbackError>,
+  vars: HashSet<Name>,
```
developedby, line 83: I think it would be better to move the gathering of unscopeds to a separate funcion to not increase the complexity of the Reader, which is already quite complex.

We could in addition make the reader always return unscoped lambdas only, to make it's result correct even before checking the scope of variables

developedby (changes requested):

developedby (approved):

PR #270, opened 2024-04-17
refactor: use patterns in more of the ast

tjjfvi:
- lambdas and lets can now bind simple patterns (e.g. `@(a, b) bod` or `let ((a, b), (c, d)) = val; nxt`)
  - `Term::Dup` and `Term::Ltp` are now part of `Term::Let`
- these patterns can now bind scopeless variables (so e.g. `let $x = ...` is valid)
  - `Term::Chn` is now part of `Term::Lam`
- tuples can now be tagged, and use the same AST node as superpositions
  - tup/sup/dup are all collectively called fan
  - the logic for these is nearly identical for the vast majority of the code
- erased lets are no longer elided, fixing some miscompilation bugs

(even though this added features, the line count is net negative!)


tjjfvi:
(this is a modern version of the original pattern-refactor started by @FranchuFranchu (https://github.com/HigherOrderCO/hvm-lang/commit/cf1bc258ee01307045169cba5a931836da9ca4c4))

src/term/net_to_term.rs
```diff
@@ -500,9 +514,10 @@ impl Term {
   pub fn apply_unscoped(&mut self, unscoped: &HashSet<Name>) {
     match self {
```
imaqtkatt, line 515: Needs to be implemented for Term::Let

tests/golden_tests/readback_lnet/bad_net1.hvm
```diff
@@ -1,2 +1,2 @@
 (a b)
-& {3 c b} ~ (c a)
\ No newline at end of file
+& {2 c b} ~ (c a)
```
developedby, line 2: Does hvm-core completely ignore the value of labels? As in, it doesn't ever look at the meaning of any specific values, just compares them between nodes.

developedby (changes requested):

tests/golden_tests/readback_lnet/bad_net1.hvm
tjjfvi: correct

tests/golden_tests/compile_file/unused_dup_var_linearization.hvm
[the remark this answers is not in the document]
tjjfvi: fair

imaqtkatt (approved):

PR #271, opened 2024-04-18
encode terms directly to hvmc nets

tjjfvi:
This changes the encoding to encode `hvml::term::Term` directly to a `hvmc::ast::Net` without the `hvml::net::INet` intermediary.

Readback still has the intermediary for now.

There's a lot of snapshot churn due to the variable name generation changing. Theoretically we could have another pass over the net to change the variables to the old format, but IMO it's more readable with the variable names this way, so I don't think it should be changed.

This will now detect vicious circles that are disconnected from the rest of the net (which did not work before).

Unfortunately, when vicious circles are created, it leaks a little bit of memory, but I think this is an acceptable tradeoff (and I don't see any way to avoid it in this approach without using a lot more memory).


src/term/term_to_net.rs
```diff
@@ -129,125 +160,126 @@ impl EncodeTermState<'_> {
...
 
-  fn fan_kind(&mut self, fan: &FanKind, tag: &Tag) -> NodeKind {
+  fn fan_kind(&mut self, fan: &FanKind, tag: &Tag) -> CtrKind {
     let lab = self.labels[*fan].generate(tag);
-    Ctr(if *fan == FanKind::Tup { Tup(lab) } else { Dup(lab.unwrap()) })
+    if *fan == FanKind::Tup { Tup(lab) } else { Dup(lab.unwrap()) }
   }
 
-  fn link_global(&mut self, name: &Name, trg: Trg) {
+  fn link_global(&mut self, name: &Name, place: Place<'t>) {
```
developedby, line 275: Do you think making scoped binds share the same logic as scopeless ones would be an improvement?
`link_global` is just a `half_link` and the only thing that differentiates scoped from scopeless at this point is the different namespaces.
With this PR you use the fact that scoped vars are scoped to always only link when the var use is encountered, but this extra bit of information is not really necessary.

developedby (commented):
Originally I had actually implemented something similar to this to encode terms (although not as nicely put together), but because I couldn't find a good way to do readback directly I went with the intermediate net to be symmetric.

tjjfvi:
> Originally I had actually implemented something similar to this to encode terms (although not as nicely put together), but because I couldn't find a good way to do readback directly I went with the intermediate net to be symmetric.

Ah, gotcha. I do plan to also make a PR for using this on the readback (using the algorithm Franchu and I worked out); just made this change separately because it's simpler.

src/term/term_to_net.rs
tjjfvi: Hmm, could it be changed to share the same logic? I was under the impression that this algorithm had to account for shadowing (hence the `Vec`), so an approach like used for global vars would break on things like `@a ((@a a), a)`.

tjjfvi: If it doesn't actually need to account for shadowing then yeah I think it'd be nicer to use the same logic.

developedby: At this point we've renamed all the variables to have unique names

developedby (approved):

PR #272, opened 2024-04-18
[sc-624] Add monadic bind syntax to hvm-lang

imaqtkatt:


tests/golden_tests/desugar_file/bind_syntax.hvm
```diff
@@ -0,0 +1,8 @@
+Result.bind (Result.ok val) f = (f val)
+Result.bind err _ = err
+
+Main = do Result.bind {
+  ask y = (HVM.store "file.txt" "Alice");
```
developedby, line 5: I'm thinking about the possibility of accepting `let` expressions mixed with `ask` expressions.
Most monadic programs are not only composes of purely sequences of that operation, but interleaved with other things.

Take for example
```
ask a = A
let b = (B a)
ask c = (C b)
(X b c)
```

We want it to generate
```
(Bind A @a let b = (B a); (Bind (C b) @c @b (X b c) b))
```
Which would require putting `let b` between the asked variable and the next term.

Of course you solve this with some sort of identity operator for that bind, but for some monads that would be very strange
```
// ok for maybe / either / result
ask a = A
ask b = (Bind.unit (B a))
ask c = (C b)
(X b c)
```
It also introduces an overhead, decreases parallelism and is generally annoying to have.

You can also always just do 
```
do Bind {
  ask a = A
  let b = (B a); do Bind {
    ask c = (C b)
    (X b c)
  }
}
```
So I'm conflicted on whether or not we should add it. What do you think?

developedby (changes requested):

developedby (approved):

PR #274, opened 2024-04-22
read terms directly from hvmc nets

tjjfvi:
This PR, along with its counterpart, #271 finish the work @FranchuFranchu started to remove the intermediate `INet` representation (last commit: https://github.com/HigherOrderCO/hvm-lang/commit/7a57879efa0ea4b9352edfb66f2605dac3e34135).

Currently, only linear readback uses the `loaned`-based algorithm. Non-linear readback uses the old algorithm with a `HashMap` of uplinks instead of the `INet` representation. I think there may be a way to use a similar algorithm for non-linear readback, but I'm not sure exactly how it might work.

This new algorithm for linear readback also makes it more powerful; it is now able to read back more of the net. For example, `@x let * = (x 123); 0` would previously be read back as `@x 0` (which is notably non-linear).

Marking this as draft because I haven't read through the full diff yet, but it should be reviewable.

The first 3 commits reorganize/add some tests; if you're curious to see how the snapshots changed with this new algorithm, you'll want to exclude those commits from the diff.

[@FranchuFranchu I'd be curious to get your thoughts]


src/term/readback/linear.rs
```diff
@@ -0,0 +1,825 @@
...
+//!       - we link port 0, `123`, up to `?v`
+//!         - we simply fill the term hole `?v` with the term `123`
+//!     - visit the right child, `x` (with uplink `(a ?v)`)
+//!       - this is a variable; add ``("x", `(a 123)`)`` to the hashmap
+//!       - we link port 0, `123`, up to `?v`
+//!   - visit the right child, `x` (with uplink `?t`)
+//!       - this is a variable; remove ``("x", `(a 123)`)`` from the hashmap
+//!       - link `(a 123)` to `?t`
+//!         - we simply fill the term hole `?t` with the term `(a 123)`
+//! - we're finished, the root term is now `@a (a 123)`
```
FranchuFranchu, line 300: Great explanation! :smiley:

src/term/readback/non_linear.rs
```diff
@@ -60,73 +59,188 @@ pub fn net_to_term(
 }
 
 // BTreeSet for consistent readback of dups
```
FranchuFranchu, line 61: Update comment here

FranchuFranchu:
> @FranchuFranchu I'd be curious to get your thoughts

The code looks great to me; it's like the old pattern refactor branch, but cleaner (and clearer!).

src/term/readback/non_linear.rs
developedby: You can just remove it since it's implied by the type. (the only reason we ever use IndexMap is for insertion ordered iteration)

nicolas-abril (commented):
We should keep only one of the two approaches

tjjfvi:
> We should keep only one of the two approaches

What do you mean? Like, only keep linear or non-linear?

developedby:
> > We should keep only one of the two approaches
> 
> What do you mean? Like, only keep linear or non-linear?

I mean use only one approach to do both linear and non-linear. I don't think it's good to have two different readback algorithms coexisting in the repo, it just complicates things.

tjjfvi:
Well, the current algorithm for linear readback does not work in all cases (it at times omits significant parts of the net), and I don't know yet how to use the new algorithm for non-linear readback, so I'm not sure if that's possible atm.

developedby:
The PR is looking good, but I'll hold it until I finish porting things to hvm32

developedby:
This wasn't merged since it doesn't do the non-linear readback (replacing dup variables by the term they're duplicating), which for users is the most regular way of looking at the result.

I still want to incorporate this into Bend, but I'll close the PR since it's going to conflict with so many things we probably want to refactor all the changes.

PR #275, opened 2024-04-26
[sc-627] Make compilation compatible with hvm32

developedby:
Huge PR, but separately it wouldn't pass the tests.
Moves the compilation target to hvm32. This has some consequences:
* Only two ctr nodes (`()` and `{}`), no labels.
  - This means we must use scott encoding and can't readback ADT constructors and matches.
  - It also means that church encoding doesn't work for a lot of things.
  - Tag-based vectorization also doesn't work.
  - Tuples are encoded as Con nodes, so they can't be readback and destructively interact with lambdas/applications.
* Numbers and numeric operations are expressed in a different way.
* Numbers now have 3 types: U24, I24 and F24. Currently only U24 is used.
* Different symbols used numeric operations (`$()`) and switches (`?()`).
* Removed `<<` and `>>` operations. They have to be added back as a syntax sugar.
* Definitions have a limit of 32 inet nodes.
* No debug, no parallel execution, no lazy mode, no execution with limited number of interactions and memory.
* No IO yet.
* Since hvm32 is in a private repo and doesn't expose itself as a rust library, still use hvm-core but using a compatibility branch with these changes to syntax.
* Running a program is done by spawning a subprocess. hvm32 must be installed.
  - This causes some concurrency problems when running tests in parallel. So only execute one "run"-type test at a time.
  - This also causes the `hangs` test to leak the subprocess, consuming one core and a lot of indefinitely.
  - We also can't do compile-time reduction (unless we also update hvm-core execution, which I didn't).
* Can't run the ignored tests anymore since they are broken.
* Running a program doesn't expand `main` so we to work around that.

Some new features that were needed:
* Add a check to see if compiled definitions exceed the size limit.
* Updated float_combinators to create definitions that are as large as possible but still within the node limit.
* Added a pass to expand generated function in the end. Since normalization doesn't expand anymore we need this so that users don't get generated functions on readback.


developedby:
Tests are going to fail on the github actions because it doesn't have hvm installed.

developedby:
Readback of numbers, numeric operations and switches is bugged and will output errors sometimes even though the term is ok.
We'll solve that in another PR

imaqtkatt (approved):

PR #276, opened 2024-04-29
[sc-625] Add procedural-style syntax for hvm-lang

imaqtkatt:


src/term/flavour_py/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+pub enum Term {
+  // "None"
+  None,
+  // [a-zA-Z_]+
+  Var { nam: Name },
+  // [0-9_]+
+  Num { val: u32 },
+  // {fun}(args,)
+  Call { fun: Box<Term>, args: Vec<Term> },
+  // "fun" {pat} -> {bod}
```
developedby, line 21: Although i prefer this, if we're imitating python, I think we should try to go all in and use `lambda x1 ... xn: term`.

src/term/flavour_py/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+  Lam { pat: AssignPattern, bod: Stmt },
+  // {nam} "{" {fields} "}"
+  Enum { nam: Name, fields: Vec<(Name, Term)> },
```
developedby, line 24: If we allow keyword arguments in functions, we could have the same syntax for function calls and enums (in the end they're the same thing right?).
That's how python does it, you create an object by calling the constructor function (which is rather complicated, but for us we can just pretend it's a normal function call).

src/term/flavour_py/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+
+// "def" {name} "(" {params} ")" ":" {body}
+#[derive(Clone, Debug)]
+pub struct Definition {
+  pub name: Name,
+  pub params: Vec<Name>,
+  pub body: Stmt,
+}
+
+// "enum" "{" {variants} "}"
```
developedby, line 80: I think we should change the enum syntax to be indentation based
```
enum MyEnum:
  Ctr1(a, b, c)
  Ctr2
  Ctr3(a, b, c, d)
```

developedby (changes requested):

src/term/flavour_py/mod.rs
imaqtkatt: It is already based on the indentation, I just forgot to update this comment after implemented it

src/term/flavour_py/parser.rs
```diff
@@ -0,0 +1,631 @@
...
+      head.extend(tail);
+      Ok(Term::Lst { els: head })
+    }
+  }
+
+  fn parse_term_py(&mut self) -> Result<Term, String> {
+    self.skip_trivia();
+    if self.try_consume_keyword("lambda") {
+      let names = self.list_like(|p| p.parse_hvml_name(), "", ":", ",", true, 1)?;
+      let bod = self.parse_stmt_py(&mut Indent(0))?;
```
developedby, line 116: Confusingly, in python syntax lambdas have an expression as body and not a statement.

Lambdas become a return statement, where the `return` keyword is implicit.

src/term/flavour_py/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+  // [0-9_]+
+  Num { val: u32 },
+  // {fun}(args,)
+  Call { fun: Box<Term>, args: Vec<Term>, kwargs: Vec<(Name, Term)> },
+  // "lambda" {pat}* ":" {bod}
+  Lam { names: Vec<Name>, bod: Stmt },
+  // {lhs} {op} {rhs}
+  Bin { op: Op, lhs: Box<Term>, rhs: Box<Term> },
+  // "\"" ... "\""
+  Str { val: GlobalString },
```
developedby, line 27: We could add an `if` expression in the future

src/term/flavour_py/parser.rs
```diff
@@ -0,0 +1,631 @@
...
+    }
+    if block {
+      return Ok(count == 0 && !self.is_eof());
+    }
+    // TODO: add the current line in Err
+    if count == 0 { Ok(true) } else { Err("Indentation error".to_string()) }
+  }
+
+  fn parse_stmt_py(&mut self, indent: &mut Indent) -> Result<Stmt, String> {
+    self.skip_exact_indent(indent, false)?;
```
developedby, line 227: A maybe_grow here

developedby (changes requested):
Right now we don't have any way of using this syntax from the cli right?

src/term/flavour_py/parser.rs
```diff
@@ -0,0 +1,631 @@
...
+      '\"' => {
+        let str = self.parse_quoted_string()?;
+        let val = STRINGS.get(str);
+        Term::Str { val }
+      }
+      c if c.is_ascii_digit() => {
+        let val = self.parse_u64()?;
+        Term::Num { val: val as u32 }
+      }
+      _ => {
```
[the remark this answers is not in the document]
developedby: Github is not showing it properly, but it's converting True to 1 and False to 0

developedby (approved):

PR #284, opened 2024-05-07
[sc-658] Add doc with bend syntax

imaqtkatt:


docs/syntax.md
```diff
@@ -0,0 +1,512 @@
...
+Main = do Result.bind {
+  ask y = (div 3 2);
+  ask x = (rem y 0);
+  x
+}
+```
+
+Receives a monadic bind function and then expects a block.
+
+The block is followed by `ask` binds and ends with a return term.
```
developedby, line 449: Add that the bind function should be of type `(Monad a) -> (a -> (Monad b)) -> (Monad b)`.

Add that it forwards any used variables and point to the doc (which we still have to write)

developedby (changes requested):

docs/syntax.md
```diff
@@ -0,0 +1,667 @@
...
+
+### Character Literal
+
+```python
+'x'
+```
+
+A Character is surrounded with `'`. Accepts unicode characters, unicode escapes in the form '\u{hex value}' and is desugared to the unicode codepoint as an `u24`.
+
+Only supports unicode codepoints up to `0xFFFFFF`.
```
developedby, line 315: Either add codepoints to the dictionary or change it into 'code points' (apparently both are correct)

docs/syntax.md
```diff
@@ -0,0 +1,512 @@
...
+```python
+def add(x, y):
+  result = x + y
+  return result
+
+def main:
+  return add(40, 2)
+```
+
+A definition is composed by a name, a sequence of parameters and a body.
```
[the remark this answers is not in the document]
developedby: You specified for variables but not for top-level names.
Currently they allow the same, except top-level can't have '__' (i use for generated names)

docs/syntax.md
```diff
@@ -0,0 +1,667 @@
...
+  None
+
+enum Tree:
+  Node(value, ~left, ~right)
+  Leaf
+```
+
+Enum names must be unique, and should have at least one constructor.
+
+Each constructor is defined by a name followed by its fields. The `~` notation describes a recursive field.
```
developedby, line 46:
```suggestion
Each constructor is defined by a name followed by its fields.
The `~` notation indicates a recursive field. To use `fold` statements with a type its recursive fields must be correctly marked with `~`.
```

docs/syntax.md
```diff
@@ -0,0 +1,667 @@
...
+
+The constructor names inherit the name of their types and become functions (`Tree/Node` and `Tree/Leaf` in this case).
+The exact function they become depends on the encoding.
```
developedby, line 49: Add a link to the document about defining datatypes

docs/syntax.md
```diff
@@ -0,0 +1,667 @@
...
+bend x = 1, y = 2 ... while condition(x, y, ...):
+```
+
+It is equivalent to this inline recursive function:
+
+```python
+def bend(x, y, ...):
+  if condition(x, y, ...):
+    ...
+    return ... bend(x, y, ...) ...
```
developedby, line 193: Which keyword are we using to indicate the recursive call? `bend` or `go`? I think both are ok. `bend` is more cool looking, but `go` might be more intuitive.
`rec` would be good for people who are used to languages that have a similar marker to indicate recursive functions, but I suspect that's not a lot of people.

docs/syntax.md
```diff
@@ -0,0 +1,667 @@
...
+
+A Tuple is surrounded by `(` `)` and should contain 2 or more elements. Elements are separated by `,`.
+
+### Superposition
+
+```python
+{1 2 3}
+```
+
+A superposition of values is defined using `{` `}` with at least 2 expressions inside.
```
developedby, line 279: Add a link to the dup/sup document that explain how they work

docs/syntax.md
```diff
@@ -0,0 +1,667 @@
...
+
+A function definition is composed of a sequence of rules, where a rule is the name of the function, a sequence of patterns and then the body.
+
+A rule pattern can be:
+- A variable.
+- A number.
+- A constructor.
+- A tuple.
+- A superposition.
+
```
developedby, line 407: Add a link to the document on pattern matching to learn about what exactly the rules for pattern matching equations are

developedby (changes requested):

developedby (changes requested):

PR #285, opened 2024-05-07
[sc-664] Add inference for do notation

imaqtkatt:


src/term/transform/apply_bnd.rs
```diff
@@ -1,35 +1,71 @@
...
-    }
+
+      Ok(())
+    })
+  }
+}
+
+fn make_fun_name(typ: &mut Name) -> Name {
+  let name: String = [typ, "/", "bind"].into_iter().collect();
+  Name::new(name)
```
developedby, line 64:
```suggestion
  Name::new(format!("{typ}/bind")
```

developedby (changes requested):

developedby (approved):

developedby (changes requested):
Fix the failing checks

PR #286, opened 2024-05-08
[sc-666] Add a map type to bend

imaqtkatt:


src/fun/builtins.hvm
```diff
@@ -2,3 +2,43 @@ data String = (String.cons head ~tail) | (String.nil)
...
 data Nat    = (Nat.succ ~pred) | (Nat.zero)
+data Map    = (Map.node value ~left ~right) | (Map.leaf)
+
+Map/empty = Map.leaf
+
+Map/get map key =
+  match map {
+    Map.leaf: (*, map)
+    Map.node:
+      switch _ = (== 0 key) {
```
developedby, line 13: I'm pretty sure we could optimize this, but for now let's go with it.

developedby (approved):

PR #298, opened 2024-05-13
Add io builtins

imaqtkatt:


src/fun/builtins.hvm
```diff
@@ -42,3 +45,39 @@ Map/set map key value =
...
+IO_DONE_TAG       = 0
+IO_PUT_TEXT_TAG   = 1
+IO_GET_TEXT_TAG   = 2
+IO_WRITE_FILE_TAG = 3
+IO_READ_FILE_TAG  = 4
+IO_GET_TIME_TAG   = 5
+IO_SLEEP_TAG      = 6
+IO_DRAW_IMAGE_TAG = 7
+
+data IO
```
developedby, line 63: Could you add `IO/bind` like this https://github.com/HigherOrderCO/Kindex/blob/master/Data/IO/bind.kind2 ?

src/fun/builtins.hvm
```diff
@@ -42,3 +45,39 @@ Map/set map key value =
...
+  = (Done term)
+  | (PutText   text      cont)
+  | (GetText             cont)
+  | (WriteFile file data cont)
+  | (ReadFile  file      cont)
+  | (GetTime             cont)
+  | (Sleep     time      cont)
+  | (DrawImage tree      cont)
+
+MkStr (String/cons x xs) = λt (t STRING_CONS_TAG x (MkStr xs))
```
developedby, line 73: We also need a function to read strings right? The opposite of this one (matches on the HVM string and converts to scott encoding

src/fun/builtins.hvm
```diff
@@ -42,3 +45,39 @@ Map/set map key value =
...
+MkStr (String/nil)       = λt (t STRING_NIL_TAG)
+
+MkIO (IO/Done term)                = λt (t IO_DONE_TAG term)
```
developedby, line 76: Same here, we need a function to revert this transformation

developedby (changes requested):
Do you think we should for now have a compiler pass that auto inserts `MkIO` and it's reverse?

developedby (approved):

PR #303, opened 2024-05-14
[sc-678] Add run-c and run-cuda commands to bend

imaqtkatt:


src/main.rs
```diff
@@ -104,6 +84,33 @@ enum Mode {
   },
 }
 
+#[derive(Args, Clone, Debug)]
+struct RunArgs2 {
```
developedby, line 88: Call this one run args and other one run opts

src/main.rs
```diff
@@ -290,6 +297,74 @@ fn execute_cli_mode(mut cli: Cli) -> Result<(), Diagnostics> {
         println!("{stats}");
       }
     }
+
+    Mode::RunC(RunArgs2 { pretty, run_opts, comp_opts, warn_opts, path, arguments }) => {
```
developedby, line 301: Instead of duplicating this, you could match the three cases here and then inside match again to get the specific function you want to run.

src/lib.rs
```diff
@@ -144,12 +144,13 @@ pub fn desugar_book(
...
 }
 
-pub fn run_book(
+pub fn run_book_with_fn(
   mut book: Book,
   run_opts: RunOpts,
   compile_opts: CompileOpts,
   diagnostics_cfg: DiagnosticsConfig,
   args: Option<Vec<Term>>,
+  run_fn: impl Fn(&str) -> Result<Output, String>,
```
developedby, line 153: Instead of passing a function, isn't passing the name of the hvm command ("run", "run-c", etc) enough?

developedby (changes requested):

developedby (changes requested):
Also remember to update the readme

README.md
```diff
@@ -9,63 +9,65 @@ It is based on the [Interaction-Calculus](https://github.com/VictorTaelin/Intera
...
-Desugar   | `bend desugar <file>` | Desugars a program to the core syntax and outputs it to stdout
-
+| Command   | Usage                 | Description                                                                    |
+| --------- | --------------------- | ------------------------------------------------------------------------------ |
+| Check     | `bend check <file>`   | Checks if a program is valid                                                   |
+| GenHvm    | `bend gen-hvm <file>` | Compiles a program to HVM and outputs it to stdout                             |
+| Run       | `bend run <file>`     | Compiles and then runs a program with the Rust HVM implementation              |
+| Run-C     | `bend run-c <file>`   | Compiles and then runs a program with the C HVM implementation                 |
+| Run-Cu    | `bend run-cu <file>`  | Compiles and then runs a program with the Cuda HVM implementation              |
+| Normalize | `bend norm <file>`    | Compiles and then normalizes a program in HVM, outputting the result to stdout |
```
developedby, line 39: Can you also add the `bend norm` command? It should just call run-c *without* the `--io` hvm flag.

developedby: Or maybe we should also have --io flag?
I don't know what's better.

I assume that if we have `bend norm` we'd also want `bend norm-cu`

developedby (changes requested):

README.md
developedby, line 39: Add the norm-cu command here. Also, change `norm` to `norm-c`. I havent checked if the rust runtime has IO, but if it does then add that command  too

developedby (changes requested):

README.md
imaqtkatt: The Rust impl does not accept the `--io` argument

developedby: Ok

developedby: Then we should remove `bend run` and add `bend norm-c`

developedby (approved):

PR #318, opened 2024-05-16
[sc-695] Add 'bend' to the functional syntax document

imaqtkatt:


docs/syntax.md
```diff
@@ -445,28 +460,39 @@ i24 = -42
...
-| Division       | x / y  |
-| Equal          | x == y |
-| Not Equal      | x != y |
-| Less Than      | x < y  |
-| Greater Than   | x > y  |
-| Bitwise And    | x & y  |
-| Bitwise Or     | x \| y |
-| Bitwise Xor    | x ^ y  |
+| Operation      | Syntax   |
+| -------------- | -------- |
```
developedby, line 464: Here and below on the other syntax, we need to explain that the three number types cannot be mixed and list what types can use which operation in the table

developedby (changes requested):

developedby (approved):

PR #329, opened 2024-05-17
Fix shader usage in README.md

kryptan:
I think there is a mistake in one of the examples in readme, `shader` argument is not actually used.


developedby:
Thank you!

developedby:
Already done in https://github.com/HigherOrderCO/Bend/pull/335

Issue #325, opened 2024-05-17
What license is this under?


purpleidea:
I can't trivially find out what license this is under?

I recommend putting some open source license. I use GPLv3+ for example in https://github.com/purpleidea/mgmt/

Otherwise this will be seen as proprietary software, and probably not be great for adoption.

HTH

developedby:
Hi, it will be open-source, but we haven't decided on the exact license yet. This will be done very soon, I'll close this once it's done

purpleidea:
Okay neat! Feel free to ping if you want non-your-lawyer-not-a-lawyer advice or to chat about this. (I have no vested interest in the project, so I'm probably unbiased, I'm a golang hacker, but I found this project because I'm interested in parallel execution runtime as I'm doing in my mgmt config project.)

truedat101:
Thanks for posting this issue @purpleidea , this would have been my first question.   I've worked with a number of commercial companies taking closed source to open source, and in most cases, these decisions and business implications of OSS license choice are not well understood by in house corporate legal unless they come from a background of OSS licensing.  Generally speaking what you are trying to enable is clarity around what is being granted and how the recipient is able to (or not able to) use the software for commercial and non-commercial use.

As long as you don't accept outside contributions, and this is clean room in house code, then you have flexibility to relicense your code base as you see fit (i.e. choose a permissive Academic license and then change your mind later and switch to a copyleft license).

purpleidea:
A license is the constitution of the community. If you use a
permissive license or a CLA, it's a signal that you may plan to make
the project proprietary and that community contributions are "free as
in labour" not "Free as in Freedom". There's nothing necessarily wrong
with that, but it's good to be clear and transparent with your
community so they can know what to expect. GPLv3 protects the
interests of your community, Apache v2 (ALv2) protects the interests
of your company (but it can back fire because others can fork and
leave you in the dust!). It's up to the project to decide, but a
decision is pretty important. It's more complicated than that, but
this project seems neat, so I'm happy to answer questions offline if
that would help them.

HTH
James

kings177:
Thank you for taking your time writing this, we decided to follow with the Apache 2.0 licensing.

kings177:
Closed by https://github.com/HigherOrderCO/Bend/pull/336

PR #336, opened 2024-05-17
Add License to BEND

Sipher:
Added Apache 2.0 License, increased version and changed the description

New version - 0.2.7

New description - "A parallel language" <- Reverted

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "bend-lang"
 description = "A high-level, massively parallel programming language"
 license = "Apache-2.0"
-version = "0.2.6"
+version = "0.2.7"
 edition = "2021"
 exclude = ["tests/snapshots/"]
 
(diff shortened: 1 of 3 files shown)


developedby:
Why did you change the description of the package?

Sipher:
I understood that Taelin wants the  description to be "A parallel language", Just like in the website, I might have misunderstood, if so, I can change back to the old one.

kings177:
> Why did you change the description of the package?

We are trying to keep the wording of things the same throughout every place we mention anything, and since we are now describing Bend as "A parallel language", we should apply this to everything else.

developedby:
That's so ambiguous and not meaningful. You have to imagine that people will see this description without the context of the website and other stuff

kings177:
hmmm, what do you propose?

developedby:
I think the current description is fine, and it matches the description of the hvm repo right now

Sipher:
Reverted the description to the old one, kept the version updated and added the license.

developedby:
Closes #325

PR #360, opened 2024-05-18
Changed return type of fn main() in src/main.rs to fn main() -> std::process::ExitCode.

PfaulJulian:
Hi, I did this because I needed it for personal use. Since the "bend check [file]" command returned 0 in $?, even if it failed or reported an error. So in order to being able to check the exit status of "bend check" I added a return ExitCode::FAILURE, when execute_cli_mode(cli) returned a diagnostics message to main().

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,8 +5,10 @@ use bend::{
   load_file_to_book, run_book_with_fn, CompileOpts, OptLevel, RunOpts,
 };
 use clap::{Args, CommandFactory, Parser, Subcommand};
-use std::path::{Path, PathBuf};
-use std::process::ExitCode;
+use std::{
+  path::{Path, PathBuf},
+  process::ExitCode,
+};
 
 #[derive(Parser, Debug)]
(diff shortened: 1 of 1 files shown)


developedby (commented):
I checked the code and this is not enough.
There's a bug in how we're handling error returns and this will not currently cover all cases unfortunately.
I'll fix it completely and then close this PR, is that ok?

developedby:
ACtually,I'll just merge yours first and then do what's missing

developedby (approved):

PfaulJulian:
Thank you, for your review. I had already guessed that this wouldn't have been enough. :)

developedby (changes requested):
Can you run `cargo fmt` to fix the indentation?

PfaulJulian:
> Can you run `cargo fmt` to fix the indentation?

Yes, give me a few minutes please. :)

PfaulJulian:
Did it work?

developedby:
yes

PfaulJulian:
:)

PR #365, opened 2024-05-18
Add text version of demo

Jacob-Lockwood:
Add text version of demo, for accessibility and just general usefulness. People I showed Bend to complained that the video was inconvenient, and I agree. I still linked to the Twitter post if anyone prefers that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ def rots(d, s, tree):
       return tree
     case _:
       (x, y) = tree
-      return down(d, s, warp(d-1, s, x, y)
+      return down(d, s, warp(d-1, s, x, y))
 ```
 
 Need a fragment shader? Just write a function.
--- a/README.md
+++ b/README.md
@@ -120,7 +120,8 @@ Result: 5005402
 - MIPS: 4534.43
 ```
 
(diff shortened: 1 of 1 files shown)


anplumbing23 (approved):

antonkesy (approved):
Great addition ^^ LGTM

developedby:
This is a bit hard to decide because there are people who will prefer both versions.
I'll definitely add the text version as well, but I don't know which will be the main one

kings177:
i think we should keep the video where it is, but add in the footer of it something like "read the whole transcription here", linking to the text version.

anplumbing23:
What is this

Jacob-Lockwood:
I think linking to the video is better than linking to the text transcription because if I just opened the Bend repo to look at the readme, I don’t want to start reading and then immediately get hit with a 2-minute-long GIF I can’t pause or skip forward in. It’s quite difficult to mentally parse the code on screen or the benchmarks that show up when they’re only on screen for a few seconds.

On the flip side, with the text version, if I don’t care about the code and just want the benchmarks, I can just scroll past that part and be done.

developedby:
The demo was removed in #547

PR #369, opened 2024-05-18
fix: improve fib.bend

### LINKED ISSUE #367: N'th Fibonacci sequence function fails to output result when n > 35
**Describe the bug**
When running the `fib.bend` program with an input value larger than 30, the time to generate an output becomes unusably slow.

In contrast, running the same program in Rust produces a near-instant output. This might be due to the `fib.bend` program not being parallelizable.

`fib.bend` (very slow):
```haskell
add = λa λb (+ a b)

fib = λx switch x {
  0: 1
  _: let p = x-1; switch p {
    0: 1
    _: (+ (fib p) (fib p-1))
  }
}

main = (fib 30)

```
fibonacci in rust (near instant)
```rust
fn fibonacci(n: u32) -> u64 {
  if n == 0 {
      return 0;
  }
  let mut prev = 0;
  let mut curr = 1;
  for _ in 1..n {
      let next = prev + curr;
      prev = curr;
      curr = next;
  }
  curr
}

fn main() {
  let n = 31;
  let fib_n = fibonacci(n);
  println!("{}", fib_n);
}

```

**To Reproduce**

Run `fib.bend` with a value larger than 30.

**Expected behavior**
Similar output speed to Rust.

**Desktop (please complete the following information):**
OS: macOS
CPU: M2 Pro

**Additional context**
Just wondering why it is slow. Does the program need to be compiled into HVM first and then executed to run fast?

partylikeits1983:
Adding the fib implementation written by @developedby.  This fibonacci sequence implementation is much faster than the previous version that was in the examples directory. 

resolves #367

### DIFF
--- a/examples/fib.bend
+++ b/examples/fib.bend
@@ -1,11 +1,7 @@
-add = λa λb (+ a b)
-
-fib = λx switch x {
-  0: 1
-  _: let p = x-1; switch p {
-    0: 1
-    _: (+ (fib p) (fib p-1))
+fib x =
+  bend x a=1 b=1 {
+     when (!= x 0): (fork (- x 1) b (+ a b))
+     else: a
   }
-}
 
 main = (fib 30)


developedby:
Thank you for your contribution. I think in this example I'll have both implementations and explain the difference between the two, so I won't merge your PR, alright?

PR #380, opened 2024-05-18
Fix/Simplify parallel Hello World

ilevyor:
It looked to me that the example was summing the first 2^28 numbers instead of the first 8 numbers like the sequential example above. (Note I didn't run/test this was just confused while reading the docs)

Btw what is the difference between `bend` and a recursive function? If I'm wrong here it's probably because I'm thinking of bend as just a recursive function.

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -548,7 +548,7 @@ example, to add numbers in parallel, we can write:
 ```python
 def main():
   bend i = 1:
-    when d < 8:
+    when i < 8:
       sum = i + fork(i * 2 + 0) + fork(i * 2 + 1)
     else:
       sum = 0
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -547,11 +547,11 @@ example, to add numbers in parallel, we can write:
 
 ```python
 def main():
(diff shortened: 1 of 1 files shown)


GUIDE.md
```diff
@@ -547,11 +547,11 @@ example, to add numbers in parallel, we can write:
 
 ```python
 def main():
-  bend d = 0, i = 0:
-    when d < 28:
-      sum = fork(d+1, i*2+0) + fork(d+1, i*2+1)
+  bend i = 1:
+    when d < 8:
```
ilevyor, line 551:
```suggestion
    when i < 8:
```

developedby:
The example is indeed adding all numbers up to 2^28, to showcase the power of automatic parallelization, so I won't merge your change. We could rephhrase it to better explain what we're trying to showcase.

Also bend is just a way to inline an anamorphic recursive function as is stated in a couple of different places.

developedby:
I'll close this PR. If you want to improve the wording on the guide, I'll ask you to do it in a new PR to not make the PR have a complicated history

PR #400, opened 2024-05-19
[sc-703] Fix superposition parsing

imaqtkatt:
Fixes superpositions not requiring comma for the first element:

```python
def main:
  {[1, 2] [3, 4, 5, 6]}
```
```
In main.bend :

- expected: Map variable name
- detected:
  12 | return {[1, 2] [3, 4, 5, 6]}
```


src/imp/parser.rs
```diff
@@ -124,7 +124,13 @@ impl<'a> PyParser<'a> {
         }
         let head = self.parse_expr(false)?;
         self.skip_trivia();
-        if self.starts_with(":") { self.parse_map_init(head)? } else { self.parse_sup(head)? }
+        if self.try_consume(",") {
+          self.parse_sup(head)?
+        } else if self.try_consume(":") {
+          self.parse_map_init(head)?
+        } else {
+          self.expected("Comma or Colon")?
```
developedby, line 132:
```suggestion
          self.expected("',' or ':")?
```
To be consistent with how we're displaying this in other situations

developedby (changes requested):

developedby (changes requested):
The tests are failing because victor updated one of the examples without refreshing the tests.

Can you rebase on main and commit the change in the test result?

developedby (approved):

PR #447, opened 2024-05-22
Make eta dont reduce on numbers

imaqtkatt:
Fixes the following program returning `10.000`:

```
main = (10.0, 10.0)
```


developedby (changes requested):
Can you add a test?

developedby (approved):

PR #454, opened 2024-05-22
List Function Examples

ematth:
Example functions for the built-in List type.


examples/list.bend
```diff
@@ -0,0 +1,169 @@
+##############################
+# Author: Ematth, 2024
+##############################
+### Singly-Linked List Type Definition: ###
+
+# type List:
```
developedby, line 6:
```suggestion
# The List type is builtin, so we don't need to declare it.
# But this is how it's defined in the builtins file.
# type List:
```

examples/list.bend
```diff
@@ -0,0 +1,169 @@
...
+#   Cons { head, ~tail }
+#   Nil
```
developedby, line 8:
```suggestion
#   Nil
#   Cons { head, ~tail }
```
The order matters for the compilation output

examples/list.bend
```diff
@@ -0,0 +1,169 @@
...
+List/len/aux = @acc @l
+  match l {
+    List/Nil: acc
+    List/Cons: (List/len/aux (+ acc 1) l.tail)
+  }
+List/len = @l
+  (List/len/aux 0 l)
+
+# List count:
+# List l -> Some s -> uint
```
developedby, line 66: Here `s` is necessarily a number.
To make it generic you could pass a comparison function

examples/list.bend
```diff
@@ -0,0 +1,169 @@
...
+
+# List length:
+# List l -> uint
+# returns the number of elements in list l.
+List/len/aux = @acc @l
+  match l {
+    List/Nil: acc
+    List/Cons: (List/len/aux (+ acc 1) l.tail)
+  }
+List/len = @l
```
developedby, line 62: In general we want functions that consume a potentialy large data structure to be linear (not require duplicating the data).
Here it would mean returning the length + the list.

examples/list.bend
```diff
@@ -0,0 +1,169 @@
...
+      }
+      use y = switch acc {
+        0: List/Nil
+        1: List/Nil
+        _: l.tail
+      }
+      (List/Cons x (List/pop_back/aux (- acc 1) y))
+  }
+List/pop_back = @l
+  switch (- (List/len l) 1) {
```
developedby, line 122: This is a great example of a case where List/len should be linear, otherwise this becomes quadratic in complexity (for each element you're duplicating all the rest of the list just to check its length)

examples/list.bend
developedby: This applies to all the functions where == or another numeric operation with the element is done

developedby (changes requested):
Very good example, thank you

examples/list.bend
ematth: That would be me assuming == working as a general comparison. I'll reserve it for uint for now and generalize comparisons later.

examples/list.bend
ematth: Writing List/len as an auxiliary function was intuitive to me, but only because I didn't find anything immediately for using an inline accumulating value.

ematth: Do you have any recommendations on inline uint accumulation?

ematth:
All issues mentioned in change request have been fixed.

developedby (changes requested):

.vscode/settings.json
developedby, line 1: Can you remove this file?

developedby (changes requested):
Very good, thank you.

Now, you just need to ruun the tests to save a snapshot of t his example.
You have to install cargo-insta (`cargo install cargo-insta`) and then run `cargo insta test`, followed by `cargo insta review`, which should show no changes, except for the result of running this new example.

developedby (approved):

Issue #443, opened 2024-05-22
Top level names with "//" do not run on HVM


developedby:
**Describe the bug**
// is considered by HVM to be a comment, so names with // will break the parser even though / is a valid name character

**To Reproduce**
Try to run this program
```
//a = 1
main = *
```

**Expected behavior**
Either we don't allow it, or we change the HVM parser.
I think the first is better

developedby:
Closed by #458

PR #458, opened 2024-05-23
Dont allow top level names start with '//'

imaqtkatt:
#443


developedby (changes requested):
Bump version to 0.2.17

PR #481, opened 2024-05-23
throw an error on invalid numbers

### LINKED ISSUE #469: Numbers only parse up to the last valid digit
When we write a number like
```
0b012345678
123456789asdfg
0x012di230ft
```
Instead of parsing errors, the parsing of the numbers stops at the last valid character for that number and the parsing of the next symbol starts right after.

A better behaviour would be to read alphanumeric symbols to the end and then try to parse that token.
eg: `0b01234` parsing as an invalid number instead of `0b01` followed by `234`.

Qelxiros:
This closes #469.

### DIFF
--- a/src/fun/parser.rs
+++ b/src/fun/parser.rs
@@ -1090,7 +1090,7 @@ pub trait ParserCommons<'a>: Parser<'a> {
     };
     let num_str = self.take_while(move |c| c.is_digit(radix) || c == '_');
     let num_str = num_str.chars().filter(|c| *c != '_').collect::<String>();
-    if let Some(c) = self.peek_one() && c.is_numeric() {
+    if let Some(c) = self.peek_one() && c.is_ascii_alphanumeric() {
       self.expected("valid digit")
     } else if num_str.is_empty() {
       self.expected("numeric digit")
--- a/src/fun/parser.rs
+++ b/src/fun/parser.rs
@@ -1090,7 +1090,9 @@ pub trait ParserCommons<'a>: Parser<'a> {
     };
     let num_str = self.take_while(move |c| c.is_digit(radix) || c == '_');
     let num_str = num_str.chars().filter(|c| *c != '_').collect::<String>();
-    if num_str.is_empty() {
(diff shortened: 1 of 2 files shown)


src/fun/parser.rs
```diff
@@ -1090,7 +1090,9 @@ pub trait ParserCommons<'a>: Parser<'a> {
     };
     let num_str = self.take_while(move |c| c.is_digit(radix) || c == '_');
     let num_str = num_str.chars().filter(|c| *c != '_').collect::<String>();
-    if num_str.is_empty() {
+    if let Some(c) = self.peek_one() && c.is_numeric() {
```
developedby, line 1093: Shouldn't be just `is_numeric`, but rather `is_alphanumeric` since it will include hex numbers and avoid confusions with names that start with numbers (which are not possible since numbers have higher priority).
```suggestion
    if let Some(c) = self.peek_one() && c.is_ascii_alphanumeric() {
```

src/fun/parser.rs
```diff
@@ -1090,7 +1090,9 @@ pub trait ParserCommons<'a>: Parser<'a> {
...
+      self.expected("valid digit")
```
developedby, line 1094: Instead of just "valid digit", would be better to show the specific radix name, so check the radix and then "valid binary digit", "valid decimal digit" or "valid hexadecimal digit"

src/fun/parser.rs
```diff
@@ -1090,7 +1090,9 @@ pub trait ParserCommons<'a>: Parser<'a> {
...
+    } else if num_str.is_empty() {
       self.expected("numeric digit")
```
developedby, line 1096: We could use the same message here to avoid confusion.

developedby (changes requested):
Thank you!

developedby (approved):

developedby:
Hey, can you just run `cargo fmt` to format the code following the rules we use for this repo?

Qelxiros:
Done, sorry about that. Neovim usually does it for me, but something weird happened with rustup toolchains.

PR #484, opened 2024-05-24
483 nested map gets

imaqtkatt:


src/imp/gen_map_get.rs
```diff
@@ -116,13 +116,14 @@ impl Stmt {
   }
 }
 
-type Substitutions = HashMap<Name, (Name, Box<Expr>)>;
+type Substitutions = IndexMap<Name, (Name, Box<Expr>)>;
```
developedby, line 119: If all we need is to do an rfold, we could have Substitutions be a simple `Vec<(Name, Name, Box<Expr>)>`

developedby (changes requested):

developedby (approved):

PR #487, opened 2024-05-24
478 dont allow variable names to contain "__"

imaqtkatt:


src/fun/parser.rs
```diff
@@ -821,8 +821,17 @@ pub trait ParserCommons<'a>: Parser<'a> {
   }
 
   fn parse_bend_name(&mut self) -> ParseResult<Name> {
+    let ini_idx = *self.index();
     let name = self.take_while(|c| c.is_ascii_alphanumeric() || c == '_' || c == '.' || c == '-' || c == '/');
-    if name.is_empty() { self.expected("name") } else { Ok(Name::new(name.to_owned())) }
+    let end_idx = *self.index();
+    if name.contains("__") {
```
developedby, line 827: You're not applying the same restrictions that you wrote in the documentation (// at the start is allowed).
One of the two should be changed.

I think it's simpler if top-lvel names and variables use the same parsing function with both restrictions. It'll avoid bugs like this one for example

developedby (changes requested):

developedby (approved):

Issue #512, opened 2024-05-27
Put constructor tags behind a reference in num-scott adt encoding


developedby:
Currently, constructors are encoded using a numeric tag to differentiate between them:
```
type Option = (Some val) | None
# becomes
Option/Some = @val @x (x 0 val)
Option/None = @x (x 1)
```

This works well, but it's hard to disambiguate the result of a program.
Without changing behaviour, we could put each tag behind a unique reference, that way we still get their names at the end, and we could possibly use that to rebuild the constructors during readback.
```
type Option = (Some val) | None
# could be
Option/SOME = 0
Option/NONE = 1
Option/Some = @val @x (x Option/SOME val)
Option/None = @x (x Option/NONE)
```

This has the drawback of making pattern matching take 1 extra dereference interaction.

PR #518, opened 2024-05-27
#512 Put constructor tags behind a reference in num scott adt encoding

imaqtkatt:


developedby (changes requested):
Update compiler-options.md to include this new bit in the part about choosing the encoding

docs/compiler-options.md
```diff
@@ -1,17 +1,17 @@
...
-| `-Oeta` `-Ono-eta` | Disabled | [eta-reduction](#eta-reduction) |
-| `-Oprune` `-Ono-prune` | Disabled | [definition-pruning](#definition-pruning) |
-| `-Olinearize-matches` `-Olinearize-matches-alt` `-Ono-linearize-matches` | Enabled  | [linearize-matches](#linearize-matches) |
-| `-Ofloat_combinators` `-Ono-float_combinators` | Enabled  | [float-combinators](#float-combinators) |
-| `-Omerge` `-Ono-merge` | Disabled | [definition-merging](#definition-merging) |
-| `-Oinline` `-Ono-inline` | Disabled | [inline](#inline) |
-| `-Ocheck-net-size` `-Ono-check-net-size` | Disabled | [check-net-size](#check-net-size) |
-| `-Oadt-scott` `-Oadt-num-scott` | adt-num-scott | [adt-encoding](#adt-encoding) | | 
+| flag                                                                     | Default       | What it does?                             |
+| ------------------------------------------------------------------------ | ------------- | ----------------------------------------- | --- |
```
developedby, line 4: I think the extra |---| is breaking the table. If you look at the rich diff, you'll see it

developedby (changes requested):

docs/compiler-options.md
imaqtkatt: I think it is fixed now

developedby (approved):

Issue #471, opened 2024-05-23
 Make the long string tests not have a giant output


developedby:
**Is your feature request related to a problem? Please describe.**
We have a test that compiles a function with a many kilobytes-long string.
It's used to test that very long objects don't cause crashes.
It generates an output of ~800kB, clobbering the git repo and messing with the stats history. Everytime we change something that affects this test, the commit gets a +800k/-800k diff.
This test needs to be changed.

**Describe the solution you'd like**
Move it to its own test, where it runs the whole compilation but doesn't print the result at the end. Bonus, also run this program and readback the giant string, to test that HVM doesn't crash and readback doesn't crash.

PR #522, opened 2024-05-28
#471 Make the long string tests not have a giant output

imaqtkatt:


developedby (approved):

developedby (changes requested):
Can you also move the "huge_tree" file?

PR #534, opened 2024-05-30
Make monadic blocks lazy by defering passing of free vars

developedby:


src/fun/transform/desugar_with_blocks.rs
```diff
@@ -1,20 +1,20 @@
 use crate::{
   diagnostics::Diagnostics,
-  fun::{Ctx, Name, Term},
+  fun::{Ctx, Name, Pattern, Term},
   maybe_grow,
 };
 use std::collections::HashSet;
 
 impl Ctx<'_> {
   /// Converts `ask` terms inside `do` blocks into calls to a monadic bind operation.
```
imaqtkatt, line 9:
```suggestion
  /// Converts `ask` terms inside `with` blocks into calls to a monadic bind operation.
```

src/fun/transform/desugar_with_blocks.rs
```diff
@@ -25,14 +25,14 @@ impl Ctx<'_> {
...
     &mut self,
     cur_block: Option<&Name>,
     def_names: &HashSet<Name>,
   ) -> Result<(), String> {
     maybe_grow(|| {
       if let Term::With { typ, bod } = self {
-        bod.desugar_do_blocks(Some(typ), def_names)?;
+        bod.desugar_with_blocks(Some(typ), def_names)?;
         let wrap_ref = Term::r#ref(&format!("{typ}/wrap"));
         // let wrap_ref = if def_names.contains(&wrap_nam) {
```
imaqtkatt, line 37: Either we remove this commented code or uncomment it, currently if some type does not have the `wrap` definition but is used inside a `with` block it will fail at the definition_pruning transformation

developedby: Shouldn't we check for unbound functions at some point? Never noticed we don't do that

imaqtkatt: I would expect this to be catch at the `check_unbound_vars` after `desugar_with_blocks`

developedby: It's not a variable, so it makes sense. We need a `check_unbound_refs` somewhere near the end of compilation.

imaqtkatt (approved):

PR #539, opened 2024-06-03
Dont expand generated recursive definitions

imaqtkatt:


src/fun/transform/expand_generated.rs
```diff
@@ -1,24 +1,107 @@
...
       }
     })
   }
 }
+
+type DepGraph = HashMap<Name, HashSet<Name>>;
+type Cycles = Vec<Vec<Name>>;
+type RecursiveDefs = BTreeSet<Name>;
+
+pub fn find_recursive_defs(book: &Book) -> RecursiveDefs {
```
developedby, line 30: Can you instead make this 
```
impl Book {
  pub fn recursive_defs(&self) -> RecursiveDefs {
    ...
  }
}
```
To be more in line with how we expose other functions on Book and Term? (like Term::free_vars for example)

src/fun/transform/expand_generated.rs
```diff
@@ -1,24 +1,107 @@
...
+    // Check if the current ref is already in the stack, which indicates a cycle.
+    if let Some(cycle_start) = stack.iter().position(|n| n == nam) {
+      // If found, add the cycle to the cycles vector.
+      cycles.push(stack[cycle_start..].to_vec());
+      return;
+    }
+    // If the ref has not been visited yet, mark it as visited.
+    if visited.insert(nam.clone()) {
+      // Add the current ref to the stack to keep track of the path.
+      stack.push(nam.clone());
```
developedby, line 74: Since this function is already stack based, you could make it a while loop and not recursive right?

developedby (approved):

developedby (changes requested):
Also, increase the version

developedby (approved):

PR #546, opened 2024-06-05
528 Add syntax for a builtin Tree type

developedby:


imaqtkatt:
Also add the syntax to doc/syntax

developedby:
> Also add the syntax to doc/syntax

I thought I did because syntax.md was in the diff... Aparently not

PR #547, opened 2024-06-05
Update README

Sipher:
The current readme isn't very clear, therefore, I suggest a change in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,81 +1,192 @@
-# Bend
+<h1 >Bend</h1>
+<p>A high-level, massively parallel programming language</p>
 
-Bend is a massively parallel, high-level programming language.
+## Index
+1. [Introduction](#introduction)
+2. [Important Notes](#important-notes)
+3. [Install](#install)
+4. [Getting Started](#getting-started)
+5. [Speedup Example](#speedup-example)
+6. [Additional Resources](#additional-resources)
 
-Unlike low-level alternatives like CUDA and Metal, Bend has the feeling and
-features of expressive languages like Python and Haskell, including fast object
-allocations, higher-order functions with full closure support, unrestricted
-recursion, even continuations. Yet, it runs on massively parallel hardware like
-GPUs, with near-linear speedup based on core count, and zero explicit parallel
-annotations: no thread spawning, no locks, mutexes, atomics. Bend is powered by
-the [HVM2](https://github.com/HigherOrderCO/hvm) runtime.
+## Introduction
 
-A Quick Demo
-------------
+Bend offers the feel and features of expressive languages like Python and Haskell. This includes fast object allocations, full support for higher-order functions with closures, unrestricted recursion, and even continuations.                             
+Bend scales like CUDA, it runs on massively parallel hardware like GPUs, with nearly linear acceleration based on core count, and without explicit parallelism annotations: no thread creation, locks, mutexes, or atomics.                     
+Bend is powered by the [HVM2](https://github.com/higherorderco/hvm) runtime.
 
-[![Bend live demo](https://github.com/VictorTaelin/media/blob/main/bend_live_demo.gif?raw=true)](https://x.com/i/status/1791213162525524076)
 
-- For a more in-depth explanation on how to setup and use Bend, check [GUIDE.md](https://github.com/HigherOrderCO/bend/blob/main/GUIDE.md).
+## Important Notes
 
-- For an extensive list of features, check [FEATURES.md](https://github.com/HigherOrderCO/bend/blob/main/FEATURES.md).
+* Bend is designed to excel in scaling performance with cores, supporting over 10000 concurrent threads.
+* The current version may have lower single-core performance.
+* You can expect substantial improvements in performance as we advance our code generation and optimization techniques.
+* We are still working to support Windows. Use [WSL2](https://learn.microsoft.com/en-us/windows/wsl/install) as an alternative solution.
+* [We only support NVIDIA Gpus currently](https://github.com/HigherOrderCO/Bend/issues/341).
 
-## Using Bend
 
-> Currently not working on Windows, please use [WSL2](https://learn.microsoft.com/en-us/windows/wsl/install) as a workaround.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -95,76 +206,71 @@ def rots(d, s, tree):
...
-single-core performance is still extremely sub-par. This is the first version of
-the system, and we haven't put much effort into a proper compiler yet. You can
-expect the raw performance to substantially improve on every release, as we work
-towards a proper codegen (including a constellation of missing optimizations).
-Meanwhile, you can use the interpreters today, to have a glimpse of what
-massively parallel programming looks like, from the lens of a Pythonish,
-high-level language!
+### Additional Resources
+ - To understand the technology behind Bend, check out the HVM2 [paper](https://docs.google.com/viewer?url=https://raw.githubusercontent.com/HigherOrderCO/HVM/main/paper/PAPER.pdf). Bend is developed by [HigherOrderCO](https://higherorderco.com/) - join our [Discord](https://discord.gg/kindelia)!
+ - Watch the [live demo video](https://x.com/i/status/1791213162525524076).
```
developedby, line 276: There is no point in putting the live demo right at the end. Either it should be at the start or not here at all

README.md
```diff
@@ -1,81 +1,192 @@
...
+
+In this version, the next value to be calculated depends on the previous sum, meaning that it cannot proceed until the current computation is complete. Now, let's look at the easily parallelizable version.
+
 
-## Parallel Programming in Bend
+#### Parallelizable version:
+First close the old file and then proceed to your terminal to create `psum.bend`
+```sh
+# Write this command on your terminal
+touch psum.bend
```
developedby, line 138: use a decent name for the files (parallel_sum, sequential_sum, etc)

README.md
```diff
@@ -95,76 +206,71 @@ def rots(d, s, tree):
...
-(i.e., scale in performance with cores, up to 10000+ concurrent threads), its
-single-core performance is still extremely sub-par. This is the first version of
-the system, and we haven't put much effort into a proper compiler yet. You can
-expect the raw performance to substantially improve on every release, as we work
-towards a proper codegen (including a constellation of missing optimizations).
-Meanwhile, you can use the interpreters today, to have a glimpse of what
-massively parallel programming looks like, from the lens of a Pythonish,
-high-level language!
+### Additional Resources
+ - To understand the technology behind Bend, check out the HVM2 [paper](https://docs.google.com/viewer?url=https://raw.githubusercontent.com/HigherOrderCO/HVM/main/paper/PAPER.pdf). Bend is developed by [HigherOrderCO](https://higherorderco.com/) - join our [Discord](https://discord.gg/kindelia)!
```
developedby, line 275: Is this a single line on purpose? It's two topics

developedby, line 275: This is not the right link to the discord, right?

README.md
```diff
@@ -1,81 +1,192 @@
...
+# For the C version of Bend, use GCC. We recommend a version up to 12.x.
+sudo apt install gcc
+```
+For the CUDA runtime [install the CUDA toolkit for Linux](https://developer.nvidia.com/cuda-downloads?target_os=Linux) version 12.x.
 
-> **_Note_: [Only Nvidia GPUs are supported at the moment](https://github.com/HigherOrderCO/Bend/issues/341).**
 
-Then, install both HVM2 and Bend with:
+#### On Mac
+```py
```
developedby, line 45: use `sh` highlighting

developedby (changes requested):
You removed what is arguably the most important part, which is making users read either GUIDE.md or FEATURES.md

README.md
Sipher: It actually is hahaha

README.md
Sipher: I am not opposed to not having it anywhere. just quoted here, but, I rather NOT

README.md
Sipher: Okay, will edit

README.md
developedby: We should probably work on that then

cspell.json
```diff
@@ -111,6 +111,10 @@
     "undefer",
     "vectorize",
     "vectorizes",
+    "parallel_sum",
```
developedby, line 114: You don't need to update the dictionary here, it already ignores things between ```.
Also, `parallel`, `sum` and `sequential` are known words, so no need to add dictionar exceptions.

developedby (changes requested):

README.md
```diff
@@ -95,76 +206,71 @@ def rots(d, s, tree):
...
+  return sum(20, sort(20, 0, gen(20, 0)))
 ```
 
-And it would actually work. Even involved algorithms parallelize well on Bend.
-Long-distance communication is performed by *global beta-reduction* (as per the
-[Interaction Calculus](https://github.com/VictorTaelin/Interaction-Calculus)),
-and synchronized correctly and efficiently by
-[HVM2](https://github.com/HigherOrderCO/HVM)'s *atomic linker*.
-
+#### Benchmark
```
developedby, line 264: This section is only shown when users click the summary. You wanted only the code to be collapsed right?

developedby (changes requested):

README.md
Sipher: Fixed

developedby:
Supercedes #540 and #365.

cspell.json
```diff
@@ -111,7 +111,9 @@
     "undefer",
     "vectorize",
     "vectorizes",
-    "walkdir",
+    "tlsv",
+    "proto",
```
developedby, line 115:
```suggestion
    "tlsv",
    "proto",  
```
I think you can remove these lines since they're only used inside ``` quotes which are being ignored by cspell

developedby (changes requested):

cspell.json
Sipher: Removed and Cspell is blaming me for using unknown words :(

Sipher: Added again and now its working, hes not hating me anymore

developedby: It's only ignoring blocks that are exactly 2 lines long for some reason. I'll think of the appropriate regex some other time.
Let's just add those two words

developedby (approved):

Issue #485, opened 2024-05-24
Improve the error message when redefining a builtin function


developedby:
**Is your feature request related to a problem? Please describe.**
When redefining a builtin function, the error message just shows `Redefinition of function 'XXX'`. This is a bit confusing if the user doesn't know that XXX is a builtin, since it seems to be coming out of nowhere and doesn't point to the first definition.

**Describe the solution you'd like**
It would be better if the message showed `Redefinition of builtin (function) 'XXX'`.

**Describe alternatives you've considered**
Storing the span of functions and printing the context of the first definition. This would greatly increase the complexity of the parser and slightly increase the complexity of the AST, so I"d rather not do it unless we really need to.

PR #567, opened 2024-06-07
#485 Improve the error message when redefining a builtin function

imaqtkatt:


developedby (commented):
I can't remember anymore since it's been a while i wrote the issue, but is this error message also bad for redefining types/constructors.
If so, we should improve it in this PR

developedby (approved):

PR #591, opened 2024-06-17
Call desugar_use before linearize matches

imaqtkatt:


src/lib.rs
```diff
@@ -128,7 +129,6 @@ pub fn desugar_book(
   ctx.check_unbound_vars()?;
 
   ctx.book.make_var_names_unique();
-  ctx.book.desugar_use();
```
developedby, line 131: Doesn't linearize matches insert some `use` terms?
We still need to call `desugar_use` after it, no?

src/lib.rs
```diff
@@ -128,7 +129,6 @@ pub fn desugar_book(
   ctx.check_unbound_vars()?;
 
   ctx.book.make_var_names_unique();
```
developedby, line 131: This line is useless now, we're calling it twice in a row

developedby (changes requested):
You also need to update the `simplify_matches` and `encode_pattern_match` tests

developedby (approved):

Issue #580, opened 2024-06-11
Add string encoding functions


developedby:
We need functions to convert encoded Lists of bytes into Bend strings and vice versa.
They are necessary to use file IO, to read/write text files.

We need to support at least ascii and utf-8, in the future we can add other encodings.

PR #594, opened 2024-06-18
#580 Add string encoding functions

imaqtkatt:


src/fun/builtins.bend
```diff
@@ -175,3 +175,120 @@ hvm to_u24:
...
+      }
+    }
+  }
+
+String/encode_utf8 (String/Nil)       acc = acc
+String/encode_utf8 (String/Cons x xs) acc =
+  if (<= x Utf8/rune1max) {
+    (String/encode_utf8 xs (Encoding/List/append x acc))
+  } else {
+    if (<= x Utf8/rune2max) {
```
developedby, line 270: did we not add `elif` to fun syntax?

src/fun/builtins.bend
```diff
@@ -175,3 +175,120 @@ hvm to_u24:
...
   ($([i24] ret) ret)
+
+# Encoding
+
+Utf8/MAX_RUNE = '\u{0010FFFF}'
+Utf8/RUNE_ERROR = '\u{FFFD}'
+#Utf8/RUNE_SELF = 0x80
+#Utf8/MAX = 4
+
+Utf8/t1 = 0b00000000
```
developedby, line 186: This is not used

src/fun/builtins.bend
```diff
@@ -175,3 +175,120 @@ hvm to_u24:
...
+Utf8/tx = 0b10000000
+Utf8/t2 = 0b11000000
+Utf8/t3 = 0b11100000
+Utf8/t4 = 0b11110000
```
developedby, line 190: Can you convert these 4 into `use` terms? We should avoid polluting the builtins with functions that are not useful for users.

src/fun/builtins.bend
```diff
@@ -175,3 +175,120 @@ hvm to_u24:
 # Casts any native number to an i24.
 hvm to_i24:
   ($([i24] ret) ret)
+
+# Encoding
+
+Utf8/MAX_RUNE = '\u{0010FFFF}'
+Utf8/RUNE_ERROR = '\u{FFFD}'
+#Utf8/RUNE_SELF = 0x80
+#Utf8/MAX = 4
```
developedby, line 184: What are these two commented bits?

src/fun/builtins.bend
```diff
@@ -175,3 +175,120 @@ hvm to_u24:
 # Casts any native number to an i24.
 hvm to_i24:
   ($([i24] ret) ret)
+
+# Encoding
+
+Utf8/MAX_RUNE = '\u{0010FFFF}'
```
developedby, line 181: This is not utf8, but rather unicode in general

src/fun/builtins.bend
```diff
@@ -175,3 +175,120 @@ hvm to_u24:
...
+Utf8/RUNE_ERROR = '\u{FFFD}'
```
developedby, line 182: This is officially called the replacement character in unicode

src/fun/builtins.bend
```diff
@@ -175,3 +175,120 @@ hvm to_u24:
...
+Utf8/t4 = 0b11110000
+
+Utf8/maskx = 0b00111111
+Utf8/mask2 = 0b00011111
+Utf8/mask3 = 0b00001111
+Utf8/mask4 = 0b00000111
+
+Utf8/rune1max = (- (<< 1 7) 1)
+Utf8/rune2max = (- (<< 1 11) 1)
+Utf8/rune3max = (- (<< 1 16) 1)
```
developedby, line 199: Do these need to be functions? Do we expect to use them in other places?
If no then we can make them into `let` or `use` terms.

src/fun/builtins.bend
```diff
@@ -175,3 +175,120 @@ hvm to_u24:
...
+
+Encoding/List/append x []               = [x]
```
developedby, line 201: This is pretty bad because inserting at the end through pattern matching is `O(len)`, making encoding be `O(len^2)`.
Instead we should either write thing in the order they appear and reverse at the end, like this:
```
String/encode_utf8 (String/Nil)       acc = (List/reverse acc)
String/encode_utf8 (String/Cons x xs) acc =
  if (<= x Utf8/rune1max) {
    (String/encode_utf8 xs (List/Cons x acc))
  ...
```
Or use a lambda terminated list like this:
```
let str = @x x
# This appends a character to the end
let str = @x (str (List/Cons char x))
# This appends multiple characters to the end
let str = @x (str (List/Cons char1 (List/Cons char2 (List/Cons char3 x))))
# This appends a character to the start
let str = @x (List/Cons char (str x))
# This finalizes it, converting into a normal list.
let str = (str List/Nil)
```
```

File: docs/builtins.md [hunk 9]
```diff
@@ -297,33 +342,86 @@ Returns nothing (`*`).
...
+```python
+def String/encode_ascii(s: String) -> [u24]
+```
+
+Encodes a String to a sequence of bytes using ascii encoding.
+
+### Utf8/decode_rune
+
+```python
+def Utf8/decode_rune(bytes: [u24]) -> (rune: u24, length: u24, rest: [u24])
```
developedby, line 423: Do you think returning the length is going to be used by users for something useful?
It could be easily derived from the value of the code point and in our code we're not using it for anything, so i think we can remove this return.

docs/builtins.md
```diff
@@ -297,33 +342,86 @@ Returns nothing (`*`).
...
+
+### String/encode_ascii
+
+```python
+def String/encode_ascii(s: String) -> [u24]
+```
+
+Encodes a String to a sequence of bytes using ascii encoding.
+
+### Utf8/decode_rune
```
developedby, line 420: Do we want to commit to calling unicode codepoints "runes"?
Not saying we shouldn't, but i'd like to hear your reasoning.

developedby (changes requested):

docs/builtins.md
```diff
@@ -278,16 +321,18 @@ Writes `bytes`, a list of U24 with each element representing a byte, to the file
 
 Returns nothing (`*`).
 
-Writing discards any preexisting content that came after the current position. For example, if your file contains the text `Hello, world!` and the current position is at the `,`, writing `!` will 
+Writing discards any preexisting content that came after the current position. For example, if your file contains the text `Hello, world!` and the current position is at the `,`, writing `!` will
```
developedby, line 324:
```suggestion
```
I started to write this, tested and saw that it was not true, but forgot to erase this line.

developedby (changes requested):

src/fun/builtins.bend
imaqtkatt: No

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project does not currently adhere to a particular versioning scheme.
 - Add `to_f24`, `to_u24` and `to_i24` number casting builtin functions. ([#582][gh-582])
 - Add `IO/sleep` builtin function to sleep for a given amount of seconds as a float. ([#581][gh-581])
 - Add primitive file IO functions `IO/FS/{read, write, seek, open, close}`. ([#573][gh-573])
+- Add encoding/decoding builtin functions `Bytes/{decode_utf8, decode_ascii} String/{encode_ascii, decode_ascii} Utf8/{decode_rune, REPLACEMENT_CHARACTER}`. ([#580][gh-580])
```
developedby, line 27: You renamed decode_rune to decode_character but forgot to update here

Issue #596, opened 2024-06-20
Add elif terms to fun syntax


developedby:
In https://github.com/HigherOrderCO/Bend/issues/427 we added `elif` statements to imp syntax, but didn't add them to fun syntax.
This doesn't really make a lot of sense, we should have it in both syntaxes.

PR #597, opened 2024-06-20
#596 Add elif terms to fun syntax

imaqtkatt:


docs/syntax.md
```diff
@@ -939,6 +939,20 @@ switch _ = condition {
 }
 ```
 
+It is possible to make if-chains using `elif`:
+
+```rust
+if condition1 {
+  0
+} condition2 {
```
developedby, line 947: you didn't write the `elif` here

developedby (changes requested):

developedby (approved):

Issue #579, opened 2024-06-11
Update print/input functions for new file IO primitives


developedby:
File IO functions are added in https://github.com/HigherOrderCO/Bend/issues/573.

We need to update print and input functions to use the new primitive file io functions.

HVM assigns fd 0 to stdin, 1 to stdout and 2 to stderr. We need to write to those files.

I think we should also add an `eprint` to print to stderr, maybe println and eprintln variants.

Print should receive a string and encode it as utf8.
Input should read a string from stdin and decode it as utf8.

Possibly add a getc or similar for reading single characters.

developedby:
Closed by #599

PR #599, opened 2024-06-21
#578 Add File IO utility functions

### LINKED ISSUE #578: Add user friendly file IO functions
https://github.com/HigherOrderCO/Bend/issues/573 adds the primitive file IO functions for Bend, building on the IO functions that HVM exposes.
However, for the common use case they are a bit hard to use.

So we'd like more usuer-friendly functions. I propose these, but we could have others:
- `IO/Fs/read_file: (path: String) -> (List U24)`, reads a file in its entirety and returns a list of bytes.
- `IO/Fs/read_until_end: (file: U24) -> (List U24)`, takes a file descriptor as returned by `IO/Fs/open` and reads it until EOF.
- `IO/Fs/read_line: (file: U24) -> (List U24)`, takes a file descriptor and reads a line until `\n` is found.
- `IO/Fs/write_file: (path: String) -> (data: List U24) -> *`, replaces the entire content of a file with some contents (or writes a new file).

Since reads have a maximum length imposed by hvm, the read functions should read in chunks and then stitch them together.

`read_line` can read a large chunk, check for \n, if found split the string, and then move back the tail length using `seek`. I think this will be a bit faster than reading char by char.

I think these functions should operate with lists of bytes, so they'll work with both binary file and encoded text files.

developedby:
Closes #578 and #579 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project does not currently adhere to a particular versioning scheme.
 ### Fixed
 
 - Fix readback of numeric operations. ([#467][gh-467])
+- Propagate the "builtin" attribute of definitions when extracting functions from `bend` and `fold` syntax.
 
 ### Added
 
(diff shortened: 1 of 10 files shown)


imaqtkatt (changes requested):
Update docs/builtins with the added functions

CHANGELOG.md
```diff
@@ -348,6 +352,8 @@ and this project does not currently adhere to a particular versioning scheme.
 [gh-528]: https://github.com/HigherOrderCO/Bend/issues/528
 [gh-581]: https://github.com/HigherOrderCO/Bend/issues/581
 [gh-573]: https://github.com/HigherOrderCO/Bend/issues/573
+[gh-578]: https://github.com/HigherOrderCO/Bend/issues/578
+[gh-578]: https://github.com/HigherOrderCO/Bend/issues/579
```
imaqtkatt, line 356:
```suggestion
[gh-579]: https://github.com/HigherOrderCO/Bend/issues/579
```

imaqtkatt (approved):

PR #602, opened 2024-06-24
#564 Add notes on square brackets syntax

### LINKED ISSUE #564: Add a note in documentation that val[key] is not a syntax for Lists
It seems to confuse some users that try to `[]` access a list.
We create lists with `[]`, but for getting values `[]` actually refers to maps.

This should be presented when we first show Lists and maybe an extra note should also be added when explaining this syntax for Maps.

cstiago:
Addition of notes to improve clarity on GUIDE.md and FEATURES.md regarding the square brackets syntax difference of lists and maps on Bend.

Fix #564.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -374,6 +374,7 @@ type Map:
 ```
 
 Notice that the getter and setter syntax induces an order on things using the map, since every get or set operation depends on the value of the previous map.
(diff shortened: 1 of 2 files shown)


developedby (approved):

developedby:
Thank you, we also wanted to add this clarification since other users have had trouble with this

Issue #470, opened 2024-05-23
Add fast access to pred variable in equational pattern matching on numbers


developedby:
When we a function like
```
Foo 0 = A
Foo 1 = B
Foo 2 = C
Foo x = ... (- x 3) ...
```
We could substitute `(- x 3)` for the matched variable since it already has that value.
Currently, it will compile to `(- (+ a 3) 3)`, instead we could just make it `a`.

PR #603, opened 2024-06-24
#470 Add fast access to pred variable

imaqtkatt:


developedby (approved):

src/fun/transform/desugar_match_defs.rs
```diff
@@ -301,8 +302,10 @@ fn num_rule(
       let mut body = rule.body.clone();
       if let Some(var) = var {
         let last_num = *nums.last().unwrap();
-        let var_recovered = Term::add_num(Term::Var { nam: pred_var.clone() }, Num::U24(1 + last_num));
+        let curr_num = 1 + last_num;
```
developedby, line 305:
```suggestion
        let cur_num = 1 + last_num;
```
Let's go with `cur` to mean "current" everywhere in the repo. (this is the only place that uses `curr`)

developedby (changes requested):

developedby (approved):

Issue #468, opened 2024-05-23
Patterns in "fun" rules shouldn't accept unscoped variables


developedby:
```
foo $x = $x
```
The definition above reaches unreachable, it should not be accepted while parsing it

developedby:
Actually, maybe we SHOULD allow them, since they can be used to disable linearization of that variable. In case we decide to do that, we need to fix the unreachable

PR #605, opened 2024-06-25
#468 Patterns in "fun" rules should accept unscoped variables

imaqtkatt:


developedby (changes requested):
Can you add a test that uses an unscoped in this position?

Issue #562, opened 2024-06-07
Add lambdas with statements / local functions in imp syntax


developedby:
Currently, in imp syntax there is no way of writing lambdas that have statements in them.
We can raise them to proper top-level functions, but then any captured variables must be passed through additional arguments, which increases the overhead.
But the main problem is that it's very inconvenient to do so. Instead we'd like a way to write lambdas with statements like we can do in Fun syntax.

I'll propose two possible syntaxes here.

```
def foo:
  lambda x:
    ...
    return y
  return z
```
Here `lambda x` defines a block for a local anonymous function bound to variable `x`.
I thought of this as just a simple lambda/closure, that can freely capture variables and doesn't get lifted into a separate definition.

Another option is
```
def foo:
  def foo/local(a, b):
    ...
    return c
  return foo/local(x, y)
```
This defines a function that is raised as a top-level. Captured variables are passed as arguments implicitly through a `use` term, like this
```
def foo:
  use foo/local = foo__local_foo/local(captured1, captured2)
  ...
  return foo/local(x, y)
```
This allow local functions to be recursive, but adds an overhead compared to just a simple lambda. And if the lambda is affine, it will get floated anyway.

I don't know which is better, need to think more about it.
I think we don't need both syntaxes, but that could also be an option.

Also, note that either syntax could have either of the described behaviours, I just think that they sort of imply intuitively the behaviours i described.

developedby:
Another advantage of the second syntax is that it would be similar to python (except for the lack of `nonlocal` and `global`)

arnfaldur:
I apologize for the off-topic comment.

When I first saw Bend, during the recent hype cycle, I was very excited.
I saw that it wasn't quite ready for anything I would want to use it for at the time.
But there was huge obvious potential.

Looking at the issues section, I'm in awe of your diligence. There is obviously great work happening here and I just want to leave you with these words of encouragement. I'm very impressed, and you should be proud of all this progress.

developedby:
> I apologize for the off-topic comment.
> 
> When I first saw Bend, during the recent hype cycle, I was very excited. I saw that it wasn't quite ready for anything I would want to use it for at the time. But there was huge obvious potential.
> 
> Looking at the issues section, I'm in awe of your diligence. There is obviously great work happening here and I just want to leave you with these words of encouragement. I'm very impressed, and you should be proud of all this progress.

[...]

PR #609, opened 2024-06-28
#562 Local functions in imp syntax

imaqtkatt:


src/imp/lift_local_defs.rs
```diff
@@ -0,0 +1,129 @@
...
+    &mut self,
+    parent: &Name,
+    defs: &mut IndexMap<Name, fun::Definition>,
+    gen: &mut usize,
+  ) -> Result<(), String> {
+    match self {
+      Stmt::LocalDef { .. } => {
+        let Stmt::LocalDef { mut def, mut nxt } = std::mem::take(self) else { unreachable!() };
+        let children = def.lift_local_defs(gen)?;
+        nxt.lift_local_defs(parent, defs, gen)?;
```
developedby, line 28: These two lines will generate a strange order for the extracted functions. Nested definitions are lifted first, then the ones in the same depth are lifted bottom up.

Also, the generated names are difficult to understand, since local functions with the same name will have similar final names.
It would be a bit easier to read the output if they were named like this:
```py
def Foo():
  def A():
    def X():
  def B():
    def Y():
```
compiling to
```
Foo
Foo__local_0_A
Foo__local_0_A__local_0_X
Foo__local_1_B
Foo__local_1_B__local_0_Y
```

What you need to do is mostly just change the order of operations and reset `gen` for each layer of nested defs.

tests/golden_tests/desugar_file/main_aux.bend
```diff
@@ -0,0 +1,9 @@
+def main:
+  y = 89
+  def aux(x):
```
developedby, line 3: Can we have an example with multiple local defs of same nesting depth? With equal and different names.
Also try calling them to see if the correct name is being used when we shadow one local def with another.

developedby (changes requested):
I thought you'd go with the approach of doing this as a fun transformation, but this is also fine.
If we implement local functions in fun syntax in the future we might wanna move this over to have only one implementation of this pass.

tests/snapshots/desugar_file__local_def_shadow.bend.snap
```diff
@@ -0,0 +1,13 @@
...
+---
+(main__local_0_A__local_0_B) = 0
+
+(main__local_1_A__local_1_B) = 1
+
+(main__local_1_A) = λa a
+
+(main__local_0_A) = λa a
+
+(main) = (main__local_1_A main__local_1_A__local_1_B)
```
developedby, line 13: It shouldn't be capturing the inner definition as closure.

developedby (changes requested):

developedby (approved):

Issue #570, opened 2024-06-08
(Request) Trigonometry and more math functions?


dogevspenguin:
I am wondering how would I be able to use trigonometric functions, As Bend looked quite promising for a speed-up In a project I've been doing, But it requires the usage of trigonometric functions such as sin, cos tan, and even inverse function. So is it possible to add these math functions?

developedby:
We can implement them as an approximation in Bend using the already existing operations, but I think they should exist as HVM operations.
I created this issue to track that https://github.com/HigherOrderCO/HVM/issues/384, but if you want to use them right now look into the approximations like I said.

If by inverse you mean x^-1, you can already do that with the `**` (pow) operator.

If by inverse you mean arcsin, arccos and arctan, HVM already implements atan2, it's just not exposed in Bend.
You can hack them in by using the internal representation of bitwise operations.
* `a & b` = atan2(a, b), arctan(a / b)
* `a | b` = log(a, b), log_b(a)
* `a ^ b` = pow(a, b), a^b

developedby:
With https://github.com/HigherOrderCO/HVM/issues/384 done, we can now work on implementing it on Bend.

developedby:
HVM implements `sin(x + y)` using the `[<<]` operator and `tan(x + y)` with the `[>>]` operator.
We can implement them like we did the others, using a native hvm definition for `sin(x) = (<< x 0.0)`. Cosine can be the same thing but + pi/2.

PR #616, opened 2024-07-04
#570 trigonometry and math functions

imaqtkatt:


src/fun/builtins.bend
```diff
@@ -383,3 +383,60 @@ Bytes/decode_ascii (List/Nil)       = (String/Nil)
...
+Math/PI = 3.1415926535
+
+# Math/sin(a: f24) -> f24
+# Computes the sine of the given angle in radians.
+hvm Math/sin:
+  ($([<<0x0] a) a)
+
+# Math/cos(a: f24) -> f24
+# Computes the cosine of the given angle in radians.
+Math/cos a = (Math/sin (- (/ Math/PI 2.0) a))
```
developedby, line 399: A slightly faster implementation would be to implement this using the underlying hvm sine with -pi/2 as the second argument.

src/fun/builtins.bend
```diff
@@ -383,3 +383,60 @@ Bytes/decode_ascii (List/Nil)       = (String/Nil)
...
+
+# Math/tan(a: f24) -> f24
+# Computes the tangent of the given angle in radians.
+hvm Math/tan:
+  ($([>>0x0] a) a)
+
+# Math/cot(a: f24) -> f24
+# Computes the cotangent of the given angle in radians.
+Math/cot a = (/ (Math/cos a) (Math/sin a))
```
developedby, line 408:
```suggestion
Math/cot a = (** (Math/tan a) -1.0)
```
to avoid duplicating `a`

developedby: or `(/ 1.0 (Math/tan a))`, don't know if either is better

src/fun/builtins.bend
```diff
@@ -383,3 +383,60 @@ Bytes/decode_ascii (List/Nil)       = (String/Nil)
...
+  i_n = to_f24(to_i24(n))
+  if n == i_n:
+    return i_n
+  else:
+    return i_n + 1.0
+
+# Math/floor(n: f24) -> f24
+# Round float down to the nearest integer.
+def Math/floor(n):
+  i_n = to_f24(to_i24(n))
```
developedby, line 438: `to_i24` already rounds down, no?

developedby (changes requested):
Can you add a round operation as well?

developedby (changes requested):
Can you also add `atan`, `asin` and `acos`? You can calculate them from atan2

src/fun/builtins.bend
imaqtkatt: Yes, but it does not work properly for negative values

developedby: Can you add some tests for negative values then? Also tests with already integer values (integers stored as floats, like 1.0, 2.0, etc)

developedby (approved):

Issue #623, opened 2024-07-07
Unbound error with correct code


imaqtkatt:
### Reproducing the behavior

```bend
def foo(x):
  def foo.aux(x, y):
    match x:
      case List/Nil:
        return y
      case List/Cons:
        return x.head
  return foo.aux(x, 0)

def main:
  return foo([1, 2, 3])
  ```
Expected:
```
Result: 1
```
Got:
```
Errors:
In definition 'foo':
  Unbound variable 'x.head'
```

### System Settings

bend-lang 0.2.36

Issue #595, opened 2024-06-19
Add multiline comment syntax


developedby:
**Is your feature request related to a problem? Please describe.**
We'd like to have multiline comment blocks so that we can use them to auto-generate documentation.

It'd allow us to identify what are normal comments and what are documentation comments just syntactically.
It also makes it easier for users to write it without needing to insert a comment in every line.

They can also be used to easily comment out a block of code.

**Describe the solution you'd like**
`"""..."""` as a syntax for block comments. This is the syntax python uses.

Block comments can be either a statement or a top-level. They shouldn't be an expression or a term since we might want to reserve it for multiline strings (even though current strings are already multiline, we might want to change that in the future). It would also diverge from python, where """ does not mean a multiline comment in the middle of an expression.

**Describe alternatives you've considered**
- Other multiline comment separators like `###...###`, `##..##`, `#*...*#`, `#<...>#`, `#{...}#` etc. I don't particularly like any of them other than ### and we can go with the python behaviour here.
- Not having multiline comments. This is fine and is done in other languages. Doc comments can use an additional marker or use their position regarding the function declaration.

PR #625, opened 2024-07-08
Add multi line comment syntax

imaqtkatt:


developedby:
Closes #595

developedby (commented):
I'm not sure about `#{ }#`.
I understand why but it just looks so ugly.

imaqtkatt:
I slightly would prefer `#{ ... #}` over `#{ ... }#`, what do you think?

developedby:
Every language does this differently, there isn't any common way to do it other than `/* */`.

If i were to choose, I'd go with `#= =#` like in julia, even if it doesn't look like it's opening a block, or with `#[ ]#` like in nim, even if we don't really use `[ ]` with this kind of meaning elsewhere.

Relevant: https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(syntax)#Comment_comparison

Do you have any preference or reason for any of the options? If not, we'll keep `#{ }#`

> I slightly would prefer `#{ ... #}` over `#{ ... }#`, what do you think?

I like it

developedby:
Does this also fix https://github.com/HigherOrderCO/Bend/issues/623 ?

imaqtkatt:
Yes, it does fix the issue

developedby (approved):

Issue #513, opened 2024-05-27
Improve error message when the input file is not found


developedby:
Currently, we just forward the message given by `std::fs::read_to_string`: `No such file or directory (os error 2)`.
We should instead give an error that more properly explains what went wrong.

Also, this same error message will show up when the HVM doesn't find its input file (in the case of the intermediary .out.hvm), so it's even more confusing

PR #628, opened 2024-07-10
#513 Improve error message when the input file is not found

imaqtkatt:


developedby (changes requested):
Can you add a cli test for this?

Issue #615, opened 2024-07-03
Add a type system to Bend


developedby:
**Is your feature request related to a problem? Please describe.**
Bend already assumes by the semantics of ADTs and matches that users are writing programs within the contraints of some kind of polymorphic lambda calculus type system.

This is not enforced anywhere since bend is currently untyped. This allows very easily writing very clever programs that escape a simple type system, but in most cases just leads to writing incorrect programs. It'd be better if there was some way of enforcing it.

Having it be untyped like it currently is makes Bend basically unusable for anything not super tiny.

**Describe the solution you'd like**
Add optional hindler-milner typing to Bend. Types are created by adt defs.

Needs to have support for axiomatic type statements to allow unscoped variables, superpositions and native hvm definitions to be used with typed code.

Untyped functions have type `Any`, meaning that they are trusted, like in typescript or mypy.

For Imp syntax, a syntax like typehints in python could work well
```py
def and3(a: Bool, b: Bool, c: Bool) -> Bool:
  x1: Bool = and(a, b)
  return and(x1, c)
```

For Fun syntax, we can have separate function type declaration like in haskell for example.
```
# This is hard to parse, but I like how it looks similar to the other one
(and3 (a: Bool) (b: Bool) (c: Bool)) -> Bool 
(and3 a b c) =
  # Parens are not necessary here
  let (x1: Bool) = (and2 a b);
  (and2 x1 c)

# This is easier to parse, but i don't like how it looks
and3: Bool -> Bool -> Bool -> Bool
and3 a b c =
  # We could also have annotation terms instead
  # Parens are also not necessary here either
  let x1 = ((and2 a b) : Bool);
  (and2 x1 c)
```

We can compile to Kind to typecheck. That would need some trickery to deal with pairs, sups and unscopeds, but should be possible. Functions that have unscopeds need to be untyped or typed with an axiom.

This leaves a couple things still open:
What syntax to use for type variables. I like these two options
```
def List/reverse<t>(list: List(t)) -> List(t):

def List/reverse(~t, list: List(t)) -> List(t):
```

What syntax to use for unchecked type statements. For example we could use `::` instead of `->` in Imp syntax for an unchecked type.
```
# Works, but a bit too subtle
def channel<a, b>() :: (a -> b, b -> a):
  return (lambda $a: $b, lambda $b: $a)

# Using '->' for type and a prefix marker for unchecked type
def !channel<a, b> -> (a -> b, b -> a):
[...]

PR #631, opened 2024-07-15
#615 Add an optional type system to Bend

developedby:


developedby:
This is fully implemented and tested, but needs way too much documentation for me to write right now, so I won't merge right away.

Thi is however, the first priority in compiler changes, we shouldn't do any non-insignificant changes before merging this because rebasing this is a lot of work.

developedby:
If anyone wants to go through this PR and write some documentation feel free

developedby:
Closes #679 and #615

PR #635, opened 2024-07-18
Add alternative and alternative_fn parser functions

imaqtkatt:
I'm not sure if this is the best way of doing it, what can be improved?

### DIFF
--- a/src/fun/parser.rs
+++ b/src/fun/parser.rs
@@ -352,6 +352,8 @@ impl<'a> TermParser<'a> {
   }
 
   fn parse_pattern(&mut self, simple: bool) -> ParseResult<Pattern> {
+    const PATTERNS: &[&str] =
+      &["Constructor", "Tuple", "Dup", "List", "String", "Char", "Number", "Unscoped Var", "Var"];
     maybe_grow(|| {
       let (tag, unexpected_tag) = self.parse_tag()?;
       self.skip_trivia();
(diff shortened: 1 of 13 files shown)


developedby:
I don't think this is a very good approach. Term already has too many alternatives for the error to be comprehensible.

I think simply wording the existing error better is more useful. "Expected a term. Found ')' <Context>".
And then combine this with better documentation that actually explains what is meant by "term", "top-level definition", "statement", etc.

Also fixing the `labelled` parser to actually be usable, like some basic level of recovering or saving some context on what it the high-level thing being parsed (to avoid errors like "Expected: Name" when we're expecting any term).

developedby:
Maybe we could give a few examples and show alternatives only when there are very few (3-4 max)

imaqtkatt:
I see. I still don't know what is the best way of "fixing" the `labelled` parser without changing a lot of code, I'm closing this PR since you are requesting a different thing

developedby:
I was thinking that maybe we could have a stack of things we're expecting in the state of the parser.

Once we're 100% sure that we're expecting something more specific then we can push a new thing to the stack. Then if an error occurs we just look at the top one.

Something like:
* Start expecting a top-level definition
* Parse `def`, expect a function definition.
* Parsing the function name, expect a top-level name, then pop.
* Parsing `(`, expect `(`, etc.

But importantly, when we have an alternative and we didn't match any of the "guard" clauses that identify a specific alternative, we still expect the construct one level above.
So for example, if the parser didn't find `def`, `type` or `object`, it'll try to parse a fun-syntax function definition but still give an error of `expected top-level definition`.

This would be a generalization of the `labelled` logic that allows for nested labels.

This is a bit different to how we did the rest of the parser, but would only require adding some `push_label` and `pop_label` calls without needing to modify the overall structure of the code.

Issue #554, opened 2024-06-06
Tuples should be defined by commas(like python), not by parentheses


TimotejFasiang:
**Is your feature request related to a problem? Please describe.**
In python, it is the commas, not the parentheses, that define the tuple. Not sure whether it's worth the effort to change this or whether there is a specific reason that it has to be this way.

**Describe the solution you'd like**
Having tuples be defined by commas and not by parentheses.

**Describe alternatives you've considered**
Leaving it the way it is, with tuples being defined by parentheses.

developedby:
There is no specific reason and I don't think it'll cause any syntax conflict.

I just forgot about this case in Python.

PR #637, opened 2024-07-19
#554 tuples should be defined by commas like python not by parentheses

imaqtkatt:
Tuples without parentheses are now freely accepted in assignments, returns and lambdas, in other cases we must use `(..., ...)`


developedby (changes requested):
Update the changelog.

Add example for assigning without parentheses, eg
```py
a, b = 1, 2
```

developedby (approved):

Issue #464, opened 2024-05-23
Improve error message when program is missing arm of branching statement


developedby:
for example
```
def is_even(n):
  if n % 2 == 0:
    return 1
  else:
    return 0

def get_even_sum_under_if(this_number):
  bend current = 0:
    when current < this_number:
      new_num = fork(current + is_even(current))
    
      return new_num

def main():
  return get_even_sum_under_if(1000)
```
will complain about wrong indentation and not the missing else

PR #645, opened 2024-07-24
#464 Improve error message when program is missing arm of branching statement

imaqtkatt:


src/lib.rs
```diff
@@ -14,6 +14,7 @@ use diagnostics::{Diagnostics, DiagnosticsConfig, ERR_INDENT_SIZE};
 use net::hvm_to_net::hvm_to_net;
 
 pub mod diagnostics;
+#[allow(clippy::mutable_key_type)]
```
developedby, line 17: Why is this needed here in the `mod` declaration?

src/lib.rs
```diff
@@ -197,6 +198,7 @@ pub fn readback_hvm_net(
   let mut diags = Diagnostics::default();
   let net = hvm_to_net(net);
   let mut term = net_to_term(&net, book, labels, linear, &mut diags);
+  #[allow(clippy::mutable_key_type)]
```
developedby, line 201: Did this start to flag after a clippy update?

imaqtkatt: I think so, with rust 1.80

src/lib.rs
imaqtkatt: To allow this warning inside the `fun` module, the only place we are using HashSet/HashMap

developedby: Shouldn't this be placed in wherever we use the hashmap with mutable key?

developedby: Oh is it because it's too many places? in every single HashMap<Name>

developedby, line 17: Add a comment explaining why this is necessary.
Eg. "`Name` is implemented with inner mutability so it triggers this warning, but we know it works."

developedby (changes requested):

src/lib.rs
imaqtkatt: Yes

developedby (approved):

PR #649, opened 2024-07-27
enable net_size_check by default with different max sizes for c and cuda

### LINKED ISSUE #622: Enable net size check when running with `run-cu`
**Is your feature request related to a problem? Please describe.**
We disabled the `check_net_size` compilation pass so that it wouldn't trigger when running things with the C runtime. But now, if a net ends up being too large for hvm-cu, it'll fail at runtime.

**Describe the solution you'd like**
We should enable `check_net_size` when the CLI command is `run-cu`.

We should also enable it when the command is `run-c` but with a different max net size (atm: 64 for hvm-cu, 4095 for hvm-c).
This max net size should probably be a cli option as well, but that's not necessary for now.

KunalSin9h:
Fixes #622


src/hvm/check_net_size.rs
```diff
@@ -2,16 +2,22 @@ use super::tree_children;
 use crate::{diagnostics::Diagnostics, fun::Name};
 use hvm::ast::{Book, Net, Tree};
 
-pub const MAX_NET_SIZE: usize = 64;
+pub const MAX_NET_SIZE_C: usize = 4095;
+pub const MAX_NET_SIZE_CUDA: usize = 64;
+
+pub fn check_net_sizes(book: &Book, diagnostics: &mut Diagnostics, cmd: &str) -> Result<(), Diagnostics> {
```
developedby, line 8: The CLI should be decoupled from Bend as a library, so we don't want to pass the command as a string.
Instead this value should be part of the compiler options struct, and set by the CLI in `compile_opts_from_cli`

developedby (changes requested):

KunalSin9h:
@developedby review it.

src/hvm/check_net_size.rs
```diff
@@ -1,17 +1,27 @@
...
 use hvm::ast::{Book, Net, Tree};
 
-pub const MAX_NET_SIZE: usize = 64;
+pub const MAX_NET_SIZE_C: usize = 4095;
+pub const MAX_NET_SIZE_CUDA: usize = 64;
+
+pub fn check_net_sizes(
+  book: &Book,
+  diagnostics: &mut Diagnostics,
+  opts: &CompileOpts,
```
developedby, line 11: This function shouldn't receive all the options, only the one it uses.

src/lib.rs
```diff
@@ -331,6 +331,9 @@ impl OptLevel {
 
 #[derive(Clone, Debug)]
 pub struct CompileOpts {
+  /// The Bend command
+  pub command: String,
```
developedby, line 335: Instead of storing the command, which is a CLI thing, we should store the target architecture.
For now we can use a structured format, like an enum
```
enum CompileTarget {
  C,
  Cuda,
  Unknown,
}
```

src/hvm/check_net_size.rs
```diff
@@ -1,17 +1,27 @@
...
-pub fn check_net_sizes(book: &Book, diagnostics: &mut Diagnostics) -> Result<(), Diagnostics> {
   diagnostics.start_pass();
 
   for (name, net) in &book.defs {
     let nodes = count_nodes(net);
-    if nodes > MAX_NET_SIZE {
+    if nodes > net_size_bound {
       diagnostics.add_rule_error(
-        format!("Definition is too large for hvm (size={nodes}, max size={MAX_NET_SIZE}). Please break it into smaller pieces."),
+        format!("Definition is too large for hvm (size={nodes}, max size={net_size_bound}). Please break it into smaller pieces."),
```
developedby, line 24: If you change the command to be a target instead, it would be nice to print it here. Something like `Definition is too large for HVM target {} ...`

developedby (changes requested):

src/lib.rs
```diff
@@ -406,13 +418,14 @@ impl Default for CompileOpts {
   /// Uses num-scott ADT encoding.
   fn default() -> Self {
     Self {
+      target_architecture: CompilerTarget::Cuda,
```
developedby, line 421:
```suggestion
      target_architecture: CompilerTarget::Unknown,
```

tests/golden_tests.rs
```diff
@@ -517,7 +517,7 @@ fn compile_long() {
       ..Default::default()
     };
 
-    compile_book(&mut book, opts.clone(), diagnostics_cfg, None)?;
-    Ok("Compiled".to_string())
+    let res = compile_book(&mut book, opts.clone(), diagnostics_cfg, None)?;
+    Ok(format!("{}{}", res.diagnostics, hvm_book_show_pretty(&res.hvm_book)))
```
developedby, line 521: We don't want to print the book in case of success. This test in specific is made to generate way too large structures.

Since the compile fails it should already be saving the error messages. I think it should work if  you just leave it like it was.

tests/snapshots/compile_long__huge_tree.bend.snap
```diff
@@ -2,4 +2,6 @@
 source: tests/golden_tests.rs
 input_file: tests/golden_tests/compile_long/huge_tree.bend
 ---
-Compiled
+[4m[1m[31mErrors:[0m
+[1mIn definition '[4mmain[0m[1m':[0m
+  Definition is too large for hvm (size=120002, max size=4095). Please break it into smaller pieces.
```
developedby, line 7: This is weird, it should be breaking this thing in pieces of the size we passed to the `float_combinators` pass (64, MAX_NET_SIZE_CUDA)

developedby: I'll have a look into why this is later

developedby (changes requested):

developedby (approved):
Looks good, good job.

KunalSin9h:
@developedby updated the snapshots!

developedby:
Thank you so much! This has been bugging me for some time

Issue #629, opened 2024-07-11
Monadic operations inside `match` nested inside `with` have unexpected behaviour


nicolas-abril:
### Reproducing the behavior

When writing a monadic operation inside of a pattern matching statement nested inside a `with` block, the value returned by the match, assigned to the variable is the bind operation.

Users will expect instead for the operations to be executed, then the variable assigned and only then the following monadic operations executed.

Example:
```py
def main():
  with IO:
    match []:
      case List/Cons:
        * <- IO/print("X")
        ok = 0
      case List/Nil:
        * <- IO/print("O")
        ok = 1
    return wrap(ok)
```

Here, the value of `ok` is ` λa (a IO/Call/tag IO/MAGIC "WRITE" (IO/FS/STDOUT, [79]) λ* 1)` when we expected it to be `1`.

This means that the `print` is never actually executed since it's stored inside of a `wrap`, which ends the IO action.

In the functional syntax, it's a bit more obvious that this happens:
```hs
main = with IO {
  let ok = match [] {
    case List/Cons:
      ask * = (IO/print "X")
      0
    case List/Nil:
      ask * = (IO/print "O")
      1
  }
  (wrap ok)
}
```

Syntactically, i don't know that is the best way of solving this.

In a previous version of Kind, we made the block inside `with` in a functional syntax be a separate AST that parses similar to the imperative one.
That's not a very elegant solution, but it worked.

With the imperative syntax the correct compilation is pretty obvious. We just need to make the final assignment of the variable of a match inside a with be a monadic bind instead of a normal variable bind.

We need to solve how to do this in a nice way for the functional syntax and how to handle the interaction of nested blocks (how to deal with nested `with`s, nested other things, how to handle nested lets with separate with blocks, how to handle recursion in these nested blocks, etc).

### System Settings

Bend: 0.2.36
HVM: 2.0.21

developedby:
We can also not solve this issue directly and instruct users on the way of writing that will lead to the correct result. Basically, when your block has a monadic operation then it must return/assign a value with that monad as the type.

In the example above they are for the both syntaxes:
```py
def main():
  with IO:
    match []:
      case List/Cons:
        * <- IO/print("X")
        ok = wrap(0)
      case List/Nil:
        * <- IO/print("O")
        ok = wrap(1)
    ok <- ok
    return wrap(ok)
```
and
```rs
main = with IO {
  ask ok = match [] {
    List/Cons:
      ask * = (IO/print "X")
      (wrap 0)
[...]

PR #652, opened 2024-08-01
#629 Monadic operations inside match nested inside with have unexpected behaviour

imaqtkatt:


developedby (changes requested):
Can you add a test with mismatched final statements? A return in one branch and an ask in the other, and a assignment in one branch and an ask in the other

developedby (changes requested):
Update the changelog

developedby (approved):
nice, well done

Issue #648, opened 2024-07-26
Hexadecimal and binary floating-point numbers


edusporto:
### Reproducing the behavior

Bend currently allows numbers to be defined using hexadecimal base by pre-pending `0x` to them, like `0xA1`. It also allows these numbers to be floating-point, like `0xA1.0`. However, every digit after the dot is not treated as hexadecimal, but as decimal, making it so that a number like `0xA.1` is equal to `10.1`.

Different languages deal with this differently:
- Python does not allow hexadecimal literals to be floating-point.
- Haskell allows it, but digits after the dot are treated as hexadecimal, having each place after the dot be subdivided in 16 to the power of the digit depth. Examples:
  - `0x0.1 == 1.0 / 16.0^1.0 == 0.0625`
  - `0x0.05 == 5.0 / 16.0^2.0 == 1.953125e-2`

There are a few ways we could deal with this.
1. Keep the current behavior.
  a. I believe it could be confusing.
2. Keep the current behavior but allow `0x` after the dot, like `0x0.0x1`.
  b. I think this is also confusing.
  c. This already compiles in Bend, but with weird behavior: `0xA.0xA == 10.010`.
3. Forbid it like Python.
  d. Considering Bend only allows binary operations to happen between values of the same number type, I think this wouldn't be good, since it would make it more difficult to use hexadecimal numbers on operations with floating-point numbers.
4. Implement it like Haskell.
  e. I believe this to be the best way, since it keeps consistency.

Edit: Edited the issue to include binary floating-point numbers. They're in pretty much the same situation as hexadecimal numbers, except that they don't work in Haskell, where `0b1.1` is treated as `(.) 0b1 1`.

### System Settings

Bend 0.2.36

PR #654, opened 2024-08-02
#648 Hexadecimal and binary floating-point numbers

imaqtkatt:


developedby (changes requested):

src/fun/parser.rs
```diff
@@ -1518,80 +1518,87 @@ pub trait ParserCommons<'a>: Parser<'a> {
...
       self.advance_one();
       let ini_idx = *self.index();
-      let fra = self.parse_u32()? as f32;
-      let end_idx = *self.index();
-      let fra = fra / 10f32.powi((end_idx - ini_idx) as i32);
+      let fra_str = self.take_while(|c| c.is_digit(radix as u32) || c == '_');
+      let fra_str = fra_str.chars().filter(|c| *c != '_').collect::<String>();
+      let end_idx = ini_idx + fra_str.len();
+      let fra = u32::from_str_radix(&fra_str, radix as u32).map_err(|e| e.to_string())?;
+      let fra = fra as f32 / (radix.to_f32()).powi((end_idx - ini_idx) as i32);
```
developedby, line 1579: You could just use `fra_str.len()` directly here
```rs
      self.advance_one();
      let fra_str = self.take_while(|c| c.is_digit(radix as u32) || c == '_');
      let fra = u32::from_str_radix(&fra_str, radix as u32).map_err(|e| e.to_string())?;
      let fra = fra as f32 / (radix.to_f32()).powi(fra_str.len() as i32);
      Some(fra)
```

developedby (changes requested):

developedby (changes requested):
Update the syntax document and the changelog

developedby (approved):
good job

PR #656, opened 2024-08-04
Fix fun syntax with/wrap example

Janiczek:


docs/syntax.md
```diff
@@ -1093,7 +1093,7 @@ Result/wrap x = (Result/Ok x)
 with Result {
   ask x = (some_operation ...)
   ask y = (some_operation ...)
-  wrap(x * y)
+  (wrap (* x y)
```
edusporto, line 1096: It's missing a second parentheses after the last one
```
(wrap (* x y))
```

Janiczek: Dang, I could swear I double-checked the change 😅  Thanks!

developedby (approved):

developedby:
Oh no, there's some bug in the test...

Issue #657, opened 2024-08-05
Make IO functions return Result


developedby:
https://github.com/HigherOrderCO/HVM/pull/407 Implements the necessary changes for us to capture both errors in the runtime (like incorrect arguments, non-existant IO called) as well as errors in the execution of the IO themselves.

The IOs now return `Result<A, IOError<B>>`

We need to add the IOError type to the prelude
```
# Not necessarily these names
type IOError(T):
  TypeError
  UnknownFn
  Inner { val: T }
```
First one is generated by the IO function, for when user passes incorrect / invalid arguments.
Second one is generated by the runtime itself, when the user calls a function with a name that isn't registered.
Third one is all the errors internal to the IO functions themselves. For example, opening a file when it doesn't exist returns an Inner with a string containing the OS message.

Looking at the pull request shows the other inner errors returned by each IO function.

We need to change the existing IO functions to reflect this change in HVM.

PR #661, opened 2024-08-07
#657 Make IO functions return Result

imaqtkatt:


developedby (changes requested):
Just a minor documentation change since `unwrap` is not type safe

developedby (approved):

PR #662, opened 2024-08-07
Adds builtin functions Tree/reverse and Tree/to_list

In-Veritas:


src/fun/builtins.bend
```diff
@@ -96,7 +106,22 @@ type Tree:
...
+      list = DiffList/concat(tree.left, tree.right)
+  return DiffList/to_list(list)
+
+# Reverses a tree swapping right and left leaves.
+def Tree/reverse(tree):
+  fold tree:
+    case Tree/Leaf:
+      return !tree.value
+    case Tree/Node:
+      return ![tree.right, tree.left]
```
developedby, line 124:
```suggestion
      return ![tree.right, tree.left]

```

developedby (changes requested):

developedby (approved):

Issue #666, opened 2024-08-08
Whenever a test in /examples doesn't compile, the test fails instead of updating the snapshot


In-Veritas:
### Reproducing the behavior

Example:
When running tests on the examples folder, if one doesn't compile, I get a "PoisonError", and following unrelated tests will also fail:

thread 'examples' panicked at tests/golden_tests.rs:476:51:
called `Result::unwrap()` on an `Err` value: "In /home/gex/Bend/examples/list.bend :\n\u{1b}[1m- expected:\u{1b}[0m term\n\u{1b}[1m- detected:\u{1b}[0m\n\u{1b}[0m  135 |   \u{1b}[4m\u{1b}[31m}\u{1b}[0m"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- import_system stdout ----
Testing /home/gex/Bend/tests/golden_tests/import_system/import_main.bend
/home/gex/Bend/tests/golden_tests/import_system/import_main.bend
thread 'import_system' panicked at tests/golden_tests.rs:186:37:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }

---- linear_readback stdout ----
Testing /home/gex/Bend/tests/golden_tests/linear_readback/church_mul.bend
/home/gex/Bend/tests/golden_tests/linear_readback/church_mul.bend
thread 'linear_readback' panicked at tests/golden_tests.rs:137:35:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }

---- run_file stdout ----
Testing /home/gex/Bend/tests/golden_tests/run_file/360_no_scope.bend
/home/gex/Bend/tests/golden_tests/run_file/360_no_scope.bend
thread 'run_file' panicked at tests/golden_tests.rs:160:37:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }

---- cli stdout ----
Testing /home/gex/Bend/tests/golden_tests/cli/compile_all.bend
/home/gex/Bend/tests/golden_tests/cli/compile_all.bend
Testing /home/gex/Bend/tests/golden_tests/cli/compile_inline.bend
/home/gex/Bend/tests/golden_tests/cli/compile_inline.bend
thread 'cli' panicked at tests/golden_tests.rs:400:35:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }

### System Settings

.

developedby:
[...]

PR #671, opened 2024-08-12
#666 Whenever a test in /examples doesn't compile, the test fails instead of updating the snapshot

imaqtkatt:


imaqtkatt:
I believe this already works for this case, but we need to take care of unwrapping things and getting this poison error

Issue #650, opened 2024-07-30
Report range on diagnostics


edusporto:
Currently, the diagnostic struct is defined as such:
```rust
pub struct Diagnostic {
  message: String,
  severity: Severity,
}
```
Some diagnostics can be related to code on a specific range of a file, like parsing errors for example. Currently, the location and file in which such errors occur is encoded within the `message` string, but that makes it difficult to get the source location of these diagnostics after they've been created.

This is especially useful to the language server for Bend we are currently developing.

VictorTaelin:
Bend is being reorganized. This repository (bendlang/bend) now belongs to Bend 2, a new implementation written from scratch, coming soon. Bend 1 lives on with its full history at https://github.com/HigherOrderCO/Bend1. We are closing every Bend 1 issue and pull request here: if this one still applies to Bend 1, please re-open it there.

PR #673, opened 2024-08-14
Changes towards a language server friendly compiler

edusporto:
This PR:
- Changes all parsing to return `TSPL::ParseError`, new in the latest TSPL version, as to report error range information,
- Adds optional `FileSpan` to each `Diagnostic`,
- Adds the `Function` diagnostic origin,
- Plenty of smaller changes to make the previous points work.

I'm still not sure about the `TextSpan` construct - the new TSPL version only reports byte range information, while the LSP requires line and column range information, so that's how I implemented that struct. I think it could be better to change `TextSpan` to only report byte range information, and provide a function to convert between the two given the source code. I could also change TSPL again to include both types of range information, but I think it's not something the team would want since many changes to it would stop it from being the _simplest_ parser library.

I'll wait until after the review to add a changelog because of the aforementioned issues.

Related to #650.


src/diagnostics.rs
```diff
@@ -28,16 +31,21 @@ pub struct DiagnosticsConfig {
...
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub enum DiagnosticOrigin {
+  /// An error when parsing source code.
+  Parsing,
   /// An error from the relationship between multiple top-level definitions.
   Book,
   /// An error in a pattern-matching function definition rule.
   Rule(Name),
+  /// An error when transforming `imp` syntax into `fun`.
+  Function(Name),
```
developedby, line 48: `Function` and `Rule` diagnostics are the same, you could have just a single one called `Function`.

I don't think its very meaningful which syntax it comes from.

src/imp/to_fun.rs
```diff
@@ -83,7 +99,7 @@ fn wrap(nxt: Option<fun::Pattern>, term: fun::Term, ask: bool) -> StmtToFun {
 }
 
 impl Stmt {
-  fn into_fun(self) -> Result<StmtToFun, String> {
+  fn into_fun(self) -> Result<StmtToFun, Diagnostics> {
```
developedby, line 102: What is the point of returning a new diagnotics context if we only ever return a single string?

tests/snapshots/parse_file__repeated_datatype_name.bend.snap
```diff
@@ -4,4 +4,8 @@ input_file: tests/golden_tests/parse_file/repeated_datatype_name.bend
 ---
 [4m[1m[31mErrors:[0m
 In tests/golden_tests/parse_file/repeated_datatype_name.bend :
-Found a repeated field 'Expr' in constructor Expr/Plus.
+[1m- information:[0m Found a repeated field 'Expr' in constructor Expr/Plus.
+[1m- expected:[0m field
```
developedby, line 8: It'd be better to not have the "expected" "detected" part here, since it is a field, just not a good one.

Either that or change the message to `unique field` or similar.

Cargo.toml
```diff
@@ -24,10 +24,10 @@ default = ["cli"]
 cli = ["dep:clap"]
 
 [dependencies]
-TSPL = "0.0.12"
+TSPL = "0.0.13"
 clap = { version = "4.4.1", features = ["derive"], optional = true }
 highlight_error = "0.1.1"
-hvm = "=2.0.19"
+hvm = "2.0.22"
```
developedby, line 30:
```suggestion
hvm = "=2.0.22"
```
Since HVM doesn't follow semnatic versioning

cspell.json
```diff
@@ -21,6 +21,7 @@
     "concat",
     "ctrs",
     "cuda",
+    "curr",
```
developedby, line 24: We were using `cur` in other parts of the code.

developedby (changes requested):

src/imp/to_fun.rs
edusporto: I changed this one because `into_fun` calls Definition's `to_fun` in this section:
```rs
Stmt::LocalDef { def, nxt } => {
  let (ask, nxt_pat, nxt) = take(*nxt)?;
  let def = def.to_fun()?; // <----- here
  let term = fun::Term::Def { def, nxt: Box::new(nxt) };
  wrap(nxt_pat, term, ask)
}
```
and I had changed that function's signature to return `Diagnostics` so it could report its origin. To keep returning strings in this function we could create a function to extract error message from `Diagnostics` with a single `Diagnostic`.
What do you think?

cspell.json
edusporto: Changed to `curr` to `cur` in the functions I wrote

tests/snapshots/parse_file__repeated_datatype_name.bend.snap
edusporto: Changed it to show `information` and `location` without `expected`, what do you think?

src/imp/to_fun.rs
developedby: But if we're returning only strings, shouldn't the `From<String>` implementation of `Diagnostics` take care of converting it at the place you point out? (`?` calls something like `into_residual` which calls a `into`)

In the same way that the `?` inside `into_fun` is converting a string into a diagnostic, wouldn't it already do the same here? I don't understand why changing the signature is necessary.

tests/snapshots/parse_file__repeated_datatype_name.bend.snap
developedby: Looks good

src/imp/to_fun.rs
edusporto: You're right; originally I'd changed `into_fun` to return Diagnostics because of a problem I was having with multiple `Error:` messages being printed into the error string and with what you said I noticed I could do it in another way; I changed it back into returning strings.

src/fun/mod.rs
```diff
@@ -78,7 +74,7 @@ pub enum Source {
   /// Was generated by the compiler.
   Generated,
   /// Source code location from the current book.
-  Local(Range<usize>),
+  Local(TextSpan),
```
developedby, line 77: I wonder if we could and should unify all these source points and spans into a single thing. It's a bit overcomplicated atm.

developedby: Maybe something like
```
struct Source {
  file: FileID/FileName,
  span: Range<usize>,
  kind: SourceKind,
}

enum SourceKind {
  User,
  Generated,
  Builtin
}
```
And then correctly propagate this information throughout the functions and terms, etc.

developedby (approved):

src/fun/mod.rs
edusporto: Done in the latest commit

edusporto:
The latest commit includes the last suggested change - all diagnostic range and file information has been unified into the `Source` struct. This allowed me to change the way errors are printed to almost always include their source files, leading to MANY test result changes. If you'd like me to separate this into multiple pull requests so it's easier to review, let me know.

src/fun/mod.rs
```diff
@@ -1026,7 +1032,8 @@ impl Definition {
   }
 
   pub fn new_gen(name: Name, rules: Vec<Rule>, builtin: bool) -> Self {
```
developedby, line 1034: I'm not exactly sure how this function is used, but generated functions should inherit the source of their generator.

So for example if we have
```
# File Nat/to_u24.bend
from Nat import {Succ, Zero}

def to_u24(x):
  match x:
    case Succ:
      return 1 + to_u24(x.pred)
    case Zero:
      return 0
```

This will generate function `Nat/to_u24__C0 = @x.pred (+ 1 (to_u24 x.pred))`.

The source of this generated function should be `Nat/to_u24.bend`, pointing at the `Succ` arm.

We don't need to carry the span to the generated function in this commit, that's a very large change for the future, but we could point it to the correct file at least.

I'm guessing that `new_gen` is used in this way.

tests/snapshots/import_system__import_main.bend.snap
```diff
@@ -3,4 +3,5 @@ source: tests/golden_tests.rs
 input_file: tests/golden_tests/import_system/import_main.bend
 ---
 [4m[1m[31mErrors:[0m
+[1mDuring execution :[0m
 Can not import the entry point of the program.
```
developedby, line 7: Can we not report the file that originates this error? It looks strange to know where the error comes from.

tests/snapshots/import_system__imports_alias_shadow.bend.snap
```diff
@@ -3,4 +3,5 @@ source: tests/golden_tests.rs
 input_file: tests/golden_tests/import_system/imports_alias_shadow.bend
 ---
 [4m[1m[31mErrors:[0m
+[1mDuring execution :[0m
 The import 'lib/nums/two' shadows the imported name 'lib/nums/one'
```
developedby, line 7: Same thing for these other import errors.

tests/snapshots/readback_hvm__bad_net.bend.snap
```diff
@@ -3,6 +3,7 @@ source: tests/golden_tests.rs
 input_file: tests/golden_tests/readback_hvm/bad_net.bend
 ---
 [4m[1m[33mWarnings:[0m
+[1mDuring execution :[0m
```
developedby, line 6: Looking at all the errors as a whole, I actually think the `During execution` is not actually adding anything. In all cases we have here we could just remove this line and the comprehension of the message wouldn't be affected.

Also, for most things it should actually be a "during compilation", right?

developedby (changes requested):

src/fun/mod.rs
edusporto: Fixed in the latest commit

tests/snapshots/import_system__imports_alias_shadow.bend.snap
edusporto: I think I'll change this so no message appears when there are only diagnostics with `source.file == None`, and when there are there are diagnostics with source files and ones without, the ones without will show in the end with the message `Other diagnostics:` or something similar. What do you think?

tests/snapshots/import_system__import_main.bend.snap
edusporto: I think I'll change this so no message appears when there are only diagnostics with source.file == None, and when there are there are diagnostics with source files and ones without, the ones without will show in the end with the message Other diagnostics: or something similar. What do you think?

edusporto:
The new diagnostics printing should be much better.

tests/snapshots/import_system__import_main.bend.snap
developedby: Sure

src/fun/parser.rs
```diff
@@ -1093,6 +1127,15 @@ impl<'a> Parser<'a> for TermParser<'a> {
     self.expected_spanned(exp, ini_idx..end_idx)
   }
 
+  /// Generates an error message with an additional custom message.
+  ///
+  /// Override to have our own error message.
```
developedby, line 1132: This one osnt an override

developedby (changes requested):
Update the changelog

src/fun/parser.rs
edusporto: It's in the newer version of the TSPL crate

src/diagnostics.rs
```diff
@@ -159,47 +192,110 @@ impl Diagnostics {
...
-            DiagnosticOrigin::Book => {
-              for err in errs {
-                writeln!(f, "{err}")?;
+      // Produce the structure described above.
+      let groups: BTreeMap<&Option<String>, BTreeMap<&DiagnosticOrigin, Vec<&Diagnostic>>> = diagnostics
+        .fold(BTreeMap::new(), |mut file_tree, (origin, diags)| {
+          for diag in diags {
+            // We need to allow this Clippy warning due to `Name` in `DiagnosticOrigin::Function`.
+            // We know how it works, so it shouldn't be a problem.
+            #[allow(clippy::mutable_key_type)]
```
developedby, line 228: We can allow it at the crate level, since we're just sprinkling this macro in different places, always for the same reason.

developedby (approved):

PR #700, opened 2024-08-29
Fix IO/FS/read_line when the line ends with a EOF

LunaAmora:


src/fun/builtins.bend
```diff
@@ -316,8 +316,13 @@ def IO/FS/read_line.read_chunks(fd, chunks):
         return wrap(bytes)
       case Result/Err:
```
developedby, line 317: Can you add a comment explaining the meaning of each branch? It's getting hard to read
(ok means found new line, backtrack and join chunks, err means no new line, length == 0 means file over, so return as if new line, else means the line is still ongoing

developedby (changes requested):
Update the changelog

src/fun/builtins.bend
```diff
@@ -307,17 +307,26 @@ def IO/FS/read_line.read_chunks(fd, chunks):
     # Read line in 1kB chunks
     chunk <- IO/done_on_err(IO/FS/read(fd, 1024))
     match res = List/split_once(chunk, '\n'):
+      # Found a newline, backtract and join chunks
```
developedby, line 310:
```suggestion
      # Found a newline, backtrack and join chunks
```

developedby (changes requested):

developedby (approved):

kings177 (approved):

Issue #713, opened 2024-09-11
Refactor Transform String Resugar Term Scott 


Liberxue:
Hey guys 

https://github.com/HigherOrderCO/Bend/blob/d91742e73b108ba1a3423e27114961813416de1c/src/fun/transform/resugar_string.rs#L16C1-L240C2

I've been reviewing our string encoding conversion code, specifically the`build_string_num_scott ` and `build_string_scott ` functions. I noticed these functions share quite a bit of similar logic,

I'm curious about a few things:

Was keeping these functions separate a deliberate design choice? If so, could you share the reasoning behind it🤔?

I appreciate any insights you can provide. If you think refactoring could be beneficial, I'd be happy to recode

PR #714, opened 2024-09-11
Refactor Transform String Resugar Term Scott 

Liberxue:
#713 
* Create  struct to wrap user-provided cons handling functions, implementing the  trait.
* Implement  struct to handle default cons patterns, also implementing the  trait.
* Replace  with  type, better expressing success or failure semantics.
* Utilize Rust's pattern matching and  method to simplify error handling process.

### DIFF
--- a/src/fun/transform/resugar_list.rs
+++ b/src/fun/transform/resugar_list.rs
@@ -266,4 +266,3 @@ fn build_list_scott(term: &mut Term, mut l: Vec<Box<Term>>) -> Result<Vec<Box<Te
     Err(l)
   })
 }
-
--- a/src/fun/transform/resugar_list.rs
+++ b/src/fun/transform/resugar_list.rs
@@ -79,7 +79,7 @@ impl Term {
   /// Converts scott-encoded lists ending with List/Nil to list literals.
   fn resugar_lists_scott(&mut self) {
     maybe_grow(|| {
-      // Search for a List/Cons pattern in the term and try to build a list from that point on.
+      // Searc                  h for a List/Cons pattern in the term and try to build a list from that point on.
       // If successful, replace the term with the list.
       // If not, keep as-is.
 
@@ -124,7 +124,7 @@ impl Term {
               match l {
                 Ok(l) => *self = Term::List { els: l.into_iter().map(|x| *x).collect() },
                 // Was not a list term, keep as-is.
-                Err(mut l) => {
+                                  Err(mut l) => {
                   *head = l.pop().unwrap();
                   assert!(l.is_empty())
                 }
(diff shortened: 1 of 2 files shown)


developedby (changes requested):
The visitor is completely unnecessary here, doesn't really simplify.

I like the idea of the `try_resugar` functions, it lets the pattern matching not be repeated twice.

You could have a single `build_string` function that takes takes the encoding, tries the correct encoded pattern and tries the non-applied constructors, without repetition.

Also, please leave the `if` tree since it mimics the shape of the AST and is easier for me to visualize the structure.

Liberxue:
> The visitor is completely unnecessary here, doesn't really simplify.
> 
> I like the idea of the `try_resugar` functions, it lets the pattern matching not be repeated twice.
> 
> You could have a single `build_string` function that takes takes the encoding, tries the correct encoded pattern and tries the non-applied constructors, without repetition.
> 
> Also, please leave the `if` tree since it mimics the shape of the AST and is easier for me to visualize the structure.

Done Please review～～thx @developedby

src/fun/transform/resugar_string.rs
```diff
@@ -7,101 +7,132 @@ impl Term {
...
+
+        // Try custom cons handler
+        if let Some((head, tail)) = cons_handler(current) {
+          s.push(head);
+          current = tail;
+          continue;
+        }
+
+        // Cons: @x (x CONS_TAG <num> <str>)
+        if let Term::Lam { tag: Tag::Static, pat, bod } = current {
```
developedby, line 123: This is specifically the "num-scott" encoding, you shouldn't apply it for all cases

developedby (changes requested):
This is nicer to read without the visitor, but my other previous comments still apply.

Also, please do the changes on `resugar_list` as well. It's almost an exact copy of this one, just looking for a slightly different pattern. It would be confusing if they were implemented differently

Liberxue:
> This is nicer to read without the visitor, but my other previous comments still apply.
> 
> Also, please do the changes on `resugar_list` as well. It's almost an exact copy of this one, just looking for a slightly different pattern. It would be confusing if they were implemented differently

In the `resugar_lists` file, after careful reading, I found that extracting code blocks could lead to reduced performance due to cloning. I haven't thought of a better approach at the moment~~~

developedby (approved):

src/fun/transform/resugar_list.rs
```diff
@@ -79,7 +79,7 @@ impl Term {
   /// Converts scott-encoded lists ending with List/Nil to list literals.
   fn resugar_lists_scott(&mut self) {
     maybe_grow(|| {
-      // Search for a List/Cons pattern in the term and try to build a list from that point on.
+      // Searc      h for a List/Cons pattern in the term and try to build a list from that point on.
```
developedby, line 82:
```suggestion
      // Search for a List/Cons pattern in the term and try to build a list from that point on.
```

developedby:
Can you just run `cargo fmt`?

Liberxue:
> Can you just run `cargo fmt`?
So Sorry OKK~~ done

developedby:
Thank you very much for the contribution

Issue #724, opened 2024-10-07
Constructor patterns in pattern matching functions are not renamed when imported


nicolas-abril:
### Reproducing the behavior

From @NoamDev at https://github.com/HigherOrderCO/Bend/issues/722#issuecomment-2396005741

** a.bend **

type MyType:
  A {data: u24}
  B {data: u24}

Foo: MyType -> MyType

(Foo (MyType/A a)) = (MyType/A a)
(Foo (MyType/B b)) = (MyType/B b)

** main.bend**

from a import *
main = "hello world"

> bend check main.bend
Errors:
In a :
In definition 'a/Foo':
  Unbound constructor 'MyType/A' in pattern matching rule.
  Unbound constructor 'MyType/B' in pattern matching rule.

### System Settings

Bend commit  [53d5262](https://github.com/HigherOrderCO/Bend/commit/53d52626eac4c4591c4346af3e85e1bca718d39b)

VictorTaelin:
Bend is being reorganized. This repository (bendlang/bend) now belongs to Bend 2, a new implementation written from scratch, coming soon. Bend 1 lives on with its full history at https://github.com/HigherOrderCO/Bend1. We are closing every Bend 1 issue and pull request here: if this one still applies to Bend 1, please re-open it there.

PR #725, opened 2024-10-07
Fix Ctr patterns not being renamed in imports

developedby:
Solves #724


NoamDev:
This doesn't seem to have fixed it

developedby:
> This doesn't seem to have fixed it

Do you have an example of where it fails? It's working for your previous example.

NoamDev:
> Do you have an example of where it fails? It's working for your previous example.
oh, you're right, well the new case is this:

**a.bend**
```
from b import *

type MyType:
  A {data: u24}
  B {data: u24}
```

**b.bend**
```
from a import *

Foo: MyType -> MyType
(Foo (MyType/A a)) = (MyType/A a)
(Foo (MyType/B b)) = (MyType/B b)
```

**main.bend**
```
from a import *
main = "hello world"
```

```
> bend run main.bend
Errors:
In b :
In definition 'b/Foo':
  Unbound variable 'a/MyType'.
  Unbound variable 'a/MyType'.
```

NoamDev:
@developedby in case you missed this

developedby:
> @developedby in case you missed this

I saw it, it's just that this one is a bit more complicated and doesn't seem to be a single-line fix. Sorry i forgot to uptade you here

PR #726, opened 2024-10-08
Fix `String/decode_utf8` returning a non-empty string when given an empty list

### LINKED ISSUE #701: String/decode_utf8 returns "\0" when decoding a empty List
### Reproducing the behavior

Strings in bend are not terminated with the null character, but the string decode function returns one when passing an empty list.

```py
def main:
  return String/decode_utf8([])
```
When run, this returns:
```
Result: "\0"
```

### System Settings

Bend commit 2446f3c23b6473c687356bbd7624dbe695db8fcd

### Additional context

_No response_

chalcolith:
Previously calling `(String/decode_utf8 [])` would return `"\0"`.  This PR makes it return `""`, adds this case to `utf8.bend` and updates the snapshot.

Fixes #701


developedby (approved):

developedby:
Thank you for the fix!

PR #747, opened 2025-01-15
Update, correct, and expand documentation

In-Veritas:
### Summary of Changes

This pull request includes the following updates and improvements:

#### **Documentation Updates**
- **Guide.md and Features.md**:
  - Updated and clarified information to ensure accuracy and better readability.
  - Expanded sections that were previously incomplete or unclear.
  - Added new examples to better illustrate key concepts.

- **Docs Folder**:
  - Reviewed and revised most files in the `docs` folder for consistency and clarity.
  - Addressed gaps in explanations and improved structure for easier navigation.

#### **Code Examples**
- Corrected example code snippets that previously failed to execute correctly.
- Verified and tested all examples to ensure they work as intended.

#### **Inline Comments**
- Added inline comments when describing built-in functions.
  - These comments provide detailed explanations of the purpose and usage of each function.

#### **Type Declarations**
- Declared types for every function in the documentation.

---

### Additional Notes
These changes aim to make the documentation more comprehensive, easier to understand, and practical for both new and experienced users. Feedback and suggestions for further improvements are welcome


docs/defining-data-types.md
```diff
@@ -3,48 +3,64 @@
...
+    return 0
 ```
 
 Rules can also have patterns.
 They work like match expressions with explicit bindings:
 
 ```py
-(Option.map (Some value) f) = (Some (f value))
-(Option.map None f) = None
+def Option/map(opt: Option(T), f: T -> U) -> Option(U):
```
developedby, line 40: The point of this section is showing pattern matching functions, you can't just change to the other syntax

docs/defining-data-types.md
```diff
@@ -3,48 +3,64 @@
...
 
 ```py
-type Boolean = True | False
+type Boolean:
+  True
+  False
 
-(Option.is_both_some (Some lft_val) (Some rgt_val)) = True
-(Option.is_both_some lft rgt) = False
+def Option/is_both_some(lft: Option(T), rgt: Option(T)) -> Boolean:
```
developedby, line 55: Same thing here

docs/ffi.md
```diff
@@ -13,7 +13,7 @@ def main():
     # The second argument is '0' if we want to load all functions immediately.
     # Otherwise it should be '1' when we want to load functions as we use them.
     # 'dl' is the unique id of the dynamic library.
-    dl <- IO/DyLib/open("./libbend_dirs.so", 0)
+    dl <- IO/DyLib/open("./libbend_dirs.so", 0) #IO(Result)
```
developedby, line 16: remove the comments

docs/ffi.md
```diff
@@ -27,8 +27,8 @@ def main():
 
     # In our example, 'ls' receives a path as a String and
     # returns a String with the result of the 'ls' command.
-    files_bytes <- IO/DyLib/call(dl, "ls", "./")
-    files_str = String/decode_utf8(files_bytes)
+    files_bytes <- IO/DyLib/call(Result/unwrap(dl), "ls", "./") #IO(Result)
```
developedby, line 30: instead of unwrapping  at every use, just save the unwrapped dl variable.

docs/imports.md
```diff
@@ -11,6 +11,8 @@ from path import name
 # or
 import path/name
 ```
+## Placement
+Imports should be placed at the top of a file.
```
developedby, line 15: This is not true (although its the recommended style)

developedby: Also it's written really unnaturaly and kind of disconnected from the rest of this document.

docs/lazy-definitions.md
```diff
@@ -2,7 +2,7 @@
 
 In strict-mode, some types of recursive terms will unroll indefinitely.
 
-This is a simple piece of code that works on many other functional programming languages, including hvm's lazy-mode, but hangs on strict-mode:
+This is a simple piece of code that works on many other functional programming languages and hangs on strict-mode:
```
developedby, line 5:
```suggestion
This is a simple piece of code that works on many other functional programming languages but hangs on Bend due to the strict evaluation of HVM2.
```

docs/lazy-definitions.md
```diff
@@ -16,20 +16,7 @@ Map = λf λlist
...
----------------------------------------
-(Map λa (+ a 1) λb λ* (b 1 Nil))
----------------------------------------
-(Cons (λa (+ a 1) 1) (Map λa (+ a 1) Nil))
----------------------------------------
-(Cons (λa (+ a 1) 1) (Nil λb λc (Cons (λa (+ a 1) b) (Map λa (+ a 1) c)) Nil))
----------------------------------------
-...
-```
+The recursive `Map`  creates an infinite reduction sequence because each recursive call expands into another call to Map, never reaching a base case. Which means that functionally, it never gets reduced.
```
developedby, line 19: It does get reduced, it never reaches normal form / finishes reducing.

docs/lazy-definitions.md
```diff
@@ -53,7 +40,7 @@ Map = λf λlist
 
 This code will work as expected, since `cons` and `nil` are lambdas without free variables, they will be automatically floated to new definitions if the [float-combinators](compiler-options.md#float-combinators) option is active, allowing them to be unrolled lazily by hvm.
 
-It's recommended to use a [supercombinator](https://en.wikipedia.org/wiki/Supercombinator) formulation to make terms be unrolled lazily, preventing infinite expansion in recursive function bodies.
+To handle recursion properly, recursive calls should be wrapped in top-level combinators. This ensures lazy unrolling of recursive terms, preventing infinite expansion during reduction. While [supercombinators](https://en.wikipedia.org/wiki/Supercombinator) are commonly used for this purpose, other combinator patterns can work as well, as long as they're lifted to the top level.
```
developedby, line 43: `recursive calls should be wrapped in top-level combinators`
I don't understand what you meant by this.

The exact way that this works is a bit hard to put in words. The recursive part of the function should be part of a combinator that is not in an active position. That way it can be lifted into a top-level function which is compiled into a lazy reference thus preventing the infinite expansion.

docs/native-numbers.md
```diff
@@ -50,13 +50,22 @@ minus_zero = -0.0
 ### Mixing number types
 
 The three number types are fundamentally different.
-If you mix two numbers of different types, HVM will interpret the binary representation of one of them incorrectly, leading to incorrect results. Which number is interpreted incorrectly depends on the situation and shouldn't be relied on for now.
+If you mix two numbers of different types, HVM will interpret the binary representation of one of them incorrectly, leading to incorrect results. Which number is interpreted incorrectly depends on the situation and shouldn't be relied on for now. Instead, you should make sure that all numbers are of the same type. We recently introduced a way to convert between the different number types, and using it is very easy, here's an example:
```
developedby, line 53: Don't say "we recently introduced" since this document will someday get old. Also separate that into a different section/paragraph

docs/pattern-matching.md
```diff
@@ -88,28 +88,37 @@ To ensure that recursive pattern matching functions don't loop in strict mode, i
...
-(Foo$C4) = λg (g Foo$C3 B)           # Foo.case_0.case_false
-(Foo$C3) = λh λi (i Foo$C1 Foo$C2 h) # Foo.case_0.case_false.case_cons
-(Foo$C1) = λj λk λl (A l j k)        # Foo.case_0.case_false.case_cons.case_cons
-(Foo$C2) = λ* B                      # Foo.case_0.case_false.case_cons.case_nil
-
-(Foo$C8) = λn λo λ* (o Foo$C6 Foo$C7 n) # Foo.case_+
-(Foo$C6) = λ* 0                         # Foo.case_+.case_true
-(Foo$C7) = λr (+ r 1)                   # Foo.case_+.case_false
+
+unchecked Foo__C0: _
```
developedby, line 92: Remove the type annotations here since they're only noise.

We should probably also add a note here that as a user you can't write a function with `__` in its name, that sequence is reserved for things generated by the compiler.

docs/pattern-matching.md
```diff
@@ -88,28 +88,37 @@ To ensure that recursive pattern matching functions don't loop in strict mode, i
...
+(Foo__C0) = λ* λa λb λc (A c a b)        # Foo.case_0.case_false.case_cons.case_cons
```
developedby, line 93: Please keep the grouping of the functions since it had a meaning. Also please align the comments.

docs/pattern-matching.md
```diff
@@ -88,28 +88,37 @@ To ensure that recursive pattern matching functions don't loop in strict mode, i
...
+unchecked Foo__C7: _
+(Foo__C7) = λa switch a { 0: λ* 0; _: Foo__C6; }  # Part of bool pattern matching
+unchecked Foo__C8: _
+(Foo__C8) = λa λ* (a Foo__C5)            # Part of main pattern matching
+unchecked Foo__C9: _
+(Foo__C9) = λa λb λ* (b Foo__C7 a)       # Foo.case_+
 ```
 
-Pattern matching equations also support matching on non-consecutive numbers:
+Pattern matching equations support non-consecutive numeric or character values. For example, this parser matches specific characters with corresponding tokens, with a catch-all case for any other character:
```
developedby, line 114: The second sentence, is not really relevant to what we're trying to show here. Also you're just describing what a tokenizer does.

docs/pattern-matching.md
```diff
@@ -123,18 +132,19 @@ Parse = λarg0 switch matched = (- arg0 '(') {
   }
 }
 ```
-Unlike with `switch`, with pattern matching equations you can't access the value of the predecessor of the matched value directly, but instead you can match on a variable.
+A key difference from direct switch expressions is how variable binding works. In pattern matching equations, you can't directly access predecessor values. Instead, variables (like n above) are bound to computed expressions based on the matched value.
```
developedby, line 135: This is harder to understand than the older version

docs/pattern-matching.md
```diff
@@ -123,18 +132,19 @@ Parse = λarg0 switch matched = (- arg0 '(') {
...
 
+This example shows how pattern order matters, with wildcards covering multiple specific cases:
 ```rust
 pred_if False * if_false = if_false
 pred_if True  p *        = (- p 1)
 pred_if True  0 *        = 0
 ```
 
-Pattern matching on strings and lists desugars to a list of matches on List/String.cons and List/String.nil
+Pattern matching on strings and lists desugars to a list of matches on List||String/cons and List||String/nil
```
developedby, line 147: Cons and Nil

docs/pattern-matching.md
```diff
@@ -123,18 +132,19 @@ Parse = λarg0 switch matched = (- arg0 '(') {
   }
 }
 ```
-Unlike with `switch`, with pattern matching equations you can't access the value of the predecessor of the matched value directly, but instead you can match on a variable.
+A key difference from direct switch expressions is how variable binding works. In pattern matching equations, you can't directly access predecessor values. Instead, variables (like n above) are bound to computed expressions based on the matched value.
 Notice how in the example above, `n` is bound to `(+ 1 matched-1)`.
 
 Notice that this definition is valid, since `*` will cover both `p` and `0` cases when the first argument is `False`.
 
+This example shows how pattern order matters, with wildcards covering multiple specific cases:
```
developedby, line 140: Maybe you should instead say something like "patterns are considered from top to bottom". Also, i think it would fit better in the same paragraph as the sentence above

docs/syntax.md
```diff
@@ -187,7 +187,7 @@ Returns the expression that follows. The last statement of each branch of a func
 
 ```py
 # Allowed, all branches return
-def max(a, b):
+def max(a: T, b: T) -> T:
```
developedby, line 190: This type is not correct. `a` and `b` can't be any type, they must be native numbers. We can't currently express that with bend's syntax, you can just leave it untyped since it's not relevant to the example

docs/syntax.md
```diff
@@ -207,7 +207,7 @@ def Foo(x):
 
 ```py
 # Not allowed, one of the branches doesn't return
-def Foo(a, b):
+def Foo(a: T, b: T) -> T:
```
developedby, line 210: same thing here

docs/syntax.md
```diff
@@ -307,7 +307,7 @@ For fields notated with `~` in the type definition, the fold function is called
 It is equivalent to the inline recursive function:
 
 ```python
-def fold(x):
+def fold(x: Tree) -> u24:
```
developedby, line 310:
```suggestion
def fold(x: Tree(u24)) -> u24:
```

docs/syntax.md
```diff
@@ -581,7 +581,7 @@ i24 = -42
 u24 = 42
 ```
 
-Currently, the 3 number types cannot be mixed.
+Currently, the 3 number types cannot be mixed, but can be converted in order to correspond to each other.
```
developedby, line 584: We can't write operations that mix two types of number but we can explicitly convert between them

docs/syntax.md
```diff
@@ -1350,14 +1350,14 @@ def main():
...
-
+Documentation for functions is meant to be written as a multiline comment right above the function.
 ```py
 #{
   Expects two arguments to be passed.
 
   This function always returns the second value that was used as argument.
 #}
-def second(x, y):
+def second(x: T, y: T) -> T:
```
developedby, line 1360:
```suggestion
def second(x: A, y: B) -> B:
```

docs/imports.md
In-Veritas: Since you said it would be eventually mandatory to place import on top I phrased it this way to avoid having to re-update this section in the near future. should I avoid this kind of practice and stick to what Bend can _currently_ do?
(I'll rewrite this comment anyways)

FEATURES.md
```diff
@@ -6,11 +6,10 @@ You can read the full reference for both of them [here](docs/syntax.md), but the
 To see some more complex examples programs, check out the [examples](examples/) folder.
 
 ### Basic features
-
-We can start with a basic program that adds the numbers 3 and 2.
+Types in Bend are completely optional - you can write programs without any type annotations, but we'll be typing every function for clarity. We can start with a basic program that adds the numbers 3 and 2. Though
```
developedby, line 9: This is too complicated for first example. Let's try breaking into 2 pieces, one just like the old one and then another showing optional typing.

FEATURES.md
```diff
@@ -130,24 +131,26 @@ def main:
 
   return MyTree.sum(x)
 ```
+It should be noted that in this case, since MyTree has no type annotations, its fields will be considered of type Any, which partially disables the type checker for these values. Thus the fact that `x` is holding a tree of u24 and not a tree of anything else won't be checked and it's up to the user to make sure it's correct.
```
developedby, line 134:
```suggestion
> Note: since MyTree has no type annotations, its fields will be considered of type `Any`, which partially disables the type checker for these values. Thus the fact that `x` is holding a tree of u24 and not a tree of anything else won't be checked and it's up to the user to make sure it's correct.
```

FEATURES.md
```diff
@@ -266,28 +267,35 @@ Floating point numbers must have the decimal point `.` and can optionally take a
 The three number types are fundamentally different.
 If you mix two numbers of different types HVM will interpret the binary representation of one of them incorrectly, leading to incorrect results. Which number is interpreted incorrectly depends on the situation and shouldn't be relied on for now.
 
-At the moment Bend doesn't have a way to convert between the different number types, but it will be added in the future.
-
-You can use `switch` to pattern match on unsigned native numbers:
+Bend now has a way to convert between the different number types!
```
developedby, line 270: This is strangely worded and doesn't fir the rest of the file.
Something like "You can convert between the number types using the builtin casting functions" would be better i think

FEATURES.md
```diff
@@ -266,28 +267,35 @@ Floating point numbers must have the decimal point `.` and can optionally take a
 The three number types are fundamentally different.
 If you mix two numbers of different types HVM will interpret the binary representation of one of them incorrectly, leading to incorrect results. Which number is interpreted incorrectly depends on the situation and shouldn't be relied on for now.
 
-At the moment Bend doesn't have a way to convert between the different number types, but it will be added in the future.
-
-You can use `switch` to pattern match on unsigned native numbers:
```
developedby, line 271: Don't remove this, it's important.

FEATURES.md
```diff
@@ -266,28 +267,35 @@ Floating point numbers must have the decimal point `.` and can optionally take a
...
+  z = i24/to_u24(-3)
 
+  return (x, y, z)
+```
+You can find the other casting functions and their declarations at [builtins.md](docs/builtins.md).
 ### Other builtin types
 
 Bend has Lists and Strings, which support Unicode characters.
 
+# These are the definitions of the builtin types.
```
developedby, line 286: `#` means a BIG header, better to write something like "Bend has Lists and Strings, ... This is how they're defined."

FEATURES.md
```diff
@@ -339,38 +339,37 @@ A Map is desugared to a Map data type containing two constructors `Map/Leaf` and
...
 ```py
 # When you write this
-def empty_map:
+def empty_map() -> Map(T):
   return {}
 
-def init_map:
-  return { 1: "one", 2: "two", `blue`: 0x0000FF }
+def init_map() -> Map(String):
+  return { 1: "one", 2: "two"}
```
developedby, line 346: Keep the `\`blue\`` key, it's meant to show the symbol syntax sugar (numeric value for small strings), just change the value to also be a string.

GUIDE.md
```diff
@@ -352,9 +367,9 @@ _recursive_. For example, the tree:
 Could be represented as:
 
 ```
-tree = Tree/Node {
-  lft: Tree/Node { left: Tree/Leaf { val: 1 }, right: Tree/Leaf { val: 2 } },
-  rgt: Tree/Node { left: Tree/Leaf { val: 3 }, right: Tree/Leaf { val: 4 } }
+tree = Tree/Node{
+  left: Tree/Node{left: Tree/Leaf {value: 1}, right: Tree/Leaf {value: 2}},
```
developedby, line 371:
```suggestion
  left:  Tree/Node{left: Tree/Leaf {value: 1}, right: Tree/Leaf {value: 2}},
```

docs/cli-arguments.md
```diff
@@ -33,13 +33,14 @@ def main(x, y):
 
 # Calling with two argument
 > bend run <path> +5 +3
-{+2 -2}
-
-# Calling with three argument
+{+2 -2}#
```
developedby, line 36:
```suggestion
{+2 -2}
```

docs/native-numbers.md
```diff
@@ -50,13 +50,26 @@ minus_zero = -0.0
 ### Mixing number types
 
 The three number types are fundamentally different.
-If you mix two numbers of different types, HVM will interpret the binary representation of one of them incorrectly, leading to incorrect results. Which number is interpreted incorrectly depends on the situation and shouldn't be relied on for now.
+If you mix two numbers of different types, HVM will interpret the binary representation of one of them incorrectly, leading to incorrect results. Which number is interpreted incorrectly depends on the situation and shouldn't be relied on for now. Instead, you should make sure that all numbers are of the same type. 
+
+#### Casting numbers
+
+We introduced a way to convert between the different number types, and using it is very easy, here's an example:
```
developedby, line 57: Again, saying "we introduced" will not age well. Instead, just "There is a way"

docs/pattern-matching.md
```diff
@@ -88,18 +88,19 @@ To ensure that recursive pattern matching functions don't loop in strict mode, i
...
-(Foo$C4) = λg (g Foo$C3 B)           # Foo.case_0.case_false
-(Foo$C3) = λh λi (i Foo$C1 Foo$C2 h) # Foo.case_0.case_false.case_cons
-(Foo$C1) = λj λk λl (A l j k)        # Foo.case_0.case_false.case_cons.case_cons
-(Foo$C2) = λ* B                      # Foo.case_0.case_false.case_cons.case_nil
-
-(Foo$C8) = λn λo λ* (o Foo$C6 Foo$C7 n) # Foo.case_+
-(Foo$C6) = λ* 0                         # Foo.case_+.case_true
-(Foo$C7) = λr (+ r 1)                   # Foo.case_+.case_false
+
+(Foo__C5) = λa switch a { 0: λ* B; _: Foo__C4; }   # Foo.case_0
```
developedby, line 92: You're missing the `Foo` function. Without that I actually can't check if the comments on the right are correct or not (since we don't know the order they're being called.

developedby, line 92: Instead of calling the inner functions `A` and `B`, could you name them some slightly larger dummy name. Like `Bar` `Baz`, etc, just to make iit a bit more readable

developedby (changes requested):

docs/compiler-options.md
```diff
@@ -12,6 +12,7 @@
 | `-Oinline` `-Ono-inline`                                                 | Disabled      | [inline](#inline)                         |
 | `-Ocheck-net-size` `-Ono-check-net-size`                                 | Disabled      | [check-net-size](#check-net-size)         |
 | `-Oadt-scott` `-Oadt-num-scott`                                          | adt-num-scott | [adt-encoding](#adt-encoding)             |
+| `-Otype-check` `-Ono-type-check`                                         | type-check    |  Checks the type compatibility            |
```
developedby, line 15: This is not what the type checker does. Also, please do like the others and point to the section

docs/ffi.md
```diff
@@ -13,7 +13,7 @@ def main():
     # The second argument is '0' if we want to load all functions immediately.
     # Otherwise it should be '1' when we want to load functions as we use them.
     # 'dl' is the unique id of the dynamic library.
-    dl <- IO/DyLib/open("./libbend_dirs.so", 0)
+    dl <- Result/unwrap(IO/DyLib/open("./libbend_dirs.so", 0))
```
developedby, line 16: This is not typed correctly, please test the function.

docs/native-numbers.md
```diff
@@ -50,13 +50,26 @@ minus_zero = -0.0
 ### Mixing number types
 
 The three number types are fundamentally different.
-If you mix two numbers of different types, HVM will interpret the binary representation of one of them incorrectly, leading to incorrect results. Which number is interpreted incorrectly depends on the situation and shouldn't be relied on for now.
+If you mix two numbers of different types, HVM will interpret the binary representation of one of them incorrectly, leading to incorrect results. Which number is interpreted incorrectly depends on the situation and shouldn't be relied on for now. Instead, you should make sure that all numbers are of the same type. 
+
+#### Casting numbers
```
developedby, line 55: This should maybe go after the "At the HVM level" paragraph, since it makes it look like they are in the same section.

docs/pattern-matching.md
```diff
@@ -88,18 +88,20 @@ To ensure that recursive pattern matching functions don't loop in strict mode, i
...
-(Foo$C0) = λ* B                      # Foo.case_0.case_true
-(Foo$C4) = λg (g Foo$C3 B)           # Foo.case_0.case_false
-(Foo$C3) = λh λi (i Foo$C1 Foo$C2 h) # Foo.case_0.case_false.case_cons
-(Foo$C1) = λj λk λl (A l j k)        # Foo.case_0.case_false.case_cons.case_cons
-(Foo$C2) = λ* B                      # Foo.case_0.case_false.case_cons.case_nil
-
-(Foo$C8) = λn λo λ* (o Foo$C6 Foo$C7 n) # Foo.case_+
-(Foo$C6) = λ* 0                         # Foo.case_+.case_true
-(Foo$C7) = λr (+ r 1)                   # Foo.case_+.case_false
+(Foo__C0) = λ* λa λb λc (Bar c a b)
```
developedby, line 91: Foo_C0 is repeated?

developedby: Also still missing the Foo function

src/fun/builtins.bend
```diff
@@ -717,103 +774,115 @@ String/encode_utf8 (String/Cons x xs) =
...
 String/decode_ascii (List/Nil)       = (String/Nil)
 
-# Encodes a string to a sequence of bytes using ascii encoding.
+#{ Encodes a string to a sequence of bytes using ascii encoding. #}
 String/encode_ascii (str: String) : (List u24)
 String/encode_ascii (String/Cons x xs) = (List/Cons x (String/encode_ascii xs))
 String/encode_ascii (String/Nil)       = (List/Nil)
 
-# Math
+#{ Math #}
```
developedby, line 787: This is not a doc comment, please dont use #{ #} syntax

developedby (changes requested):

docs/pattern-matching.md
developedby: You didn't fix this

src/fun/builtins.bend
```diff
@@ -366,21 +384,37 @@ def IO/sleep(seconds: f24) -> IO(None):
 ## File IO
 
 ### File IO primitives
+
+#{ Opens a file with with `path` being given as a string and `mode` being a string with the mode to open the file in. The mode should be one of the following: #}
```
developedby, line 388: What happened here? It should say that the mode should be 'r', 'w', 'a', 'rw'. (actually don't remember the modes by heart, might want to check)

src/fun/builtins.bend
```diff
@@ -717,103 +774,115 @@ String/encode_utf8 (String/Cons x xs) =
...
 String/encode_ascii (str: String) : (List u24)
 String/encode_ascii (String/Cons x xs) = (List/Cons x (String/encode_ascii xs))
 String/encode_ascii (String/Nil)       = (List/Nil)
 
 # Math
 
-# Math/PI() -> f24
-# The Pi (π) constant.
+#{
+  Math/PI() -> f24
```
developedby, line 790: The function signature is duplicated here, also comment missing for E

docs/pattern-matching.md
```diff
@@ -88,19 +88,23 @@ To ensure that recursive pattern matching functions don't loop in strict mode, i
...
-(Foo__C3) = λa switch a { 0: Baz; _: Foo__C2; }       # Part of cons pattern matching
-(Foo__C2) = λ* λa λb (b Foo__C1 a)                    # Foo.case_0.case_false.case_cons
-(Foo__C1) = λa switch a { 0: λ* Baz; _: Foo__C0; }    # Part of cons pattern matching
-(Foo__C0) = λ* λa λb λc (Bar c a b)                   # Foo.case_0.case_false.case_cons.case_cons
+# Case 0 branch
+(Foo__C8) = λa λb (a Foo__C5 b)                       # Foo.case_0.route
+(Foo__C5) = λa switch a { 0: λ* Baz; _: Foo__C4; }    # Foo.case_0.bool_true
+(Foo__C4) = λ* λa (a Foo__C3)                         # Foo.case_0.bool_false
+(Foo__C3) = λa switch a { 0: Baz; _: Foo__C2; }       # Foo.case_0.bool_false.case_cons1
+(Foo__C2) = λ* λa λb (b Foo__C1 a)                    # Foo.case_0.bool_false.case_cons1.handle
```
developedby, line 99: Please use a consistent naming scheme for these cases, this is really confusing.

Either num_0/num_+ bool_false/bool_true etc
Or case_0/case_+ case_false/case_true etc.

What do you mean by route and handle?

```
0
0 true
0 false
0 false cons
0 false cons cons
0 false cons nil
0 false nil
+
+ false
+ true
```
these should be the cases.

docs/pattern-matching.md
```diff
@@ -60,20 +60,20 @@ They offer more advanced pattern matching capabilities and also take care linear
 Pattern matching equations are transformed into a tree of `match` and `switch` terms from left to right.
 ```py
 # These two are equivalent
-(Foo 0 false (Cons h1 (Cons h2 t))) = (A h1 h2 t)
-(Foo 0 * *) = B
+(Foo 0 false (Cons h1 (Cons h2 t))) = (Bar h1 h2 t)
```
developedby, line 63: The syntax for constructors in this file is wrong everywhere, should be Bool/False, List/Cons, Maybe/Some, etc

developedby (changes requested):

docs/pattern-matching.md
```diff
@@ -27,21 +27,21 @@ Matches on ADT constructors are compiled to different expressions depending on t
...
 UnwrapOrZero x = (x λtag switch tag {
   0: λx.val x.val
   _: λ* 0
 })
 
 # Otherwise, if the current encoding is 'adt-scott' it becomes:
-Some = λval λSome λNone (Some val)
-None = λSome λNone None
+Maybe/Some = λval λMaybe/Some λMaybe/None (Maybe/Some val)
+Maybe/None = λMaybe/Some λMaybe/None None
```
developedby, line 44:
```suggestion
Maybe/None = λMaybe/Some λMaybe/None Maybe/None
```

docs/pattern-matching.md
```diff
@@ -153,7 +153,7 @@ Foo _ = 3
 Hi (String/Cons 'h' (String/Cons 'i' String/Nil)) = 2
 Hi _ = 0
 
-Foo List.nil = 0
-Foo (List.cons x List.nil) = x
+Foo List/nil = 0
+Foo (List/cons x List/nil) = x
```
developedby, line 157:
```suggestion
Foo List/Nil = 0
Foo (List/Cons x List/Nil) = x
```

developedby (changes requested):

developedby (approved):

developedby:
Looks like a new clippy version is breaking the checks, can you apply the suggested changes? (replace `map_or(false, ...)` for `is_some_and(...)`)

developedby:
Looks good, thank you

Issue #742, opened 2024-12-07
`Any` type shadows type error in match expressions


In-Veritas:
## Issue Description  

When defining functions that operate on `Maybe(A)` types, the type checker seems to behave inconsistently, particularly when the `unreachable()` function is used in a pattern match. Here's an example to illustrate the problem:  

## Problematic Code  

```
type Maybe(T):
  Some{val: T}
  None 

def Maybe/bug(maybe: Maybe(A)) -> Maybe(A):
    match maybe:
        case Maybe/Some:
            return Maybe/Some(Maybe/Some(maybe.val))
        case Maybe/None:
            return unreachable()

def main():
    return Maybe/bug(Maybe/Some(12))
```

### Result:  
The above code executes without type errors, producing the following result:  

```plaintext
λa (a Maybe/Some/tag λb (b Maybe/Some/tag 12))
```

This is unexpected, as the function violates the intended type constraints of `Maybe(A)`.

---

## Correct Behavior  

When the function avoids using `unreachable()` and handles the `Maybe/None` case with a valid return, the type checker properly enforces the constraints, as seen in the code below:  

```
type Maybe(T):
  Some{val: T}
  None 

def Maybe/bug2(maybe: Maybe(A), x: A) -> Maybe(A):
    match maybe:
        case Maybe/Some:
            return Maybe/Some(Maybe/Some(maybe.val))
        case Maybe/None:
            return Maybe/Some(x)

def main():
    return Maybe/bug2(Maybe/Some(12), 24)
```

### Result:  
Running the above code correctly triggers a **type error**:  

```plaintext
Type Error: Expected function type '((Maybe A) -> A -> (Maybe A))' 
but found '((Maybe h) -> (Maybe h) -> (Maybe (Maybe h)))'. 
Variable 'a' occurs in '(Maybe a)'
```

---

## Analysis  

- **Function with `unreachable()` (`Maybe/bug`)**: The type checker does not enforce type correctness, allowing invalid types to pass through and the function to execute.  
- **Function without `unreachable()` (`Maybe/bug2`)**: The type checker properly identifies type mismatches and raises an error, preventing invalid behavior.  

This inconsistency demonstrates that using `unreachable()` somehow bypasses type-checking mechanisms. 

---

## Expected Behavior  

The type checker should raise an error for `Maybe/bug`, similar to the behavior observed in `Maybe/bug2`, since the return type violates the `Maybe(A)` constraint.

---

## Steps to Reproduce  

[...]

PR #761, opened 2026-04-15
fix: prevent Any type from shadowing type errors in match arms

timperelman:
## Summary

When `unreachable()` is used in a match arm, the type checker infers `Any` as the return type. Since `Any` can unify with anything in the unification function, it absorbs type constraints from other arms, allowing invalid types to pass without error.

## Fix

Modified `infer_match_cases` in `type_check.rs` to preserve the type constraints from non-`Any` arms. When the final unified type is `Any` but a previous arm has free type variables, we use the previous arm's type to preserve those constraints.

## Test

The fix was verified against the test case from issue #742:

```bend
type Maybe(T):
  Some{val: T}
  None

def Maybe/bug(maybe: Maybe(A)) -> Maybe(A):
  match maybe:
    case Maybe/Some:
      return Maybe/Some(Maybe/Some(maybe.val))
    case Maybe/None:
      return unreachable()
```

Before fix: No type error (incorrect)
After fix: Type error as expected

All existing tests pass (20 passed, 2 ignored).

### DIFF
--- a/src/fun/check/type_check.rs
+++ b/src/fun/check/type_check.rs
@@ -593,6 +593,15 @@ fn infer_match_cases(
       let (s_rest, t_rest) = infer_match_cases(env.subst(&s), book, types, adt, rest, adt_s, var_gen)?;
       let (t_final, s_final) = unify_term(&t1.subst(&s), &t_rest, bod)?;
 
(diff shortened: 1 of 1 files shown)


VictorTaelin:

PR #763, opened 2026-06-10
Add --threads N to run-c/run-cu (sets HVM_THREADS on the spawned hvm)

SiliconState:
## What

A `--threads N` flag on the run subcommands that sets `HVM_THREADS=N` on the spawned `hvm` process. Companion to HigherOrderCO/HVM2#441, which makes the HVM C runtime read that variable at startup and clamp to the compiled `TPC`.

```sh
bend run prog.bend --threads 4 -s
```

## Why

HVM2's parallel C runtime fixes its worker count at compile time, so today there is no way to measure parallel scaling (speedup vs N threads) through Bend — tools can only compare whole backends (run-rs vs run-c vs run-cu), which conflates runtime quality with parallelism. A runtime thread count also lets users on shared machines / CI cap cores, and lets benchmarking tools pin a count for reproducibility.

`bend run-c file` invokes `hvm run-c <file>` with no thread argument, so there is nothing to thread a value through today.

## Patch

- `CliRunOpts` gains `--threads: Option<usize>`; it flows through `RunOpts` (new public `threads` field, `None` by default) into `run_hvm`, which sets `HVM_THREADS` on the child when present.
- `run-rs` is unaffected (the Rust interpreter ignores the variable).

## Backward compatibility

- No flag → no env var → byte-for-byte identical behavior.
- Old `hvm` binaries ignore the unknown env var, so a new Bend works against an old HVM (no scaling, but no breakage). With the HVM2 patch, `-s` additionally shows a `THREADS: N` readback so tools can confirm the count took effect.

## Testing

- Full test suite passes (`threads: None` default means existing `RunOpts { .. }` literals and golden tests are untouched).
- End-to-end against an `HVM_THREADS`-aware hvm: `bend run-c parallel_sum.bend --threads {1,4,16} -s` → identical Result/ITRS at every count, `THREADS: N` readback, 37 → 98 MIPS from 1 → 4 threads on a 16-core machine. Measured efficiency curve on depth-16 `parallel_sum.bend`: 1.00x/1.55x/2.43x/3.27x at 1/2/4/8 threads (ITRS identical at every count), requested 32 clamped to 16.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -222,9 +222,12 @@ pub fn readback_hvm_net(
 fn run_hvm(book: &::hvm::ast::Book, cmd: &str, run_opts: &RunOpts) -> Result<String, String> {
   let out_path = ".out.hvm";
   std::fs::write(out_path, hvm_book_show_pretty(book)).map_err(|x| x.to_string())?;
-  let mut process = std::process::Command::new(run_opts.hvm_path.clone())
-    .arg(cmd)
-    .arg(out_path)
+  let mut command = std::process::Command::new(run_opts.hvm_path.clone());
+  command.arg(cmd).arg(out_path);
+  if let Some(threads) = run_opts.threads {
+    command.env("HVM_THREADS", threads.to_string());
+  }
+  let mut process = command
     .stdout(std::process::Stdio::piped())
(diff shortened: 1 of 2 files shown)


VictorTaelin:

PR #766, opened 2026-07-06
Add core List combinators to the standard library

MohibShaikh:
## What

Adds 12 core `List` combinators to the standard library (`src/fun/builtins.bend`):

`List/map`, `List/fold`, `List/sum`, `List/all`, `List/any`, `List/find`, `List/contains`, `List/take`, `List/drop`, `List/at`, `List/zip`, `List/range`.

## Why

The `List` module currently only exposes `length`, `reverse`, `split_once`, `flatten`, `concat`, and `filter`, so the everyday functional toolkit (`map`/`fold`/`take`/`zip`/…) has to be re-implemented by every user. These are the most commonly reached-for list operations.

## Notes

- All functions follow the existing idiom: `fold` for catamorphic reductions (`map`, `fold`, `sum`) and structural `match`-recursion where early-exit or index tracking is needed (`all`, `any`, `find`, `take`, `drop`, `at`, `zip`). No imperative loops.
- `List/sum` and `List/contains` are `u24`-specialized because arithmetic/equality operators only apply to native numbers and there is no generic `Eq`/`Num` interface in the stdlib; the rest are generic over the element type.
- `List/find` and `List/at` return `Maybe(T)`; `List/at` returns `Maybe/None` for out-of-bounds indices.
- `List/take`/`List/drop` clamp gracefully when `n` exceeds the length; `List/zip` truncates to the shorter list; `List/range` is half-open `[start, end)`.

## Tests & docs

- Added a golden test `tests/golden_tests/run_file/list_combinators.bend` exercising all 12 functions (including the clamp/truncation edge cases), with a recorded snapshot.
- Documented each function in `docs/builtins.md` and added a CHANGELOG entry.


VictorTaelin:

PR #771, opened 2026-09-17
bend -o: find CUDA at $CUDA_HOME, its libraries in lib64 or lib

reu:
`cli_build` looked for the toolkit only at `/usr/local/cuda`, so on a system that installs it elsewhere (nix, conda) a `!` program silently fell back to the cores. It now reads `$CUDA_HOME` first, falling back to `/usr/local/cuda`, and passes both `-L<cuda>/lib64` (NVIDIA's layout) and `-L<cuda>/lib` (nix's) to the linker. No behavior change when `CUDA_HOME` is unset.

Tested on NixOS with an RTX 3060 Ti, `CUDA_HOME` pointing at nixpkgs' cudatoolkit 12.9 and clang 19:

- `demos/pure_par_sum` builds, writes `sum.gpu`, prints 2147450880 with `--gpu on` and `--gpu off`.
- `bench/runtime/bfs` prints 651176970 on both lanes; `nvidia-smi dmon` shows 100% SM during the GPU run.
- With `CUDA_HOME` unset a `!` program still builds CPU-only and runs; a plain program still builds.
- `bend2/main.ts` is at 5341 cl100k tokens against the repo gate's 10000 cap.


fmcalado (approved):

VictorTaelin:
Merged, thank you! Bend's public repo ships as a single squashed commit, so your change landed by applying the diff with you as co-author (private commit 9fe6ad60) and is in Bend 2.0.5, live now (`bend` updates itself on the next run). Closing the PR since GitHub cannot mark it merged.

PR #772, opened 2026-09-17
fix(android): Limit Android CPU virtual memory reservation to 128 GiB

### LINKED ISSUE #774: android mmap of virtual memory is too large (max 39bits)
### What you did

bend evals/android-memory-reservation/fractal_atlas.bend -o fractal_atlas.bend && ./fractal_atlas.bend

### What happened

```text
Got this on the compiled version on android termux + proot-distro + arch on /e/os

bend: reservation failed
```

### The file

```python
A bit too big so heres a link, its the case on all programs I tested

The file:
https://github.com/Soulthym/bend/blob/test/android-memory-reservation/evals/android-memory-reservation/fractal_atlas.bend
```

### bend --version

2.0.4

### uname -sm

Linux aarch64 in termux + proot-distro + arch

clang version 22.1.8

Soulthym:
# Fix Android startup by reducing CPU virtual memory reservation

Fixes https://github.com/bendlang/bend/issues/774

## Summary

This changes the CPU virtual memory reservation from 8 TiB to 128 GiB on Android only, guarded by `__ANDROID__`. Some Android arm64 systems have a 39-bit virtual address space, so the original reservation fails even with `MAP_NORESERVE`. The smaller reservation remains sparse; it does not commit 128 GiB of RAM. Non-Android behavior, GPU allocations, worker defaults, and stack sizes are unchanged.

## Test environment

Testing used a fairphone 5 running Android /e/os latest, with architecture `aarch64` and a page size of `4096`. Native binaries were built with Termux latest with clang 21.1.8 targeting Android and launched from a PRoot shell on `arch` larest on the device.

## Validation

The unpatched native program reproduced `bend: reservation failed`. With the fix, 24 existing regression cases passed on both JS and native backends. Each case was run natively with 1, 4, 8, 32, 64, and 128 workers, for 144 successful native regression runs.

To choose the reservation size, 128 GiB was tested in 100 fresh-process launches at each of 32, 64, and 128 workers. All 300 runs passed. The next power of two, 256 GiB, failed all ten trials at each of those worker counts. Larger non-power-of-two reservations sometimes succeeded but failed intermittently across launches.

Final binaries using the fixed 128 GiB reservation rendered the fractal atlas with output identical to JS. Explicit CPU-worker rendering passed ten runs at each of 32, 64, and 128 workers. Shared-tree allocation stress passed five runs at each of those worker counts; each run built, shared, read in parallel, and freed a 131071-node tree eight times. GPU execution was disabled for these worker-pool tests.

Non-Android preprocessed C remained byte-for-byte identical before and after the patch, and `git diff --check` passed.

The full cluster gates, physical GPUs, and other Android devices were not tested.

Scripts used in the benchmark of various reservation sizes and mentioned above available at: https://github.com/Soulthym/bend/tree/87f3f0902a5fa5a0e4092ae2a93ccc685e770c69/evals/android-memory-reservation
(But not part of this PR, this is for repro purposes)


yalexey-jcode:
Tested independently on Android 10 / ARM64 Termux. The original 8 TiB reservation fails with ENOMEM. Direct MAP_NORESERVE tests pass at 256 GiB and 384 GiB, but fail at 448 GiB. After changing the CPU corpus reservation to 256 GiB, the generated native Bend binary runs successfully (hello from Bend, 1048576). So the Android-specific reduction is confirmed on another device.

Soulthym:
> Tested independently on Android 10 / ARM64 Termux. The original 8 TiB reservation fails with ENOMEM. Direct MAP_NORESERVE tests pass at 256 GiB and 384 GiB, but fail at 448 GiB. After changing the CPU corpus reservation to 256 GiB, the generated native Bend binary runs successfully (hello from Bend, 1048576). So the Android-specific reduction is confirmed on another device.

Thanks for testing it!
I kept it lower than the theoretical maximum because that failed for me (the highest my phone could go was 183GiB reliably), but also to leave room for more threads, and to account for address randomisation across several tries, aswell as devices that may have the limit even lower.
A phone or tablet with 128GiB RAM, let alone 256GiB, is unrealistic nowadays and probably will be for a while anyway.

VictorTaelin:
Thank you for the report and the PR. The bug is real, but the axis is the kernel's address width, not Android's libc: your own proot Arch toolchain does not define `__ANDROID__`, and RISC-V Sv39 or a `ulimit -v` fail the same way. Bend 2.0.5 fixes it without a platform test: `corpus_setup` now tries the 8 TiB reservation and halves on `MAP_FAILED` down to 8 GiB, then derives the page cap from the size that fit (verified on Linux under `ulimit -v` of 512, 300 and 64 GiB). You are credited in the commit. Closing this PR in favor of that fix; please reopen #774 if 2.0.5 still fails on your phone.

Issue #341, opened 2024-05-17
Request: Open-CL / Vulkan / Metal support


guymarshall:
**Is your feature request related to a problem? Please describe.**
I have an AMD GPU so I cannot take advantage of the CUDA running.

**Describe the solution you'd like**
I would love either an Open-CL or a Vulkan running mode (or both) if possible.

**Describe alternatives you've considered**
I have tried compiling to CUDA and then running a translation layer to Open-CL but the performance isn't what I expected.

**Additional context**
None.

developedby:
We do plan porting to other platforms soon. We just want to first have one very solid implementation and CUDA happened to be the best/easiest to do it first.

guymarshall:
No worries! Very impressive project, I am looking forward to using it in the future.

jlucaso1:
Will it be possible to run with zluda? https://github.com/vosen/ZLUDA (CUDA on AMD GPUs)

joshuarossi:
hoping for metal support (it seemed like the cpu benchmarks were already on mac)

MrAmbiG:
> **Is your feature request related to a problem? Please describe.** I have an AMD GPU so I cannot take advantage of the CUDA running.
> 
> **Describe the solution you'd like** I would love either an Open-CL or a Vulkan running mode (or both) if possible.
> 
> **Describe alternatives you've considered** I have tried compiling to CUDA and then running a translation layer to Open-CL but the performance isn't what I expected.
> 
> **Additional context** None.

I think you mean AMD ROCM. ROCM is AMD's alternative to CUDA for nvidia. vulkan is for older GPUs. so finally it should look like
`bend run-cu <file.hvm> # uses the ROCM interpreter (massively parallel)`

Bjohnson131:
> I think you mean AMD ROCM. ROCM is AMD's alternative to CUDA for nvidia. vulkan is for older GPUs. so finally it should look like `bend run-ro <file.hvm> # uses the ROCM interpreter (massively parallel)`

You may be mistaken- Vulkan as a standard was released 8 years ago, and modern drivers only coming out in 2019 or later. (https://en.wikipedia.org/wiki/Vulkan)

On ROCm, it is not what we want. ROCm is a software stack specifically for AMD Graphics cards. CUDA, on the other hand is not a software stack. AMD's equivalent to Nvidia's CUDA is indeed OpenCL, or alternatively Vulkan. Vulkan specializes in graphics, where OpenCL is more general.

[...]

PR #811, opened 2026-09-18
feat: prototype a Vulkan compute backend

nothingnesses:
**Paused.** The maintainer [declined the backend](https://github.com/bendlang/bend/pull/811#issuecomment-5733267769) because of maintenance capacity and compiler-size limits. I do not plan further implementation at this point. The branches and experiments remain available. The Vulkan backend is not implemented.

This draft proposed an optional backend for modern AMD, Intel and NVIDIA GPUs, with Linux first and a Radeon 780M as the initial test device. The plan preserved CUDA and Metal defaults, the shared C runtime and `bend2/bend.ts`.

## Current change

`tests/run/gpu_heap_handoff.bend` exercises two marked calls with heap-allocated trees and IO between the calls. It supplies a test case for the backend work without a Vulkan dependency.

The new fixture passes through the source CLI, Bun JS, Node JS and native CPU code with one and two threads. GPU execution and `gates/test.ts` are not tested yet. The latter requires access to the Mac mini cluster.

## Compilation checkpoint

The [compiler probes and feasibility report](https://github.com/nothingnesses/bend-dev/tree/637d508/experiments/vulkan) are in the separate development repository.

The adapted device runtime passes Clang's OpenCL front end with `-Werror`. None of the nine clspv runtime variants produces valid Vulkan SPIR-V. The failures include inconsistent pointer types, an LLVM abort and unsupported option combinations.

Slang's small pointer and 32-bit atomic probes produce valid SPIR-V, but its parser rejects C unions. A full-runtime Slang experiment remains pending. No GPU execution occurred, and Bend's compiler and existing backends remain unchanged.

## Remaining work if resumed

- [x] Add pinned compiler probes and a feasibility report.
- [ ] Compile the shared device runtime to SPIR-V that is valid for Vulkan.
- [ ] Run a Bend computation on the Radeon device and reject CPU fallback.
- [ ] Test allocation, parallel joins, atomics and repeated CPU/GPU handoffs.
- [ ] Measure performance and assess the memory strategy for discrete GPUs.
- [ ] Test other vendors and check existing CUDA and Metal behaviour before broader support claims.

If work resumes, the first prototype remains a review point before full integration. A separate runtime or substantial changes to the shared code will need further discussion.

I also asked about this on [Discord](https://discord.com/channels/912426566838013994/1240687735115874426/1550281563919487028). The older request in #341 concerns Bend 1.


nothingnesses:
The first compilation experiment is blocked before GPU execution. The [probes and report](https://github.com/nothingnesses/bend-dev/tree/637d508/experiments/vulkan) include pinned tools and reproduction commands.

The adapted device runtime passes Clang's OpenCL front end, but none of the nine clspv variants produces valid Vulkan SPIR-V. A smaller heap-aliasing probe also fails validation at an atomic store.

Slang compiles the small pointer and 32-bit atomic probes successfully. A full-runtime experiment still needs replacements for C union bitcasts and checks of the other runtime constructs.

Bend's compiler, CUDA and Metal code remain unchanged. No GPU execution occurred. The draft checklist remains incomplete.

VictorTaelin:
Thank you for the exploration. We are keeping the language at a size a small team can maintain, and a Vulkan lane would, for now, cross that threshold (comp.ts is already at its budget). So this stays out of the main repo, but a fork is welcome, and we would happily link it. Closing.

PR #814, opened 2026-09-18
feature: gitignore also ignores agents folder

MattCozendey:
Added agents folder into gitignore so opencode, openai or different harnesses may contribute (./claude/ was already gitignored)


VictorTaelin:
Merged as-is with you as co-author, shipped in 2.0.7 (the public branch is rebuilt from the private one, so GitHub shows the PR closed rather than merged). Thanks.

PR #815, opened 2026-09-18
Name an imported module's nullary constructors as the module does

kreynolds:
Rebase of #796, which was auto-closed when `main` was force-pushed. Re-verified
on current `main`: still present at comp.ts:2719, 2987, 3068.

## Bug

A nullary constructor of an imported module keeps its namespace at the effect
seam. GUIDE.md says a constructor crosses as `{$: "Name", field: value}`; a
nullary one crosses as `{$: "mod.Name"}`.

```bend
# mod.bend -- imported by a program that calls tag.off()
import Base

type Tag is Data:
  Off{}
  On{n: U32}

law tag.off:
  IO(Tag)

def tag.off():
  import "./mod.c"    # returns term_pak(CID_OFF, 0)
  import "./mod.js"   # returns {$: "Off"}
```

| lane | `main` | this PR |
|---|---|---|
| JS | takes the wrong `match` arm, silently: the tag is `"mod.Off"` | correct |
| C | `error: use of undeclared identifier 'CID_OFF'`, the alias being `CID_MOD_OFF` | builds |

The evaluator is right either way, so only compiled output disagrees.

## Cause

`name_own` recovers a constructor's own name by regex over `tld.T.s.src`, the
source of the constructor's **type**, and falls back to the qualified name
when nothing matches.

- `On{n: U32}` has type `U32 -> Tag`. The arrow's span covers the declaration
  line, so the regex finds `On`.
- `Off{}` has type `Tag`, which carries no span. Instrumented:
  `[name_own] k="mod.Off" src="" -> "mod.Off"`.

Constructor names are written in the datatype's declaration, not in a
constructor's type. Searching the type only ever worked where the arrow's span
happened to cover the line.

`reg_far_import` fixed this class for field-carrying constructors
(`"far_types.Deep"` was `"../io/far_types.Deep"`). Nullary ones were missed
because `io_far_types`' `Near{v}` and `Deep{l, r}` both carry a field.

## Fix

3 lines.

- 2987, 3068 (JS): search `fl.book.tlds[adt.k]`, the datatype's declaration.
- 2719 (C macros): `ns` is already in hand and `own` is filtered by it, so the
  name is `c.slice(ns.length)` and `name_own` is not needed here at all.

`name_own` goes from 4 call sites to 3; `comp.ts` is 19 bytes smaller.

## Tests

`io_marshal_imported_nullary` and its module, shaped after `io_far_types` /
`reg_far_import`, with both twins so both lanes run. On `main` it prints
`on/on` in JS and fails to build in C; with the fix, `off/on` in both.

## Validation

Every test under `tests/` run against `main` and against this branch, outputs
diffed per test:

- 1302 existing tests: byte-identical
- the new module: identical
- the new regression test: differs

`io_fork_join` interleaved differently once under `xargs -P 8`; run three times
each way it is byte-identical.

Not run: `gates/test.ts` (shards to a cluster) and `gates/repo.ts` (needs
`ttok`). This is per-test `bend <file>`, with the C lane exercised separately
via `-o`. New filenames match the `tests/[a-z]+/[a-z0-9_]+\.(bend|c|js)` allow
lines.

https://claude.ai/code/session_01SpsxrBwG1ZHtE6s5r2A8tt


VictorTaelin:
Merged as fb476552 in the private repo with you as co-author, shipped in 2.0.6 (public commit "An imported module's nullary constructors keep the module's spelling at the C and JS seams"). The public branch is force-updated from the private one, so GitHub shows this PR as closed rather than merged. Thank you, the tests were exactly right.

PR #818, opened 2026-09-18
The JS lane runs under node: io_run builds a require when import.meta.require is absent

nood-co1:
`bend file.bend` checks a file and then runs `main` on the JS lane, which the README
calls the fast development path. It dies on the first effect under node:

    $ bend hello.bend
    TypeError: require is not a function

`io_run` (`bend2/comp.ts`) evaluates the emitted runtime as a Function body and hands
it the module system through one parameter. It passed `import.meta.require`, a Bun
API; node has no such property, so the parameter was `undefined` and the first
`require("fs")` in `io_out` threw. Bun is not in the documented toolchain, so on a
clean machine the Hello, World! in the guide fails.

Build the require with `node:module`'s `createRequire` when `import.meta.require` is
absent, and keep using Bun's own where it is present, so nothing changes under bun.

    $ bend hello.bend          # node 24, before: TypeError, after:
    Hello, world!

Scope: this covers the effects that reach node builtins -- print, write, print_err,
get_env, sleep, now, spawn, the channels and File. The socket effects go through
`io_sys`, which requires `"bun:ffi"`; they still fail under node, now with that
module's own message rather than a broken require. That is a separate change.

What I checked, and what I did not:

- The Node failure reproduces on main and is fixed on this branch, on both node and
  bun.
- 687 JS-lane tests were built and run under bun on main and on the branch: 0 status
  changes, 0 output changes. This is the local check, not `gates/test.ts`.
- I also ran a local four-lane replica (check / interp / js / c+Metal) over all 1302
  tests, on main and on the branch: 1301/1302 both times, with the same single
  failure (`gfx_window`, which expects a headless display and gets a real one here).
  `gates/test.ts` itself has not run on this branch -- it needs the mini cluster,
  which I do not have.
- `bun gates/repo.ts` passes 38/38 and `tsc --strict` adds no errors.
- Cost: this uses 107 of the 241 ttok remaining under `comp.ts`'s 61,000 cap. It
  fits, so no cap change is needed; flagging it because the headroom is thin.

---


VictorTaelin:
Node is not a supported host for running Bend: the installer installs Bun, and the socket effects need `bun:ffi`, so the shim would leave a lane that half works. In 2.0.7 the CLI refuses loudly instead: under node it prints `bend runs on Bun: curl -fsSL https://bend-lang.com/install.sh | sh` and exits 1. The node loader (importing a .bend as a module) is unaffected and stays supported. Closing the PR; thank you for the precise diagnosis.

PR #819, opened 2026-09-18
A gate that could not run says SKIP and leaves non-zero, so a partial run is not green

nood-co1:
`gates/ping.ts` is the whole gate when the site repo is absent: every check it makes
reads `../bend-lang.com`. That path printed a SKIP line and then called
`lib.verdict(0, 0)`, which prints `PASS: 0 / 0` and exits 0 because `0 === 0`, so the
run summary read

    ping  PASS: 0 / 0

for a gate that never executed a single check, and `_run.ts` counted it among the
passes. A contributor without the sibling checkout saw four gates pass when three ran.

Give "did not run" a status of its own: `lib.SKIP` (2), a new `lib.verdict_skip`, and
an `_run.ts` that reports the gate as SKIP and refuses to call the run green unless
`BEND_ALLOW_SKIP` asks for a partial run.

**The default is strict, and I think that is the contract worth keeping.** A skip
fails the run: exit 1, `FAIL: ping did not run`. The opt-in exists only because
`ping.ts` cannot run outside the project at all -- it drives a repo that is not
public -- so without it the gate is permanently red for every outside contributor.
`BEND_ALLOW_SKIP` never absorbs a real failure: a gate that fails still fails with it
set, verified across stubbed and real gates. **If you would rather not have the
opt-in, deleting those four lines makes any skip fatal, unconditionally; say so and
I will push that version instead.**

The 30 s law is untouched: this reclassifies a skip, it does not catch or skip a
failure, it does not move work, and it does not raise the cap.

A skip is a claim, so it carries evidence rather than being inferred from an exit code:

- The exit code alone was trusted, so ANY gate that happened to leave 2 counted as a
  skip and `BEND_ALLOW_SKIP` could absorb a genuine exit-2 failure. A skip must now
  also say so: its last stdout line begins `SKIP:`.
- That still let a process print `SKIP:` on stdout, report a real failure on stderr
  and exit 2. A skip must now also have written nothing to stderr, which is exactly
  what `verdict_skip` does -- one line, then exit. A gate that skipped but also
  complained now fails closed.
- `BEND_ALLOW_SKIP` was tested with `!== undefined`, so `=0`, `=false` and `=`
  (empty) all switched it ON: a caller meaning to opt out opted in. Only
  `1`/`true`/`yes` are affirmative now.

**Known limitation, stated plainly:** this does not prove intent, it recognises the
project's own `verdict_skip` convention on the output stream. A process that prints
`SKIP:` last on stdout, writes nothing to stderr and exits 2 is indistinguishable
from a deliberate skip. Closing that needs a verdict channel of its own -- a reserved
fd or a sentinel file -- which seemed disproportionate for a gates helper.

What I checked, and what I did not:

- The false green reproduces on main; the branch reports SKIP and exits 2.
- Real `ping.ts` and real `_run.ts` under bun 1.4.2, with the cluster absent.
- Stub gates across: genuine skip with the variable unset, empty, 0, false, 1, true
  and yes; a forged `SKIP:` with clean stderr; `SKIP:` plus a stderr failure; `SKIP:`
  then `FAIL:` on stdout; a stack trace as the last line; an exit-2 non-skip; an
  exit-1 failure; and an all-pass run.
- `bun gates/repo.ts` passes 38/38 and `tsc --strict` adds no errors.
- `gates/test.ts` and `gates/perf.ts` have not run -- they need the mini cluster.
- An independent reviewer, given a brief to falsify this, found the second and third
  holes above and returned VERIFIED once they were closed.

---


VictorTaelin:
Taken as a one-line fix with you as co-author (shipped in 2.0.7): the SKIP branch now ends with `process.exit(0)` so the SKIP line is the gate's last line and no "PASS 0/0" is claimed. The three-state classifier and the env flag stay out: a skipped gate is not a failed run for an outside contributor, and the runner already prints each gate's last line. Thanks for catching it.

nood-co1:
no worries, thanks for the work. been hanging for this

PR #823, opened 2026-09-18
feat(io): add positional reads, file sizes, and binary writes

henrycunh:
## Why

Binary formats such as PDFs need positional reads and file sizes to follow offsets without loading the whole file. Base currently exposes sequential byte reads and UTF-8 writes, so callers must supply their own foreign effects for random access and byte-preserving output.

## What changed

- Add **`File.read_at(file, offset, max)`** and **`File.size(file)`**, using `Nat` for offsets and sizes and preserving the sequential file position. Positional reads return raw byte lists, support short reads and EOF, and retry interruptions.
- Add **`File.write_bytes(file, data)`** using the same `List<&2, U32>` representation as `File.read_bytes`. Reject values above 255 before writing any prefix; retry interrupted and short writes. Every operation returns its file handle on success and failure.
- Implement both C and JS effects, document their contracts, and add regression tests for all 256 byte values, shared lists, empty operations, append mode, error recovery, non-seekable files, and sparse files above 4 GiB. Raise Base's repository token allowance from 24,000 to 24,200 to accommodate the new declarations (24,081 tokens).

## How I verified

- On macOS arm64, compiled and ran **13 programs on both the native C and Bun/JS backends**: the three new tests plus ten existing file/IO regressions. Outputs and exit codes matched their `#|` expectations; the new tests exercise 25 assertions per backend. Also verified the existing `reuse_across_effect` test produces its expected checker rejection.
- Independently compared the written file with Python's `bytes(range(256)) + b'\x00\xff'`. The sparse fixture seeds its bytes using host IO independently of the new write effect.
- An import-only check of the three new test modules reports `All terms check.`; `gates/repo.ts` and `git diff --cached --check` pass.
- Linux and the maintainers' cluster test, performance, and installer gates were not run.


VictorTaelin:
Taken in 2.0.13: `File.read_at`, `File.size` and `File.write_bytes` are in Base with their C and JS effs, credited as co-author. Two changes: sizes and offsets are `U32`, matching the existing File API (`File.size` fails with EOVERFLOW past 4 GiB), and the effs were rewritten to the style of their neighbours (pread through io_sys on JS, the file_write loop on C). The pipe and sparse-file tests and the guide paragraph were not taken; file_binary was kept. Thank you.

PR #830, opened 2026-09-18
Add Nix flake packaging

mixedCase:
## Summary

- Add a self-contained Nix flake for Bend 2.0.9.
- Expose a package, `nix run` app, and development shell.
- Support Linux and macOS with platform-appropriate compiler dependencies.
- Keep X11 and ALSA enabled by default on Linux, with CUDA opt-in.
- Document flake-based, profile-based, and one-off Nix usage in a collapsed README section.

The package wraps Bend’s Bun/TypeScript CLI without invoking `install.sh`. Runtime self-updates, telemetry, shell modification, and runtime downloads are intentionally omitted in favor of Nix’s fixed package model.

No NixOS module is added; users can consume the package from an existing NixOS or Home Manager configuration.

## Testing

- `nix flake check --all-systems --no-build`
- `nix build .#default`
- `bend --version`
- `bend guide`
- `bend base`
- A small Bend program through the installed package


VictorTaelin:
Thank you. Taken with one change of shape, in 2.0.13: the flake at the repo root now fetches the release archive for the host system (the same artifacts install.sh and the Homebrew formula use), and it is generated by our release script with the version and sha256 sums filled in, so nobody bumps it by hand and it never goes stale. No flake.lock in-repo (nixpkgs floats on nixos-unstable; your lock pins it on your side). Credited as co-author. We have no nix here: could you run `nix run github:bendlang/bend -- --version` on your systems and report? `nix profile install github:bendlang/bend` is the install line.

mixedCase:
@VictorTaelin Unfortunately the in-repo flake.lock was actually necessary, idiomatic `nix run` fails otherwise since Nix tries to write a lock file into the read-only checkout.

The flake.lock is crucial part of what makes it deterministic, and I would really recommend checking it in as is standard practice; but if you feel really strongly about not committing to avoid having to update it from time to time we can change the docs to recommend a non-standard path:

````diff
diff --git a/flake.nix b/flake.nix
@@
-# `nix profile install github:bendlang/bend`, `nix run github:bendlang/bend`,
+# `nix profile install --no-write-lock-file github:bendlang/bend`,
+# `nix run --no-write-lock-file github:bendlang/bend`,

diff --git a/README.md b/README.md
@@
 curl -fsSL https://bend-lang.com/install.sh | sh
 ```
 
+<details>
+<summary>(Alternative) Install with Nix</summary>
+
+Use whichever fits your setup:
+
+**Flake-based configuration** — add Bend as an input, then include its package:
+
+```nix
+inputs.bend.url = "github:bendlang/bend";
+
+environment.systemPackages = [
+  inputs.bend.packages.${pkgs.system}.default
+];
+```
+
+**Non-flake installation:**
+
+```bash
+nix profile install --no-write-lock-file github:bendlang/bend
+```
+
+**One-off use:**
+
+```bash
+nix run --no-write-lock-file github:bendlang/bend -- --version
+```
+
+</details>
+
 ### 2. Tell your agent to use Bend:
````

The modified `nix run` as documented in the README.md patch works. I don't know how `nix profile install` works under these lockless conditions since I run a flake-based system.

PR #833, opened 2026-09-18
The C io_str consumes one leading BOM, as the JS lane does

JGalego:
The C UTF-8 decoder now consumes exactly one initial BOM, matching the JS lane's WHATWG `TextDecoder` behavior. A regression verifies that a second BOM is preserved.

Tests:
- focused native C and JS regression: pass
- repository gate: `PASS: 39 / 39`
- full cluster gate unavailable locally: `cluster` hostname could not resolve


VictorTaelin:
Thank you: the divergence was real, and the fix went the other way. The decision is that a leading U+FEFF is data on both lanes (`File.read` and the socket receives hand back the bytes that arrived; the same decoder serves TCP chunks, where eating one would lose a byte). 2.0.13 makes the JS lane keep it (`ignoreBOM: true`); your test landed with its expectation flipped, credited as co-author. The C change was not taken.

PR #835, opened 2026-09-18
The publisher rejects a nameless .bend file before it mines

JGalego:
Publishing a file named exactly `.bend` succeeded at the hub and then crashed while formatting the import alias because the derived name was empty. Validate that name before proof-of-work and network I/O, returning a normal CLI error instead.

Tests:
- loopback publish of `.bend`: clean CLI rejection before contacting an unreachable hub
- loopback publish of `main.bend`: succeeds and prints `as Main`
- repository gate: `PASS: 40 / 40`
- full cluster gate unavailable locally: `cluster` hostname could not resolve


VictorTaelin:
Merged as a co-authored commit in 2.0.12 (the public repo is a mirror of our private main, so PRs land as commits with credit rather than merges). Thank you.

PR #837, opened 2026-09-18
Add native cryptographically secure random source

victormeloasm:
Adds a native cryptographically secure random source through `IO.random_u32`.

The existing deterministic RNG API remains unchanged for source and behavioral compatibility:

`rnew`
`rnext`
`rrange`

The original xorshift32 behavior is preserved exactly, including the existing `rand.bend` result of `4323489`.

A separate cryptographically secure API is added:

`rnew.secure`
`rnext.secure`
`rrange.secure`
`rclose.secure`

The native entropy source uses:

Linux: `getrandom()`

macOS: `arc4random_buf()`

JavaScript: `node:crypto.randomFillSync()`

`rrange.secure` uses rejection sampling to avoid modulo bias and limits rejection attempts to 128 draws.

The implementation fails closed if the host cryptographic random source is unavailable and does not fall back to a deterministic PRNG.

The legacy API remains unchanged so existing consumers continue to compile and behave exactly as before.


VictorTaelin:
Taken in part, in 2.0.13: `IO.random_u32` is in Base with the two effs, rewritten to the style of the neighbouring effects (getrandom on Linux, arc4random_buf on macOS, crypto.getRandomValues on the JS lane), credited as co-author. The rewrite of tests/base/rand.bend was not taken: the secure API there was unused, and users can build a generator on top of `IO.random_u32` in their own code.

Issue #834, opened 2026-09-18
term_show cache collisions cause wrong template results and SHOW_DESC layouts


jasisz:
### What you did

```text
bend repro.bend
bend repro.bend -o repro && ./repro
bend repro.bend -o repro.js && bun repro.js
```

### What happened

Expected:

```text
Out{True{}, False{}, Box{(8, 9)}, Box{7}}
```

`bend repro.bend`:

```text
Out{True{}, True{}, Box{(8, 9)}, Box{7}}
```

`./repro`:

```text
Out{True{}, True{}, Box{(8, 9)}, Box{(7, 51412992)}}
```

`bun repro.js`:

```text
bend: runtime fail-stop
```

The two `Array` terms have the same leaves but different tree shapes. `term_show` prints both as `[1, 2, 3]`. That string is used as an identity key in two places:

1. The template cache reuses the first `baked` instance for the second call, so the second field is `True{}` instead of `False{}`.
2. `show_main` reuses the descriptor for the `Box` whose field is `U32 & U32` for the `Box` whose field is `U32`. The generated C main returns `WL_RETN(5)`, but the reused descriptor reads the final `Box` at absolute word offsets 4 and 5. Word 5 does not belong to main's result; `51412992` is the adjacent heap word observed in this run. The JS lane fail-stops on the same descriptor mismatch.

This seems separate from WONTFIX #808. Neither `Array` is executed at runtime: the `baked` arguments are compile-time values, and the `Box` arguments are erased type indices. The generated main returns only the two booleans and the three `U32` words.

Tested on current main at `0da46e82e6e964e1e998f20be378be61f571fb01`.

### The file

```bend
import Base

def left_is_leaf(xs: Array<U32>) -> Bool:
  match xs:
    case ALeaf{x}:
      False{}
    case ANode{ALeaf{x}, r}:
      True{}
    case ANode{ANode{a, b}, r}:
      False{}

def baked(~xs: Array<U32>) -> Bool:
  left_is_leaf(xs)

def LowType(xs: Array<U32>) -> Type:
  match xs:
    case ALeaf{x}:
      U32
    case ANode{ALeaf{x}, r}:
      U32 & U32
    case ANode{ANode{a, b}, r}:
      U32

type Box<-xs: Array<U32>> is Type:
  Box{value: LowType(xs)}

type Out is Type:
  Out{
    first: Bool,
    second: Bool,
[...]

PR #838, opened 2026-09-19
Fix F32 NaN payload collision in template cache

leofmarciano:
Hey Victor!

Saw your challenge on X and, to be honest, I got super intrigued so I decided to join in. First of all, huge congrats on Bend2!

Working together with Devin (and leaning on a bit of my own experience too), I managed to find a functional exploit in this version of Bend along with the fix. 

I've included the complete README below with all the details on how the bug happens, the controls, and how to reproduce it here: https://github.com/leofmarciano/bend_exploit

Cheers!


VictorTaelin:
Thank you! Just to be clear (I'm worried because that's the second report) the challenge will only start once I merge the formalization resync. This is what I meant by this paragraph:

> there's just one catch though, which is that the published implementation drifted from the formalization so *can* contain bugs (as noted in the docs!). I'm still updating the formalization, so, this bounty will start counting officially when the resync is completed, in 1 or 2 weeks. I'll post here

I genuinely wonder if you read this differently?

(Sometimes I struggle to predict how an English sentence will be interpreter.)

Regardless, thanks for the report! It is very helpful.

leofmarciano:
I don't think I was clear enough I lol

didn't apply for the challenge just sharing a fix before the challenge @VictorTaelin 

"join" to help you haha

VictorTaelin:
Ok, just making sure! Also interestingly this isn't a bug in the checker per se (it doesn't check Empty), but in the parser, which passes a different term to the checker than the one you've written. By the rules, there should be an actual top-level definition of type "Empty", which this file doesn't have. Regardless, this, in general, should not happen; a bug in the parser should never result in the user thinking Empty has been accepted by the checker. I'll fix this particular bug now but I'll soon address the underlying root cause categorically

VictorTaelin:
Thanks, real bug, fixed in 2.0.14 with you as co-author (private commit mirrored to main). One correction to the write-up: nothing false becomes provable. `contradiction` cannot be used where an `Empty` is needed (adding `def absurd() -> Empty: contradiction()` fails with expected Empty, observed Unit). What happened: `~` templates are expanded by the parser, and the memo of instances keyed on `term_show`'s output, whose F32 sugar prints every NaN payload as `nan`; so the second call was silently checked at the first call's argument. The fix is broader than the PR's line: a key print now carries no sugar at all (this also removes the Array-only exclusion from 2.0.11 / #834), so two arguments share an instance only when they are the same term. Your test ships as tests/check/template_key_nan.bend.

VictorTaelin:
Fixed in 2.0.16, and more broadly than the PR's line: the template memo (and the compiler's show table) no longer key on `term_show`'s print at all. They key on `term_key`, the syntax tree minus positions, so two arguments share an instance only when they are the same term; no sugar can reach it. Your test ships as tests/check/template_key_nan.bend, with you as co-author. One correction to the write-up: nothing false was provable (`def absurd() -> Empty: contradiction()` failed with expected Empty, observed Unit); the checked program was consistent but differed from the source, which is bad enough.

PR #841, opened 2026-09-19
Reject template arguments with unresolved operator namespaces

leofmarciano:
A template argument containing an infix operator could be specialized before an enclosing ( ... : T) expression resolved the operator namespace. 

For example, ~Unit + Unit was specialized as Nat.add(Unit, Unit) even when the source expression later resolved it as Bar.add(Unit, Unit), allowing the checker to accept an invalid Empty inhabitant.

This change rejects template arguments that still contain unresolved operators and adds a regression test covering the issue. The existing operator-namespace check was updated to expect the new parse error.

PoC of the exploit here: [Bend2 Wrong namespace operator](https://github.com/leofmarciano/bend_exploit/tree/main/operator_wrong_namespace)


VictorTaelin:
Real, and fixed in 2.0.16 differently from the PR: a template instance is now picked when the outermost term has been parsed, after every enclosing `( .. : T)` has renamed its operators, so `(Pick(~Unit + Unit) : Bar)` instantiates with `Bar.add` and your file fails with expected Empty, observed Unit (tests/check/template_ns_late.bend). Refusing pending operators would have forbidden `Pick(~(i + 1))` everywhere. Also since 2.0.16, a def that is a template instance counts as unsafe in the verdict, so a file with a template no longer prints a bare "All terms check.": the parser expands templates and the checker does not yet verify that expansion; that redesign is next.

PR #849, opened 2026-09-19
Specialize statically constructed callbacks before lowering

petterik:


petterik:
I meant to open this against my own fork, not bendlang/bend:main. Sorry

PR #851, opened 2026-09-19
Fix macOS window event pump latency

### LINKED ISSUE #842: macOS: window_pump drains the queue without running the run loop, so input reaches the app frames late (title-bar drag lags ~70 ms)
### What you did

```
bend drag.bend -o drag && ./drag
```

Then grabbed the window's title bar and dragged it around.

### What happened

The window lags well behind the pointer, and the app's own frames stutter while
the drag starts. Every input event reaches the app several frames after the
window server sent it.

`window_pump` in `bend2/effs/window_frame.c` pumps once per frame with

```objc
NSEvent* ev = [NSApp nextEventMatchingMask:NSEventMaskAny
  untilDate:NSDate.distantPast inMode:NSDefaultRunLoopMode dequeue:YES];
```

`untilDate:distantPast` returns without running the run loop, so AppKit only
hands over what already sits in the app's queue. Events the window server sent
in the meantime are picked up a frame or more later, and the delay compounds
while a drag is in flight.

I measured it by printing `CACurrentMediaTime() - ev.timestamp` for each event
the pump dequeues (the age of the event when the app finally handles it), plus
the durations of `window_pump`, `[layer nextDrawable]` and `[cb
waitUntilCompleted]`.

Bend 2.0.5, macOS, app that draws a solid color and nothing else:

| | drag events (type 6) | mean age | max age |
| --- | --- | --- | --- |
| as shipped | 13 | 68.6 ms | 98.8 ms |

Frame pacing was fine throughout (no gap over 25 ms), and `[NSApp sendEvent:]`
itself took 0-7 ms, so the time is spent waiting for events to show up, not
handling them.

Giving the pump a small deadline fixes it. The deadline is taken once, at the
pump's start, so a dense event stream still ends the loop and the frame gets
drawn:

```diff
 static void window_pump(void) {
   @autoreleasepool {
+    NSDate* due = [NSDate dateWithTimeIntervalSinceNow:0.002];
     for (;;) {
       NSEvent* ev = [NSApp nextEventMatchingMask:NSEventMaskAny
-        untilDate:NSDate.distantPast inMode:NSDefaultRunLoopMode dequeue:YES];
+        untilDate:due inMode:NSDefaultRunLoopMode dequeue:YES];
       if (ev == nil) {
         break;
       }
       [NSApp sendEvent:ev];
```

With that patch, my real app (an ERD editor, a few hundred rectangles and glyphs
per frame) shows a mean age of 13.3 ms over 543 drag events, with 1787 frames at
a mean spacing of 11.1 ms. The two rows are not measured on the same program, so
they do not compare directly; I did not measure the shipped runtime on the
editor. The improvement is obvious to the eye either way.

Costs nothing that I can see: the frame waits for the display's next tick
anyway, and the call returns immediately whenever an event is already there.

Things I tried that made it worse, in case they come up:

| change | mean frame spacing | frames over 25 ms |
| --- | --- | --- |
| the 2 ms deadline above | 11.1 ms | 1% |
| 8 ms deadline | 18.0 ms | 24% |
| `maximumDrawableCount = 3` | 14.0 ms | 10% |
| `displaySyncEnabled = NO` | 22.9 ms | 41% |

One thing the patch does not address: the mouse-down that starts a title-bar
drag arrives with a mean age of ~40 ms, while the matching mouse-up arrives at
5.6 ms. Raising the deadline to 8 ms only moved that to 38 ms, so the pump
window is not the lever there. I have not worked out why the two differ.

`bend2/effs/window_frame.c` and `window_open.c` are byte-identical between
2.0.5 and 2.0.13, so this is still current.

### The file

```python
import Base

def tick(events: List<Event>, s: U32) -> IO(Maybe<U32>):
  match events:
    case Nil{}:
      IO.pure(Maybe<U32>, Some{s})
    case Con{Close{}, rest}:
      IO.pure(Maybe<U32>, None{})
    case Con{e, rest}:
      tick(rest, s)

def main() -> IO(Unit):
  App.run(~U32, ~App{+s => (s, Pix{s}), tick}, "drag me", 800, 600, 2105376)
```

### bend --version

bend 2.0.5

### uname -sm

Darwin arm64

Apple clang version 21.0.0 (clang-2100.1.1.101)

bythyag:
Use a fixed 2 ms deadline while pumping AppKit events so the run loop can receive newly arriving input without allowing a dense event stream to postpone frame rendering indefinitely.

Fixes #842.


VictorTaelin:
Thank you for taking this on, and sorry for the long answer — you deserve the measurements rather than a verdict.

We cannot take this one, for a reason that is not obvious from the report it fixes.

**The premise does not hold.** `nextEventMatchingMask:untilDate:distantPast` *does* run the run loop. A probe with a run-loop observer, a version-0 source, a due timer and a main-queue block sees all of them serviced by that call, in 30-43 µs, with 8 observer callbacks — the same as passing `[NSDate date]`. GLFW's `_glfwPollEventsCocoa` and SDL's `Cocoa_PumpEvents` poll exactly this way. So the pump is not starving the window server.

**The deadline sits on the frame's critical path.** `window_show` pumps before `[layer nextDrawable]`, so with your change an empty queue blocks the frame for up to 2 ms before the drawable is taken. Measured on a 120 Hz display, means over 120-frame windows:

| | main | this PR | branch `pump` |
|---|---|---|---|
| plain window: frame gap | 8.40-8.96 ms | 8.33-9.10 ms | 8.33-8.48 ms |
| plain window: run-loop residency | 0.1-0.4% | 24.6-28.5% | 96.6-99.4% |
| `demos/app_slash_boss_3d`: frame gap | 8.32-8.40 ms | **12.4-14.0 ms** | 8.33-8.40 ms |

On a solid-colour window the 2 ms hides inside the vsync slack. On a real app it overruns the 8.33 ms period, so each frame takes about one and a half of them: 120 FPS becomes 72-80.

**What your report gets right is the fault underneath.** The main thread does sit outside the run loop for ~95% of an idle frame — blocked in `nextDrawable`, not in the pump. Branch `pump` (commit 59fcc766) fixes that at no frame cost: the Metal sequence moves to a `QOS_CLASS_USER_INTERACTIVE` helper thread and the main thread waits *inside* the run loop, woken by `CFRunLoopStop` when the frame is presented. Idle run-loop residency goes from 1% to 97-99% with the frame gap unchanged.

So I am closing this in favour of that branch, not because the problem is imaginary. If you want to push it over the line, build from `pump` (`git clone -b pump https://github.com/bendlang/bend`) and tell us whether the lag is gone on your machine — that is the one measurement we cannot take here, since we cannot reproduce the symptom on an M4 Max at 120 Hz. If it fixes it, `pump` lands and your report is what got us there.

PR #854, opened 2026-09-19
Some misc compiler fixes

nicolas-abril:
Five emitter fixes from the fuzzer: a family's return through a cut, a foreign twin's seal by the constructor, a constant table only at a word type, a constructor built at its position, and a fork's held box

| commit | change | kind |
|---|---|---|
| 30e5b936 | a once-called def whose return is a box no longer fuses into a packed-return segment | fix |
| 7803b8e7 | a foreign twin seals a field by the constructor's own CID_HOT_T entry, no flag | fix, effects |
| 4dd12043 | a Nat match folds to a table only at a word type; the value filter goes | fix, simplification |
| 2fb67be7 | a constructor is built at the layout of its position, never reshaped | fix, simplification |
| 6f5ec3a5 | a shared flat box unboxes at its first share; a binding always carries its type | fix |

comp.ts: 61357 -> 61282 ttok (-75).

| bench | SEQ-CPU | PAR-CPU | PAR-GPU (Metal) |
|---|---:|---:|---:|
| bfs | +0.1% ± 0.1 | -0.7% ± 0.8 | +3.0% ± 7.1 |
| editdist | +1.0% ± 2.8 | +0.2% ± 1.1 | +0.3% ± 2.6 |
| gameoflife | +0.9% ± 1.6 | +2.2% ± 3.4 | -0.8% ± 2.0 |
| hashmap | +0.1% ± 0.8 | +0.5% ± 2.1 | -0.1% ± 0.5 |
| kmeans | -0.5% ± 0.5 | +0.1% ± 1.4 | -0.0% ± 0.4 |
| lexer | +0.7% ± 1.3 | +0.6% ± 1.1 | +1.1% ± 3.5 |
| mandelbrot | **-2.7% ± 1.1** | +0.1% ± 0.4 | -0.4% ± 1.5 |
| merkle | -1.4% ± 0.8 | +0.2% ± 0.5 | -0.3% ± 1.2 |
| nbody | +0.0% ± 0.2 | +0.2% ± 0.3 | -0.7% ± 2.3 |
| queens | +0.1% ± 0.2 | -0.0% ± 0.4 | +3.4% ± 6.2 |
| raytrace | -0.8% ± 1.2 | -0.5% ± 1.0 | +0.1% ± 1.9 |
| symreg | -0.7% ± 0.3 | +0.2% ± 0.7 | +0.2% ± 0.7 |
| terrain | +0.1% ± 0.3 | +0.0% ± 1.5 | -0.3% ± 1.8 |
| tree-bitonic | -0.1% ± 0.4 | +2.1% ± 3.4 | -2.2% ± 6.8 |
| tree-matmul | -0.2% ± 0.5 | +0.1% ± 1.3 | +7.6% ± 16.8 |
| tree-radix | -0.8% ± 0.3 | +0.5% ± 3.0 | +0.0% ± 0.4 |

hyperfine on the M4 against main at 18b2641a, 10 runs after a warmup per cell
(the branch is on this repo's main at 15ae0c86; every bench's C is still
byte-identical to main's but for the second commit's 30 runtime lines):
the delta of means with one sigma of the ratio, bold beyond two sigma and 2%;
PAR-CPU is 8 threads, PAR-GPU the Metal lane; every cell prints main's output.
No commit here touches a bench's C except the second, whose 30 lines are the
runtime's seal macros, off every bench's path, the numbers are noise.
mandelbrot SEQ came out past the threshold in this run and 2.3% ± 1.3 in a
run against b9afc634, on identical hot code: layout, not this branch.

Verified locally: the C lane over every test under tests/io, tests/run and
tests/show, the branch and main passing and failing the same set plus the
three new tests passing; the sixteen benchmarks' C byte-identical to main's
except the second commit's runtime change; the test gate run locally on the
rebased branch: every test passes but gfx_window on the C lane, a headless
expectation this Mac cannot meet, and flatten/literal_rows_cubic, whose C
hits clang's bracket-depth limit of 256 from main's own compiler as well,
with the branch's C for it differing only in the runtime lines; the 40 programs in the fuzzer's
repro corpus behaving as on main except the two this fixes; 3400 fuzzer
programs across the sequential, parallel and JS lanes. The cluster gates were
not run.

## 1. 30e5b936 A once-called def whose return is a box no longer fuses into a packed-return segment

A def called once is fused into its caller's segment. A def returning a
family at an open index returns a box, since the emitter cannot see the
index; fused into a caller typed at one index, whose return is that index's
packed layout, the body's arms were laid at the caller's layout, and an arm
of another index died with `a layout mismatch` or `a constructor outside its
layout`. Such a def now jumps, and its return converts through the cut like
any other call whose return disagrees with the segment's. The witness needs
the family's arm to be a call the folder cannot evaluate, and the caller's
argument likewise, else both fold away before emission.

```
import Base
def F(x: Bool) -> Data:
  match x:
    case False{}:
      U32
    case True{}:
      Maybe<&2, String>
def mf(n: Bool) -> F(n):
  match n:
    case False{}:
      7
    case True{}:
      Some{("a" ++ "b")}
def fn(c: Cmp) -> F(True{}):
  mf(True{})
def main() -> IO(Unit):
  IO.print(Maybe.default(&2, String, fn(U32.cmp(8, 11)), "none"))
```

- expected : ab
- observed : Error: a layout mismatch

## 2. 7803b8e7 A foreign twin seals a field by the constructor's own CID_HOT_T entry

A foreign def's C twin building a value of the program's own type had to
pass `io_node` a hot flag saying whether the program shares that type, a
fact only the emitter has. `IO_HOTS` covered six base constructors and said
0 for the program's own, so a shared tree from a twin arrived unsealed, and
once the program put it under a node of its own the shared take met a raw
field and fail-stopped. `io_node` and `io_box` now take no flag: they seal
by the constructor's entry in `CID_HOT_T`, the table the emitter writes for
its own constructors. This repo's `bend2/effs/file_read_at.c` passed the flag
too; the commit drops it there.

```
import Base
type T is Data:
  L{}
  N{a: T, b: T}
law mk:
  U32 -> IO(T)
def mk(d):
  import "./mk.c"
  import "./mk.js"
def sum(t: T) -> U32:
  match t:
    case L{}:
      1
    case N{a, b}:
      (sum(a) + sum(b) : U32)
def both(+t: T) -> U32:
  (sum(N{t, L{}}) + sum(t) : U32)
def main() -> IO(Unit):
  do IO<Unit>:
    t : T <- mk(2)
    IO.print(U32.show(both(t)))
```

with `mk.c` (on main, `io_node` takes a fifth argument, the flag, that the
twin can only pass as 0):

```
static Term mk_at(Env e, u32 d) {
  return d == 0 ? term_pak(CID_L, 0) : io_node(e, CID_N, mk_at(e, d - 1), mk_at(e, d - 1));
}
Term mk_run(Env e, Term* f, IoWork* w) {
  return mk_at(e, (u32)f[0]);
}
static void __attribute__((constructor)) mk_use(void) {
  io_eff(CID_MK, mk_run, 0);
}
```

and `mk.js`:

```
function mk(d) {
  return d === 0 ? { $: "L" } : { $: "N", a: mk(d - 1), b: mk(d - 1) };
}
```

- expected : 9
- observed : bend: runtime fail-stop

Test: `tests/io/foreign_shared`.

## 3. 4dd12043 A Nat match folds to a constant table only at a word type

A Nat match over constant arms folds to a table whose rows are evaluated in
JS and written as C words. Any one-word layout qualified: a Char or Cmp row
evaluated to a string or an object and died in `Failed to parse String to
BigInt`, and a Bool table folded only by the accident of `BigInt(true)` being
`1n`. The fold now asks the match's result type first and takes a table only
at U32, F32 or Nat, whose values are their C words. Main had meanwhile added
a filter dropping a non-number after evaluating the rows; the gate reaches no
such row, so the filter goes.

```
import Base
def ch(n: Nat) -> Char:
  match n:
    case 0n:
      'a'
    case 1n+p:
      'b'
def main() -> IO(Unit):
  IO.print(U32.show(Char.to_u32(ch(5n))))
```

- expected : 98
- observed : SyntaxError: Failed to parse String to BigInt (before main's filter; main compiles it, by the filter)

## 4. 2fb67be7 A constructor is built at the layout of its position

A constructor was emitted at its own type's layout and reshaped by its
consumer: a call's parameter, a node's field slot, a tail, a foreign
action's box. A constant reshaped through the arm chain had its wrong-arm
slices still marked constant, and the boxing shortcut then read a slice of
the wrong arm as a tag. A constructor is now built at the layout of its
position, passed down through `emit_expr` the way the type already is, so a
constant is never converted and its bit is read only where its node is
built. The shortcut, `node_build` and the bit's passage through slices go;
a value that already exists, a variable or a call's result, still converts
at its use. Five tests lose a chain on a literal tag in their C.

```
import Base
def g() -> Result<&1, &2, U32, U32>:
  do Result<&1, &2, U32, U32>:
    b : Bool <- Fail{7}
    return 0
def h(r: Result<&1, &2, U32, U32>) -> U32:
  match r:
    case Fail{e}:
      e
    case Done{u}:
      u
def main() -> IO(Unit):
  IO.print(U32.show(h(g())))
```

- expected : 7
- observed : TypeError: undefined is not an object (evaluating 'arm.k')

Test: `tests/run/ctr_at_position`.

## 5. 6f5ec3a5 A shared flat box unboxes at its first share, and a binding always carries its type

A box of a flat type, a family or polymorphic result, unboxes before it is
shared so its words copy instead of its node being kept. That ran at binding
time, gated on a nullable type. A fork emits its calls twice, in parallel
and in sequence, and the join needs both paths to hold the same values: the
sequential path rebinds a held cut through a frame whose hold entry inflates
the use count, so it unboxed a value the parallel path held boxed, dropped it
as a dead word, and the join died. The rule now lives in `bind_pop`, the one
place a value is duplicated, ahead of the keep it replaces; rebinding decides
nothing. A binding's type is no longer nullable: a let's value and a cut's
are annotated by the checker, so both die otherwise instead of silently
skipping their type facts.

```
import Base
def Fm(x: Bool) -> Data:
  match x:
    case False{}:
      Bool
    case True{}:
      U32
def mf(n: Bool) -> Fm(n):
  match n:
    case False{}:
      String.eq("a", "b")
    case True{}:
      7
def g(q: Bool, a: Fm(q)) -> U32:
  1
def r() -> U32:
  a b = mf(True{}) g(True{}, mf(True{}))
  (a + b : U32)
def main() -> IO(Unit):
  IO.print(U32.show(r()))
```

- expected : 8
- observed : Error: a fork's paths hold different values

Test: `tests/run/fork_held_family`.


nicolas-abril:
Superseded by #855, the same branch from the fork nicolas-abril/bend.

PR #856, opened 2026-09-19
--check checks a file and its imports without running main or building it

aldeni:
## What changed

`bend <file.bend> --check` reads the book and prints the check's verdict on stdout (`All terms check.`, or the unsafe count), and does nothing else: no run, no build. A broken file prints its report and exits 1; an open law prints `N TODOs found` and exits 1. It takes no other option (`-o`, `--checkup`, `--publish`, run arguments), as `--publish` does. `--help` and the guide's Tooling block name it.

Only `bend2/main.ts` (the CLI) and `guide/GUIDE.md` change; `bend.ts` is untouched.

## Why

`bend f.bend` checks the file and then runs `main`, so a file whose `main` serves a socket never gives the shell back. The one way to check such a file was `-o /tmp/x.c`, which also emits C and needs clang. A check-only mode is also what a proof gate wants: the verdict, and nothing running.

## How verified

`bun bend2/main.ts` at this commit, bend 2.0.16, Linux x86_64:

```
$ bend server.bend --check          # main loops on IO.sleep
All terms check, with 1 unsafe annotation.
[exit 0]

$ timeout 3 bend server.bend        # the plain run, for contrast
All terms check, with 1 unsafe annotation.
[exit 124]

$ bend broken.bend --check          # U32.add(1, "x")
Error:
- expected : U32
- observed : String
Location: main
[exit 1]

$ bend todo.bend --check            # an open law
Error: 1 TODO found.
The code is incomplete, and not a valid proof yet.
[exit 1]

$ bend value.bend --check           # main returns 42
All terms check.
[exit 0]
$ bend value.bend
42

$ bend value.bend --check -o x
bend: --check takes no other option (see bend --help)
$ bend value.bend --check arg
bend: arguments go to a run: bend <file.bend> [args] (see bend --help)
$ bend value.bend --publish --check
bend: --publish takes no other option (see bend --help)
$ bend page.html --check
bend: a page bundles with -o <dir> (see bend --help)

$ bend tests/check/alpha_equivalence.bend --check
All terms check.
$ bend value.bend -o v.c            # unchanged: still emits C
```

## Notes

- No test file: `tests/<ns>/*.bend` end in the `#|` lines their run must print, which exercises programs, not the CLI's options. If a gate has a place for a CLI probe, I will add one where you point.


VictorTaelin:
Shipped in 2.0.17 as `bend <file.bend> --check-only`. The name changed: `--check` reads as if plain `bend` did not check, and what this flag adds is that it runs nothing. Thank you.

PR #860, opened 2026-09-19
Structural Nat.min/max with direct order laws and split lemmas

nbardy:
Replaces the Bool.pick-over-computed-test Nat.min/max with structural recursion unfolding in lockstep with Nat.cmp (same values on all inputs: four constructor cases + induction; nothing in Base unfolds the old bodies).

Why: the pick form is proof-opaque (Bend cannot match on a computed test), so no order law over Nat.max/min was provable by case analysis. Structural bodies unlock Nat.cmp_refl/ge_refl/max_ge_l/max_ge_r, proved here adjacent to the defs.

Honest cost note: old form returns an existing node (O(1) allocation after O(min) compare); new form rebuilds O(min) Succ nodes. Provability outweighs the allocation on unary Nats.

Also adds Bool.split/Cmp.split decision lemmas (case analysis with the equation remembered) — the root fix for all 46 staging helpers.

Verification: full base.bend prints 'All terms check' under the source-built checker (bun bend2/main.ts). Prior bridge proofs over the old bodies (pick_succ machinery) become obsolete on this change; migration is one cong per arm.

Draft status: seeking maintainer read on direction before final polish.


VictorTaelin:
Shipped in 2.0.17. Your base was 80 commits behind, so the work was cherry-picked rather than merged, and trimmed to fit `base.bend`'s size budget: the two comment essays and the two unused lemmas (`Bool.split`, `Cmp.split`) came out. The structural `Nat.min`/`Nat.max` and the order laws are in as you wrote them; equivalence with the old definitions was checked on a 41x41 grid and at values past the literal bound. Thank you.

PR #862, opened 2026-09-19
Demo: bounded Bend core evaluator in Bend

luccahuguet:
## Why

The longer-term question is whether Bend can implement its own language frontend and eventually replace the parser and checker in `bend2/bend.ts`. The current TypeScript implementation could bootstrap that work. This PR tests one narrow premise: can Bend express part of its own reduction and conversion semantics and agree with `bend.ts` on representative cases? It changes no production path and is not a self-hosted parser or type checker.

## What

- Model six first-order term forms (`Var`, `Lam`, `App`, `Let`, `Ref`, `Ann`) with de Bruijn indices.
- Implement fuel-bounded weak-head reduction and `EQ` comparison. Fuel exhaustion is an inconclusive result, not inequality.
- State and prove the annotation, let, and beta reduction equations for arbitrary terms, books, and fuel.
- Compare the model against this branch's `bend2/bend.ts`: 16 structural reduction cases, 2 sharing cases, 24 equality cases, and fuel exhaustion. The sharing cases check the same head/spine and convertibility because the TypeScript evaluator memoizes variable cells while this model substitutes terms.

## Run

```sh
bun bend2/main.ts demos/bend_core/main.bend
bun bend2/main.ts demos/bend_core/PROOF.bend
bun demos/bend_core/parity.ts
```

Locally, the demo prints `["Equal","Different","Exhausted"]` through the interpreter, compiled JS, and native C. The proof reports `All terms check.` and the parity script passes. `bun gates/repo.ts` reports `PASS: 46 / 46`. I did not run the cluster test gate; this demo has no file under `tests/`.

## Limits and feedback sought

The fragment omits types, quantities, ADTs, matching, rewriting, and source parsing. The three laws pin the modeled reduction steps; they do not prove parity with `bend.ts` beyond the executable cases.

Is a Bend-written frontend a direction worth pursuing? If so, would this model be useful to keep as a demo, or would the next contribution be more valuable if it replaces one real `bend.ts` path first?


VictorTaelin:
This is impressive, genuinely: a bounded evaluator for Bend's core written in Bend, and it checks. Thank you for building it and for showing it to us.

We are already working on Bend-in-Bend internally, with some important changes and improvements to the core that will land in the language first, so a demo of the current core would not stay in step with it and we cannot give it a home in this repo. bendhub is a good place for it, where people can run it and learn from it today. Closing the PR, with admiration for the work.

luccahuguet:
Thanks! I will restrict my work to more mergeable (and less architectural) work

To avoid taking too much of our time

While still moving the needle forward

Very excited about bend-to-bend

PR #863, opened 2026-09-19
A nat literal in a pattern checks its constructors (#852)

### LINKED ISSUE #852: A nat-literal pattern skips the constructor arity check, so a non-decreasing self-call passes and a closed term inhabits Empty
### What you did

bend EMPTY.bend

### What happened

```text
All terms check.
```

### The file

```python
type U is Data:
  Un{}

type N is Data:
  Zero{w: U}
  Succ{p: N}

type D is Data:
  K{z: N}

type Empty is Data:

def f(b: N) -> Empty:
  match b:
    case U32.to_nat(0):
      {\{K: q => w => f(q)} : D -> U -> Empty}(K{Zero{Un{}}})
    case Succ{p}:
      f(p)

def boom() -> Empty:
  f(Zero{Un{}})
```

### bend --version

bend 2.0.16

### uname -sm

Linux x86_64

Ubuntu clang version 21.1.8 (6ubuntu1)

PedroVIOliv:
Fixes #852.

`parse_patt` has a fast path for nat literals that builds the pattern and returns before the constructor case, so neither the constructor lookup nor the arity check runs on it. A `Zero` that carries a field then matches as if it were nullary, the field becomes a binder on the branch, and a non-decreasing self-call slips past the termination check — leaving a closed term in `Empty` under `All terms check.`

### Where it came from

9e362398, "A Nat literal past 256n parses as U32.to_nat(n), up to the u32 bound", shipped in 2.0.7. Before it every literal expanded into `Ctr` terms in `parse_term_num` and so reached the `case "Ctr"` branch, which does both checks. That commit stopped large literals expanding as terms; patterns still expand, so it added the fast path, and the checks went with it.

```
2.0.6 (9e362398^)   case 257n   ->  "a Zero pattern with 1 field"
2.0.16              case 257n   ->  accepted, branch goal becomes @w:U -> U
```

### Scope

Only literals above `NAT_LITERAL_MAX` (256) take the unchecked path; smaller ones still parse as `Ctr` and were never affected. The `U32.to_nat(0)` in the report is the same path written longhand.

### The fix

Both paths now go through one `parse_patt_ctr`, so the literal form reports what the braced form reports and the two cannot drift apart again. `Zero` is checked always and `Succ` only when the literal is non-zero, which is exactly what the pre-2.0.7 expansion checked.

Literal patterns resolve `Zero`/`Succ` by name, so a user's own Peano-like type gets them too, and `tests/grade/*.bend` rely on the same resolution for their locally declared `Nat`. This keeps that: it adds the arity check and nothing else.

### Tests

- `tests/parse/nat_literal_pattern_arity.bend` — the rejection, at 257n
- `tests/parse/nat_literal_pattern_big.bend` — a literal past 256 still matches on a real `Nat`

There was no coverage above 256 before; every literal pattern under `tests/` is `case 0n:`.

### Verification

I can't run `gates/test.ts` or `gates/perf.ts` — they ssh into the mini cluster. Locally I ran:

- every `.bend` under `tests/` through the checker, compared against its `#|` block: 1311 match, 5 differ. Those 5 differ identically on an unmodified tree (my runner does the check lane only, not JS or C), so they are my harness and not this change.
- `tsc --noEmit`: clean for `bend2/`. The two `canvas` errors in `docs/gen_*.ts` are pre-existing and unrelated.
- before and after on the file from #852 and on `case 257n`.

Glad to adjust if the gates disagree.

---

*This patch was written by Claude (Claude Code) working with me. I read it, ran the checks above, and am submitting it; if anything here is wrong, say so and I will fix it.*


VictorTaelin:
Thank you. #852 is fixed in 2.0.17, at the same place your patch points at but one step earlier: `parse_patt`'s literal branch now builds a `Ctr` term and falls through to the `Ctr` case, so a literal takes the same arity check as every other pattern instead of a check of its own. That also made the second arity check in `match_flatten` unreachable, and it is deleted. Closing as fixed.

PR #865, opened 2026-09-19
Expand to a compiler-backed Bend 2 LSP

don2e4:
## Summary

Rename the unreleased formatter package and executable from `bend2-fmt-lsp` to `bend2-lsp`, retain formatting, and add compiler-backed diagnostics and Markdown hover support.

## Implementation

- bundle the trusted `bend2/bend.ts` compiler and `base.bend` with esbuild without modifying compiler sources
- run compiler analysis in a persistent, recoverable worker thread
- resolve unsaved open-document overlays before workspace files
- support Base and hash-addressed `BEND_LIB` packages, including verified `BEND_HUB` downloads with a 10-second timeout
- publish diagnostics on open and after a 250 ms change debounce
- suppress stale document and imported-overlay results, clear diagnostics on close, and recheck open dependents
- categorize parse, type-check, import, hole, and incomplete-law diagnostics with compiler source ranges
- retain independent lexical recovery for incomplete literals, delimiters, and `?TODO`
- provide hover for declarations, laws, datatypes, constructors, imports, keywords, operators, quantities, proofs, and GPU calls
- keep compiler analysis limited to `file:` documents; untitled buffers retain formatting, lexical diagnostics, and syntax hover

Local-variable type hover remains out of scope because it would require trusted-checker instrumentation.

## Public interface

- package and executable: `bend2-lsp`
- invocation: `bend2-lsp --stdio`
- language IDs: `bend` and `bend2`
- capabilities: full document sync, formatting, push diagnostics, and hover

## Validation

- `npm test`: 18 tests passed
- exact diagnostic categories and UTF-16 source offsets covered
- live debounce, stale suppression, close clearing, overlays, imported hover, and worker recovery covered
- formatter idempotence checked across the Bend test corpus
- `npm pack --json`: server, bundled compiler worker, and `base.bend` present
- `git diff --check` passed

The Bun-based repository gate could not run in the validation environment because Bun is not installed.


kbrianps:
While the compiler-backed server is not there, a cheap approximation is possible, if it is wanted.

The VS Code extension in #879 has a branch that runs `bend <file> -o <tmp>.js` on a save and marks the line the error names, with the reason, the context and the def as the message. `-o` emits instead of running, so a `main` with effects stays put, and a file with no `main` answers "no main to run" once it has checked; a file that leaves goals open marks each `?TODO` instead.

It is not an LSP and does not pretend to be one: bend stops at its first error, so one error shows at a time, and it reads the file on disk, so an unsaved buffer keeps the last verdict. No hover, no completion. The part that reads what bend prints is a plain string function with tests, so it is easy to throw away once the server reports diagnostics itself.

Branch: https://github.com/kbrianps/bend/tree/vscode-check

Happy to send it as a follow-up PR, or to drop it if the compiler-backed server should be the only thing that reports errors.

don2e4:
drop it if #865 lands, but keep it for now

VictorTaelin:
Thank you for this. A compiler-backed LSP is a substantial thing to have built, and the approach is sound.

We keep the compiler repo to the compiler: one test gate, one complexity budget, nothing that has its own release cycle. Please publish the LSP as its own repository; we would gladly link it from our README or guide so people find it, and we are open to exposing whatever compiler entry points it needs. Closing the PR, with gratitude for the work.

don2e4:
> Thank you for this. A compiler-backed LSP is a substantial thing to have built, and the approach is sound.
> 
> We keep the compiler repo to the compiler: one test gate, one complexity budget, nothing that has its own release cycle. Please publish the LSP as its own repository; we would gladly link it from our README or guide so people find it, and we are open to exposing whatever compiler entry points it needs. Closing the PR, with gratitude for the work.
> 
> *Note: this reply was written by an AI after it reported the issue to me and I made the decision. If anything here is wrong, reply and I will review it myself.*
> 

Hi im gonna do this now

Issue #809, opened 2026-09-18
C and JS lanes decode invalid UTF-8 differently: io_str accepts overlong, surrogate and out-of-range sequences (File.read, get_env, TCP.recv, UDP.recv)


JGalego:
### What you did

```
bend u.bend          # JS lane
bend u.bend -o u && ./u   # C lane
```

### What happened

Every effect that turns OS bytes into a `String` uses a different UTF-8 decoder on
each lane, so the same program on the same input produces two different `String`s.

The C kit decodes with `io_str` (`bend2/comp.ts:5357`), a hand-rolled backward
scanner with no validation. The JS kit decodes with `io_text` (`bend2/comp.ts:6114`),
i.e. `TextDecoder`, which is WHATWG-strict and substitutes `U+FFFD`.

`io_str` is shared by `File.read`, `IO.get_env`, `TCP.recv`, `UDP.recv_from` and
`UDP.poll` — every untrusted-input seam, including the network ones.

Reading these 23 bytes

```
41 7c 80 7c c0 80 7c ed a0 80 7c f7 bf bf bf 7c ff 7c e6 97 7c c3 a9
```

and printing `Char.to_u32` of each char:

```
JS: 65 124 65533 124 65533 65533 124 65533 65533 65533 124 65533 65533 65533 65533 124 65533 124 65533 124 233
C:  65 124 128   124 0           124 55296            124 2097151              124 255   124 230 151   124 233
```

| input bytes | what it is | C lane | JS lane |
|---|---|---|---|
| `80` | stray continuation byte | `Chr{128}` (read as Latin-1) | `Chr{65533}` |
| `c0 80` | **overlong NUL** | `Chr{0}` | `Chr{65533}` ×2 |
| `ed a0 80` | **surrogate U+D800** | `Chr{55296}` | `Chr{65533}` ×3 |
| `f7 bf bf bf` | **U+1FFFFF, above U+10FFFF** | `Chr{2097151}` | `Chr{65533}` ×4 |
| `ff` | invalid | `Chr{255}` | `Chr{65533}` |
| `e6 97` | truncated 3-byte | `Chr{230} Chr{151}` | `Chr{65533}` |

`String.length` of the same file is **16** in C and **21** in JS. Neither lane
reports an error; both exit 0.

`IO.get_env` shows it too, so it is not specific to `File.read`. With
`BADUTF` set to `41 ed a0 80 c0 80`:

```
JS: 65 65533 65533 65533 65533 65533
[...]

PR #873, opened 2026-09-19
Strings are views over adaptive-width buffers; UTF-8 streams in chunks

phenomenon0:
A String was a list of Char cells, so length, indexing, slicing and compare walked it, and 8 MiB of text cost a list of 8 M nodes. The C runtime now lays a String as a two-word view (buffer, offset and length) over cells 1, 2 or 4 bytes wide, the narrowest that fits its code points. Slices share their buffer, and length, compare, search, split, replace, trim, case and padding run natively in C and JS. The String API keeps its types.

Utf8.Dec decodes bytes in chunks with a carry across the cut. Leads are sync points, so any chunking answers the whole decode (WHATWG maximal subparts, as in #809). File.read_text and TCP.recv_text stream through it, and File.fold_text folds a step over a file's chunks and closes it.

- C decode of 8 MiB with File.read + String.length: ASCII 128 -> 25 ms, Unicode 87 -> 28 ms. Peak RSS 198,600 -> 18,512 KiB (ASCII). JS is at parity.
- tests/strings: 21 tests, each passing check, interp, JS and C. All tests: 1335 / 1335 (1314 upstream plus these 21).
- Out of tree: an ASan+UBSan probe of the string runtime. 84 M allocations end with zero live; the streaming partition law holds on 177 k C and 180 k JS partitions; 83 injected allocation failures.

**How it compares.** Same work - read the whole 8 MiB UTF-8 file and answer its character count - on a busy box (load ~6 of 16), median of 7, startup subtracted, all counts agree (8,388,608 / 4,934,470):

| | ASCII | Unicode |
|---|---|---|
| Rust 1.92 (read_to_string + chars().count()) | 4.0 ms / 10 MB peak | 7.8 ms / 10 MB |
| Python 3.11 (open().read() + len()) | 8.0 ms / 31 MB | 16.3 ms / 43 MB |
| Go 1.23 (ReadFile + RuneCountInString) | 10.4 ms / 18 MB | 11.7 ms / 19 MB |
| Node 22 (readFileSync utf8 + count) | 11.3 ms / 62 MB | 56.8 ms / 64 MB |
| Bend (this branch, C lane) | 26.4 ms / 18 MB | 27.9 ms / 30 MB |

Rust's count is a validation-only walk and Node's ASCII count is O(1) on a one-byte string; Bend materializes every code point into a fitted buffer and answers from the view. Among the dynamic runtimes Bend is the memory-leanest, Node's Unicode decode costs 2x the time at 2x the RAM, and the whole-text family (search, split, replace, compare, slices) then runs over the view with no extra buffers.

Review notes:
- gates/repo.ts lifts comp.ts from 62000 to 75000 ttok (74112 now).
- List and Char join RUNTIME_ADTS, since the runtime builds Nil/Con/Chr cells. A user constructor that mangles to CID_NIL or CID_CON now clashes, as one mangling to CID_CHR already does.
- Non-scalar Chars behave as upstream: JS rejects them at char_new, and interp and C keep the cell.

🤖 Generated with Claude Fable 5.1


VictorTaelin:
Thank you for this, and we want to be clear about how we see it: strings as views over adaptive-width buffers with streamed UTF-8 is important, and it is the direction String should go. This is not a "no" to the idea.

We cannot take the PR: it is about 16,000 tokens across `comp.ts` and `base.bend`, far past the complexity budget those files live under, and it changes the representation of String that every backend, effect and test currently agrees on. Taking it would mean owning that representation change across the whole runtime at once. What we will do is try to bring this in within the complexity cap in the future, in smaller steps, and your design is the reference for that. Closing the PR, with thanks for the work and the ambition.

PR #845, opened 2026-09-19
Add a formatting-only Bend 2 language server

don2e4:
Hi, i do something.

Bend 2 currently has no formatter or editor integration. This adds a standalone, formatting-only language server that preserves Bend's whitespace-sensitive syntax while normalizing indentation and safe token spacing.

The server supports `textDocument/formatting` over stdio for `bend` and `bend2` documents. It intentionally does not advertise diagnostics, completion, hover, range formatting, or on-type formatting.

Validation:

- `npm test` (12 unit and stdio integration tests)
- Parser compatibility check across 1,165 valid Bend files
- `npm pack --dry-run`
- `git diff --check`

PR #879, opened 2026-09-19
Add a Bend 2 VS Code extension: the grammar and a formatter client

kbrianps:
`tools/bend-vscode` colors Bend 2 in VS Code and formats it through `bend2-fmt-lsp` (#845), which has no VS Code setup of its own yet.

**The grammar** carries the scopes of `bend2/docs/bend.sublime-syntax` one by one, so a `.bend` file in the editor and a code block in the papers are colored the same way. It parts from it in three places, all additions: `~` is an operator (the template sigil), `.|.`, `.^.` and `.&.` are one operator each rather than a dot beside a bit symbol, and a `0x` hash (a package import) is a number. If you would rather keep one source of truth, I can generate the tmLanguage from the sublime syntax at build time instead.

**The client** starts the formatting server, looking for it in the `bend.formatter.path` setting, then the sibling `tools/bend-fmt-lsp/dist/server.js`, then `bend2-fmt-lsp` on the PATH. With none of them the extension still colors, and says so in its `Bend` output channel. It adds no diagnostics: `bend <file>.bend` is the checker.

The layout follows `tools/bend-fmt-lsp` (package.json, tsconfig.json, src/, README), and `gates/repo.ts` carries the three allow lines: `bun gates/repo.ts` is PASS: 49 / 49 here.

Checked: the grammar tokenized with vscode-textmate over `bend2/base.bend` and over a file covering the syntax reference (`@unsafe`, templates, rewrites, holes, arrays, floats, escapes, `Nat` literals); formatting exercised over stdio against the server in this repo.

Publishing to the Marketplace is yours to decide: this is source only, and the `bendlang` publisher does not exist yet.


don2e4:
drop if #865 gets merged but keep it for now

VictorTaelin:
Thank you for building this. A grammar and a formatter client are exactly what editor users ask for, and the work is good.

We are not going to keep editor extensions inside the compiler repo: everything here is under one test gate and one complexity budget, and an extension has its own release cycle and its own users. Please publish it as its own repository (and on the VS Code marketplace); we would be happy to link it from our README or guide so people find it. Closing the PR, with real thanks.

kbrianps:
Thanks, that makes sense. It's up as its own repo now: https://github.com/kbrianps/bend-vscode

v0.1.0 has a .vsix under Releases, and a Marketplace listing will follow. It ships don2e4's bend2-lsp inside, so VS Code users get the grammar plus errors as they type, hover and formatting, with nothing else to install.

If the offer to link it from the README or the guide still stands, that would be great.

Issue #846, opened 2026-09-19
Would bend have built-in std libs for Http/Tls/Crypto?


H4ad:
### What Bend should do

```python
// this is a demo from https://github.com/H4ad/bend-stdlib/blob/master/demo/api.bend

import Base
import ./vendor/json.bend as JSON
import ./vendor/http/http.bend as HTTP
import ./vendor/http/router.bend as Router
import ./vendor/http/server.bend as Server

def hello(ctx: Router.Ctx) -> IO(HTTP.Response):
  Router.text(200, "bend-stdlib http api")

def health(ctx: Router.Ctx) -> IO(HTTP.Response):
  Router.json(200, JSON.Obj{[JSON.Member{"status", JSON.Str{"ok"}}]})

def greet(ctx: Router.Ctx) -> IO(HTTP.Response):
  Router.text(200, "hello, " ++ Router.param("name", "world", ctx))

def u32_or.read(mb: Maybe<&2, U32>) -> U32:
  match mb:
    case Some{+n}:
      n
    case None{}:
      0

def u32_or(s: String) -> U32:
  u32_or.read(U32.read(s))

def sum(ctx: Router.Ctx) -> IO(HTTP.Response):
  +ps = Router.query(ctx)
  a = u32_or(Router.param.list("a", "0", ps))
  b = u32_or(Router.param.list("b", "0", ps))
  Router.json(200, JSON.Obj{[JSON.Member{"sum", JSON.Num{U32.show(U32.add(a, b))}}]})

def echo(j: JSON.Json, ctx: Router.Ctx) -> IO(HTTP.Response):
  Router.json(200, j)

def powered(resp: HTTP.Response) -> HTTP.Response:
  Router.header("X-Powered-By", "bend-stdlib", resp)

def routes(ctx: Router.Ctx) -> IO(HTTP.Response):
  Router.dispatch([
    Router.GET("/", hello),
    Router.GET("/health", health),
    Router.GET("/hello/:name", greet),
    Router.GET("/sum", sum),
    Router.POST_JSON("/echo", echo)
  ], ctx)

def app(ctx: Router.Ctx) -> IO(HTTP.Response):
  Router.map(~powered, routes)(ctx)

def main() -> IO(Unit):
  Server.listen(~app, 8080)
```

### Why

I'm doing a experiment to create a HTTP Framework for bend, and I'm missing few features in the standard lib:

| patch | effect it adds | used by | why Bend cannot |
|---|---|---|---|
[...]

PR #890, opened 2026-09-20
DNS.resolve: getaddrinfo for one IPv4 dotted quad

angerman:
## Summary
- Adds `DNS.resolve(host) -> IO(Result<…, String>)`: one IPv4 dotted quad via `getaddrinfo` (C, on a helper thread) / the same call through `bun:ffi` (JS).
- Numeric hosts keep the shared no-leading-zero `io_sys_addr` / `io_addr` rule; digit-and-dot strings that fail that rule return `EINVAL` and never fall through to `getaddrinfo`.
- `TCP.connect` / `UDP.send_to` stay numeric-only — compose with `IO.try` + `DNS.resolve`.
- Test `tests/io/dns_resolve.bend` covers identity, leading-zero reject, `localhost` (hosts file), and `no-such-host.invalid`.
- `demos/io_http_fetch` uses `DNS.resolve("ip-api.com")` instead of a hardcoded IP.

Addresses the hostname gap called out in #846 (without baking resolution into `TCP.connect`).

## Test plan
- [x] `bun bend2/main.ts tests/io/dns_resolve.bend` (interpret)
- [x] `bun bend2/main.ts tests/io/dns_resolve.bend -o t.js && bun t.js`
- [x] `bun bend2/main.ts tests/io/dns_resolve.bend -o t && ./t`
- [ ] Cluster `gates/test.ts` shard including `io_dns_resolve`


VictorTaelin:
Thank you for this, really. The implementation is careful (a helper thread for `getaddrinfo`, the `bun:ffi` twin, a test and a demo), and DNS is something Bend will want.

We cannot merge it right now for one reason: `base.bend` is at its complexity cap, and each addition to it has to be paid for by a removal. Rather than hold your work hostage to that, the best home for it today is bendhub, where it can be published and used as a library at once. We will consider bringing DNS into Base in the future when there is room. Closing the PR with gratitude, and please do publish it.

angerman:
Published on BendHub as requested (Base complexity cap):

```
0xf45a7ca0255c5a863fba26e24971a8e0
import 0xf45a7ca0255c5a863fba26e24971a8e0/dns.bend as DNS
```

Same getaddrinfo implementation as this PR (C helper thread + bun:ffi JS, canonical-quad / soft-numeric guards). Compose with `IO.try` before `TCP.connect`:

```python
import Base
import 0xf45a7ca0255c5a863fba26e24971a8e0/dns.bend as DNS

def main() -> IO(Unit):
  do IO<Unit>:
    ip : String <- IO.try(String, DNS.resolve("example.com"))
    c  : Socket <- IO.try(Socket, TCP.connect(ip, 80))
    …
```

Package sources live at `pkgs/dns/` on the `dns-resolve` branch of the fork.

PR #894, opened 2026-09-20
README: community libraries link

nbardy:
Adds a Libraries line to the Community section pointing at bend_stdlib (proof-carrying community standard library: order theory, containers, canvas/numerics roadmap; 59 checked laws, all green). One line, additive, matches the existing list format.


VictorTaelin:
Thanks for the link, and congratulations on `bend_stdlib`. We are not adding links to the README for now; it is kept to what ships in this repo. Publishing on bendhub is the way to make a library findable from inside Bend, and we would be glad to see it there. Closing.

PR #898, opened 2026-09-20
docs: add REPORTING.md, a pre-issue checklist

umag:
Adds `REPORTING.md`: the work to do before filling a `.github/ISSUE_TEMPLATE`
form. Bend has the forms but not the procedure around them, so reports arrive
without a minimal repro, without the environment fields, or as things
`WONTFIX.txt` already answers.

The doc is one page and agent-agnostic — a person or an agent reaching it from
`AGENTS.md` follows the same eight steps: classify it as Bend's bug (not your
program) and re-decide each time, read `WONTFIX.txt`, search the tracker for
dupes, reduce to the smallest `.bend`, collect the exact `bug.yml` /
`feature.yml` fields, sanitize paths and secrets, file it (form or one
`gh issue create`), and follow up as a comment rather than a second issue.

Wiring:
- `AGENTS.md` — one map entry pointing at it (676/2000 ttok).
- `gates/repo.ts` — `allow("REPORTING.md", 900)` so it passes the shape gate
  (the file is 746 ttok).

`bun gates/repo.ts` passes 47/47.


VictorTaelin:
Thank you for taking the time to write this. We are going to pass: the issue forms already ask for the environment fields and a minimal program, and `WONTFIX.txt` is linked from them, so a second document would mostly restate the form. We keep the repo's prose deliberately small so that every file in it is one we maintain. Closing with thanks.

umag:
Idea is here to provide an agent a way to find that docs and report issues according to them.

PR #903, opened 2026-09-20
Fix operator namespacing for parent-relative imports

hg7643:
Parent-relative imports can fail inside explicitly typed arithmetic because `parse_term_ns` treats every reference beginning with `.` as an unresolved operator. For example, a recursive call resolved to `../import/op_parent_lib.sum` becomes the undefined name `Nat../import/op_parent_lib.sum`.

Restrict operator namespacing to a leading dot followed by a non-dot, matching the existing distinction in `term_higher`. Imported ordinary calls then retain their resolved names and the existing early return, so an outer annotation still cannot annotate operators inside their arguments.

The regression fixtures cover `../` and `../../` imports, recursion within an imported module, and rejection of an unannotated operator inside an imported ordinary call. The positive import fixtures fail on unmodified main (`7561656`, also present in 2.0.20).

Validation on macOS arm64:

- New fixtures match the upstream gate's exact `#|` expectations, including aggregate `--checkup`.
- Three positive fixtures pass native one/four-worker and JavaScript execution from both the checkout root and their namespace directories (18 runs).
- The imported-call negative fixture still reports the expected missing operator annotation.
- Six existing operator/annotation-boundary fixtures pass unchanged.

The distributed mini-cluster gate was not run locally. No CLI version or runtime changes are included.


VictorTaelin:
Merged in spirit and shipped in 2.0.22, with credit to you: your line was at the cause, and the release takes the rule one step further. An operator placeholder is the name whose only dot leads it (`.add`, `.is_lt`), so both sites now test `lastIndexOf(".") === 0`, which also covers a module in a dot directory (`import ./.hid/lib.bend`). One test pins the `../` case (`tests/parse/op_parent_import.bend`) in place of the four fixtures. Closing the PR; thank you.

Issue #892, opened 2026-09-20
Three string-literal match arms exceed 512 MiB during C generation on 2.0.19


gouveags:
### What you did

I hit compiler memory pressure while building an OpenAI SDK for Bend. I reduced one part of it to three string-literal cases on Bend 2.0.19. This looks related to #867, but uses string patterns rather than the dense U32 tables fixed there.

Compile the file below to C. On Linux I ran the reproduction in a memory-limited scope so it could not take down the desktop:

```sh
systemd-run --user --scope -p MemoryMax=512M -p MemorySwapMax=0 \
  timeout 30s bend literal.bend -o literal.c
```

### What happened

The compiler exceeded the 512 MiB cgroup limit and was killed before emitting C. The kernel recorded a memory-cgroup OOM with 521956 KiB anonymous RSS. This is a bounded reproduction, not a measurement of how much memory an unrestricted run would eventually need.

Replacing the string patterns with `String.eq` and a match on three Bools compiled successfully under the same limit: 1.06 s, 136048 KiB maximum RSS, approximately 97 KiB of C.

In the larger SDK I also observed multi-gigabyte compiler processes and a host OOM while two builds overlapped. I have stopped overlapping builds and now use memory limits. The SDK workaround is to use enums/comparisons for protocol dispatch. I am not claiming the three-case example alone caused the host OOM.

Is there a small emitter-side optimization for these literal string matches, or should this remain a documented limitation? I avoided touching the language/checker.

### The file

```bend
import Base

def classify(s: String) -> U32:
  match s:
    case "response.completed": 1
    case "response.failed": 2
    case "response.incomplete": 3
    case _: 0

def main() -> IO(Unit):
  do IO<Unit>:
    s : String <- IO.try(String, IO.get_env("BEND_REPRO_INPUT"))
    IO.print(U32.show(classify(s)))
```

Comparison version:

```bend
import Base

def classify_flags(a: Bool, b: Bool, c: Bool) -> U32:
  match a b c:
    case True{} _ _: 1
    case _ True{} _: 2
    case _ _ True{}: 3
    case _ _ _: 0

def classify(+s: String) -> U32:
  classify_flags(String.eq(s, "response.completed"), String.eq(s, "response.failed"), String.eq(s, "response.incomplete"))

def main() -> IO(Unit):
  do IO<Unit>:
    s : String <- IO.try(String, IO.get_env("BEND_REPRO_INPUT"))
    IO.print(U32.show(classify(s)))
```

[...]

Issue #791, opened 2026-09-18
Checker recursion depth: valid programs with ~10k-deep terms (e.g. a 10000n literal) fail with "machine stack overflowed"


docxology:
## What you did

```python
# lit.bend
import Base

def main() -> Nat:
  10000n
```

`bend lit.bend` (same via the bun/node import hook).

## What happened

```
Error: the machine stack overflowed (a deep recursion, or a literal too large to expand)
```

- `5000n` checks and runs fine; `10000n` (and everything larger up to the advertised `Number.MAX_SAFE_INTEGER` bound) dies.
- The same overflow hits valid non-literal programs: a single def body with ~4,000 sequential do-block statements fails the same way while the same program at 2,000 statements compiles (in 73s — see the separate quadratic-compile report).

So the practical term-depth budget of the checker is ~10k constructor nodes / call frames, well below any documented limit, and the `MAX_SAFE_INTEGER` guard in `parse_term_num` (bend.ts:2234) is decorative: nothing between ~10k and 9e15 can ever expand.

The failure itself is handled (caught RangeError with a decent message), so this is a robustness/usability issue, not a crash: valid programs fail with a message that doesn't say which literal or construct was too deep.

## Expected

- Make the checker's recursion depth budget explicit and iterative where cheap (the Nat literal expansion loop itself is iterative; the overflow happens downstream in checking/normalization).
- At minimum, catch the overflow at the offending span and report "this literal/term is too large (limit ~N)".

## Version / system

- bend 2.0.5 (main @ e6676b0)
- Darwin arm64

VictorTaelin:
Duplicate of #779 (same recursion: `term_check` <-> `tele_check` over the Succ chain of a Nat literal, same host-stack limit). The literal cap and the message are being decided there. Closing this one to keep the discussion in one place.

*Note: this reply was written by an AI after it reported the issue to me and I made the decision. If anything here is wrong, reply and I will review it myself.*

Issue #779, opened 2026-09-18
Type checking a large Nat literal overflows the host stack without identifying the literal


iamjoaovytor:
### What you did

bend main.bend

### What happened

```text
Error: the machine stack overflowed (a deep recursion, or a literal too large to expand)

The source contains a single Nat literal and no recursive function.

On my environment (Bend 2.0.5, Bun 1.3.3, macOS 27.0 build 26A428, Apple M4 Pro, arm64):
- 5314n succeeds and prints 5314n.
- 5315n fails with the error above.
I reproduced both cases with bend main.bend and with bend main.bend -o main. The successful native executable also prints 5314n.

These are observed thresholds in this environment, not a documented language limit.

Cause investigated at commit e6676b0:
The parser already expands Nat literals iteratively into Succ constructors (bend2/bend.ts:2248-2250). Calling book_load and book_valid directly shows that loading succeeds and the RangeError occurs in book_valid, with this repeating stack:

RangeError: Maximum call stack size exceeded.
    at term_check (bend2/bend.ts:3480:21)
    at tele_check (bend2/bend.ts:1111:19)
    at term_check (bend2/bend.ts:3497:26)
    at tele_check (bend2/bend.ts:1111:19)
    ...

The CLI catches RangeError in bend2/main.ts:469-474 and replaces it with the generic message above. That message identifies neither the literal nor its source location. In a larger program, it is difficult to distinguish this from recursion in user code.

Expected:
- Avoid exhausting the host stack while checking the expanded literal, or reject unsupported depth with a clear diagnostic.
- Identify the offending literal and its source location when possible.
- If an intentional depth limit is introduced, document it and report it explicitly; 5314 should not be treated as a universal cap.
```

### The file

```python
import Base

def main() -> Nat:
  5315n
```

### bend --version

bend 2.0.5

### uname -sm

Darwin arm64

### clang --version (the first line)

Apple clang version 21.0.0 (clang-2100.3.34.2)

phenomenon0:
Extra data points on this class, from a fork experiment (main `0b7e2b11`; also reproduced via the 2.0.5 CLI):

- Minimal case: `def main() -> Nat: 4096n`.
  - **Non-deterministic on one machine**: idle, it frequently fails in <1 s with `RangeError: Maximum call stack size exceeded` (0–2/5 pass across several series); with other CPU cores busy, the identical command passes (3/3 recorded).
[...]

PR #907, opened 2026-09-20
Fix/string literal term

MattCozendey:
# A string literal checks as one node

Pull request for bendlang/bend from `fix/string-literal-term`.

## What changes

A string literal now parses as one term node, `Lit`, that holds its text.
Its type is `String`. It means the same `SCon` chain as before, and the
checker unfolds it one character at a time where a chain is read:

- a match frame in `term_wnf` reads `"ab"` as `SCon{'a', "b"}`;
- `term_compare` compares two literals by text, and unfolds a literal that
  meets a constructor, so `{"ab" == SCon{'a', "b"} : String}` holds by `{==}`;
- `term_descend` unfolds a literal argument against a constructor column;
- a literal pattern (`case "one":`) is its chain's pattern, as before;
- `term_check` accepts a literal at `String` when its head constructor is not
  peeled off the type. For any other type it checks the first unfolding, so
  the error message is the one the constructor gave.

Before this change the parser turned `"text"` into
`SCon{Chr{U32{WCon{...}}}, ...}`, about 70 term nodes per character. The
checker lifted, copied, checked and kept that chain. A literal cost about
40 KB of memory per character, and a literal past about 5000 characters
overflowed the host stack in `term_check`. A Nat literal had the same
problem before 2.0.7 (#779, #791).

The compiler sees no difference. `def_body` in comp.ts expands every literal
back into its chain (`lit_expand`) before the compiler lifts a body. The
emitted C and JS are byte-identical to main for every test in `tests/`. The
C, Metal and CUDA lanes run the same code as before.

A JS string holds Unicode scalar values only. A literal that spells a
surrogate or a code point past U+10FFFF still parses as its chain
(`tests/check/string_literal_surrogates.bend`).

## Numbers

Checker peak memory and wall time with `--check-only`, bun 1.4.2, WSL2 on
x86_64. "before" is main at 6018e28e.

| Program | before | after |
|---|---:|---:|
| 50 defs of one 1000-character literal each | 2.60 GB, 4.00 s | 74 MB, 0.14 s |
| one 200,000-character literal | stack overflow | 76 MB, 0.17 s |
| a CLDR table module of 185 K literal characters (out of tree) | 11.2 GB, 26 s | 248 MB, 0.9 s |
| a 2466-line check suite over the CLDR plural rules (out of tree) | 10.3 GB, 16 s | 190 MB, 0.5 s |
| bench/checker/proofs_3200 (3 runs) | 2.38 to 2.56 s, 821 to 843 MB | 2.24 to 2.50 s, 831 to 839 MB |
| bench/checker/generics_3200 (3 runs) | 1.12 to 1.28 s, 353 to 369 MB | 1.09 to 1.20 s, 350 to 358 MB |

Emission: `-o x.js` and `-o x.c` of a 4000-character literal overflowed on
main. They work now: 15.5 KB of JS, 255 KB of C. The emitters still walk a
literal as its chain, so one literal past about 8000 characters still
overflows in `-o`. The checker has no limit.

## Files

| file | change |
|---|---|
| bend2/bend.ts | the `Lit` node: type, `Lit`, `lit_of`, `lit_chain`, `lit_step`, `lit_expand`; one case each in `term_higher`, `term_lower`, `term_snf`, `term_show`, `parse_patt`, `match_flatten`; the unfolding in `term_wnf`, `term_compare`, `term_descend`; the `check-lit` rule; the string case of the parser; `chr_show`, shared by the char and string printers |
| bend2/comp.ts | `def_body` expands literals before lifting |
| gates/repo.ts | bend2/bend.ts may reach 43500 ttok (was 42000) |
| tests/check/string_literal_long.bend | a 6000-character literal checks and interprets |
| tests/check/string_literal_surrogates.bend | a surrogate escape keeps its two characters |
| tests/printer/string_literal_tail.bend | a literal prints back; a chain that ends in a literal prints as one string |
| tests/proof/string_literal_unfolds.bend | equalities between literals and chains, `String.length`, `String.append`, `String.eq`, a match on literal patterns |

bend2/bend.ts: 41915 -> 43203 ttok (cl100k, counted with js-tiktoken; the
`ttok` binary was not available here). bend2/comp.ts: 63466 -> 63471.

bend2/bend.lean does not change. The core has no literal forms. Every
literal sugar already lives outside it and means its constructor chain.
`Lit` follows that convention, as a Nat literal above 256n does since 2.0.7.
The core judges the chain. The checker computes the same judgment on the
compact form, and every read of a `Lit` unfolds it.

AGENTS.md says bend2/bend.ts is human-written and not to be edited. This
change is in that file because the cost is in the checker's representation.
No other file can remove it. The diff is 130 lines. Every new function is
named `lit_*` and sits in one section after `u32_to_term`.

## Verified locally

- `tests/`, 1382 files (1378 on main plus the four here): the JS-lane run
  output (stdout, stderr and exit code) of the 1273 tests outside tests/io is
  identical before and after, except `string_literal_long`, which overflows
  on main. This harness does not run tests/io, which waits on the gate's
  harness. The emitted C (601 files) and JS (617 files) of every test are
  byte-identical.
- The four new tests pass. On main, `string_literal_long` overflows the
  stack, and the other three pass with the same output.
- `tsc -p bend2/pack/tsconfig.json` reports the same two pre-existing errors
  and no new one.
- Not run here: gates/test.ts on the mini cluster. The C, Metal and CUDA
  lanes are covered by the byte-identical emission only. Please run the gate.

## Related

- #779 and #791: the Nat literal precedent, fixed in 2.0.7 by parsing a large
  literal as `U32.to_nat(n)`.
- #892: three string-literal match arms exceed 512 MiB during C generation.
  That report has the same root, on the emitter side. This change does not
  touch the emitters. A `Lit` reaches `def_body` intact, so an emitter that
  wants the text has it.
- #873: strings as views over buffers in the runtimes. A `Lit` is a direct
  constant for that representation.


VictorTaelin:
Merged together with #924 as one commit under your authorship (942573fd), thank you. See the note there.

PR #911, opened 2026-09-20
Feat/array map

jtranq:
## Summary

This PR removes `Array.map`'s `O(n log n)` split/rebuild copying and replaces it with direct destination construction. 

**On a `4096 × 4096` transformer weight-initialization workload, the same Bend program drops from 2.01 s to 0.143 s (~14× faster), moving from ~16× slower than C to within 14% of C, while peak RSS falls from 194 MB to 130 MB.**

Optimize the existing `Array.map` in the native compiler by replacing recursive array splitting and rebuilding with compiler-generated workers that read the source and write directly into a single destination allocation. This addresses a mismatch between the abstraction and the representation. Arrays expose a tree-shaped interface but their native storage is contiguous.

The lowering is very similar to [Futhark's](https://futhark-lang.org/blog/2023-12-04-end-of-an-optimisation.html?) sequentialisation and array short-circuiting passes, which compile map into indexed traversal over preallocated result storage and push result construction into its final destination to avoid redundant copies. You can read a deeper description of that by [Troels Henriksen](https://sigkill.dk/) at the above hyperlink.

Existing callers benefit with no code changes. The public signature remains `Type`-generic, including support for affine elements, nested arrays, boxed values, and different input/output layouts. Parallel mapping and callbacks requiring continuations or containing their own forks remain supported. The key observation is that the new algorithm is fundamentally doing less data movement than the old one. So the 3–6% regressions are likely to be overhead introduced by the new worker/scheduling machinery.

One basic result I was able to get was with a materializing FizzBuzz implementation. The `Array.map` formulation became about 1.85× faster and substantially reduced its memory use. Currently the `Array.map` FizzBuzz formulation is not the fastest way to materialize the result, a `List`-based implementation already avoided the pathological array splitting and rebuilding, but now the `Array.map` implementation has become roughly 10% faster than the `List` based one, making it the fastest materializing Bend formulation I tested. 

## Performance

This removes the pathological overhead of cheap elementwise maps producing an **~8.7× improvement on a large trivial map**, but it remains basically performance neutral on compute-heavy callbacks, you can see the other cases are within ~5% of the existing implementation though sometimes mildly slower.

| Workload | CPU workers | Original tree map | This PR |
|---|---:|---:|---:|
| Increment 16,777,216 `U32` elements | 1 | 694 ms | 80 ms |
| 4,096 elements, 50,000 xorshift iterations per callback | 1 | 471 ms | 487 ms |
| 4,096 elements, 50,000 xorshift iterations per callback | 4 | 129 ms | 137 ms |
| Four elements, 50 million xorshift iterations per callback | 4 | 239 ms | 237 ms |

The large cheap map is approximately **8.7× faster**. The 4,096-element expensive map retains approximately **3.6× scaling** from one to four workers, although its four-worker median is about 6% slower than the tree baseline in this run. 

For the same large numeric mapping workload:

| Memory metric | Original tree map | This PR |
|---|---:|---:|
| Median peak resident memory | 193.0 MiB | 128.8 MiB |

That is approximately **33% less peak resident memory**. This implementation still uses separate input and output buffers rather than reusing the input in place.

### Practical example: transformer weight initialization

As a more representative workload, I tested initializing transformer-style weight matrices using `Array.map`.

The benchmark initializes `4096 × 4096` `F32` weight matrices using a Kaiming-uniform bound of `1 / sqrt(fan_in)`, with each weight generated deterministically from its global index using a counter-based hash. This gives the map callback a realistic amount of work in hashing, integer-to-float conversion, and a few floating-point operations, while remaining a pure pointwise transformation. 

#### One worker

Each row below generates the same total number of weights (`2^27`), but groups them into differently sized matrices:

| Workload | C | Old Bend | This PR | Old / New | Old RSS | New RSS |
|---|---:|---:|---:|---:|---:|---:|
| 64 × `2^20` weights | 0.060 s | 0.933 s | 0.067 s | **13.9×** | 14 MB | 10 MB |
| 8 × `2^24` weights | 0.125 s | 2.010 s | 0.143 s | **14.1×** | 194 MB | 130 MB |
| 2 × `2^26` weights | 0.140 s | 2.116 s | 0.168 s | **12.6×** | 770 MB | 514 MB |

For the `4096 × 4096` case, the whole Bend workload goes from roughly **16× slower than C to within ~14% of it** on one worker.

Subtracting the index-generation and result-folding baseline makes the effect on `Array.map` itself clearer:

| Workload | Old map cost | New map cost | Improvement |
|---|---:|---:|---:|
| 64 × `2^20` weights | 14.2 ns / weight | 0.66 ns / weight | **21.6×** |
| 8 × `2^24` weights | 14.5 ns / weight | 0.61 ns / weight | **23.8×** |
| 2 × `2^26` weights | 15.0 ns / weight | 0.53 ns / weight | **28.4×** |

Previously, most of the cost of this operation was not generating the weights at all: it was recursively splitting the input block and rebuilding the output block. With this PR, each generated weight is instead written directly into its final destination slot.

#### Parallel execution

Using Bend's default 10 workers:

| Workload | Old Bend | This PR | Improvement | Old RSS | New RSS |
|---|---:|---:|---:|---:|---:|
| 64 × `2^20` weights | 0.310 s | 0.066 s | **4.7×** | 119 MB | 14 MB |
| 8 × `2^24` weights | 0.646 s | 0.126 s | **5.1×** | 1,238 MB | 131 MB |
| 2 × `2^26` weights | 0.733 s | 0.134 s | **5.5×** | 3,405 MB | 515 MB |

The memory difference becomes especially visible at larger tensor sizes. Producing a 256 MB weight matrix previously reached roughly **3.4 GB of resident memory** with 10 workers; the new lowering reduces that to roughly **515 MB**.

On this workload, the new one-worker implementation lands within roughly **12–20% of the C reference**, while the parallel Bend implementation is roughly competitive with the single-threaded C implementation.


VictorTaelin:
Thank you for this PR: the analysis is right, the numbers are real, and the write-up with the Futhark references is a pleasure to read. Base's `Array.map` was the tree algorithm and paid O(n log n) copies for every map; that had to go.

We are closing it without merging, for a reason of shape rather than of result. The lowering adds about 4,900 tokens to the C emitter, which sits at its complexity cap, and the same result is reachable in Bend itself: a loop of `Array.get`/`Array.set` already compiles to one `for` of a block read, the callback and a block write, with no allocation. Measured on the same M4 node, one worker: your lowering maps 16M U32 in 3 ms; the Bend loop does it in 4 ms in place (66 MB) or 10 ms into a fresh array (130 MB), against the tree's 166 ms and 194 MB; with the existing fork/join pattern it is 1-2 ms on ten workers. So in 2.0.23 Base's `Array.map` is that loop, for `Data` elements. Your workloads are what we measured it on.

Two things you found are kept in mind: the remaining gap on cheap callbacks comes from the masked index in `blk_at` (your worker emits an affine address and vectorizes; the loop's does not), which is a compiler change for another day; and the affine (`Type`) element case, which your lowering covers and the loop cannot, stays on the tree until someone needs it. Closing with real gratitude; please keep sending measurements like these.

Issue #905, opened 2026-09-20
Duplicate template binder names collide during generic checking, forge Data, and admit Empty


jasisz:
### What I did

```text
bun bend2/main.ts duplicate_binder_curry.bend --check-only
```

### What happened

```text
All terms check.
```

The file has a closed `boom() -> Empty` and derives `False{} == True{}` from it. It does not import Base and uses no `@unsafe`, FFI, law, hole, native definition, or runtime backend.

```bend
type Empty is Data:

type Bit is Data:
  False{}
  True{}

def cast(~T: Type, ~x: T, ~T: Type) -> T:
  x

type Bad is Type:
  MkBad{apply: @+x: cast(~Type, ~Bad, ~Data) -> Empty}

def Bad.apply(b: Bad) -> (@+x: cast(~Type, ~Bad, ~Data) -> Empty):
  match b:
    case MkBad{f}:
      f

def step(+b: cast(~Type, ~Bad, ~Data)) -> Empty:
  Bad.apply(b)(MkBad{Bad.apply(b)})

def boom() -> Empty:
  step(MkBad{step})

def absurd(-A: Type, e: Empty) -> A:
  match e:

def false_theorem() -> {False{} == True{} : Bit}:
  absurd({False{} == True{} : Bit}, boom())
```

### Why it happens

`def_check` names each opaque template parameter with `k + "~" + h.k`. The two source binders named `T` therefore both become `cast~T`, and the second declaration overwrites the first one in the generic checking book. That makes this invalid generic cast check as though its source and target types were the same.

Renaming the last binder makes the same definition fail at its body, as expected:

```bend
def cast(~T: Type, ~x: T, ~U: Type) -> U:
  x
```

```text
Error:
- expected : cast~U
- observed : cast~T
Location: cast
```

The exploit uses `cast(~Type, ~Bad, ~Data)` only in type positions. Its declared result type is `Data`, so the checker accepts it as the domain of a `+` binder. Normalization unfolds the raw template body to `Bad`, which is actually `Type`-kinded. This gives the negative `Bad` value a forged contraction license, and the ordinary Curry term in `step` inhabits `Empty`.

This is separate from #902. No live `cast` instance is created here; after checking the file, `book.tmps["cast"]` is still empty. A live use of the same cast does create an instance and is rejected at `cast~0` with `expected Data, observed Type`. The hole is specifically the generic opaque-name collision combined with type-position template reduction.

### Version

```text
bend 2.0.21
main commit 8a51f9de
[...]

Issue #904, opened 2026-09-20
JS emitter produces invalid identifiers for imports from hyphenated directories


2undvierzig:
### What you did

Compiled a small program importing a module from a directory containing a hyphen:

```sh
bend main.bend --check-only
bend main.bend
bend main.bend -o main.js
node main.js
```

Reproduced with the official Linux x64 releases **2.0.20 and 2.0.21**. Node was explicitly in CommonJS mode (`package.json` below).

### What happened

Both Bend versions give the same results:

| Command | Result |
| --- | --- |
| `bend main.bend --check-only` | `All terms check.`, exit 0 |
| `bend main.bend` | `1`, exit 0 |
| `bend main.bend -o main.js` | No diagnostics, exit 0 |
| `node main.js` | Syntax error, exit 1 |

The generated JavaScript contains this invalid identifier:

```js
function $src-lib$Flags$flag$(x_0) {
```

Node reports:

```text
SyntaxError: Unexpected token '-'
```

**Control:** Copying the same `Flags.bend` to `src_lib/Flags.bend` and changing only the import to `./src_lib/Flags.bend` produces JavaScript that prints `1` and exits 0, on both versions.

Expected: emit valid JavaScript preserving the evaluator's result. If such import paths are unsupported, compilation should reject them with a diagnostic instead of successfully emitting unparsable JavaScript.

This appears to be missing identifier escaping for path-derived names. Related area: #790 concerns mangling collisions; this example has no collision and instead produces a lexically invalid identifier.

### The files

```text
main.bend
src-lib/Flags.bend
package.json
```

`main.bend`:

```python
import Base
import ./src-lib/Flags.bend as F

def main() -> U32:
  F.flag(True{})
```

`src-lib/Flags.bend`:

```python
import Base

def flag(x: Bool) -> U32:
  match x:
    case False{}: 0
    case True{}: 1
```

`package.json`:

```json
{"type":"commonjs"}
```

### Bend version

`bend version`: `bend 2.0.20` and `bend 2.0.21`, official release binaries.

### System

`uname -sm`: `Linux x86_64`

Node: `v22.22.0`

### C compiler

Not used; this report concerns the JavaScript target only.

VictorTaelin:
[...]

Issue #875, opened 2026-09-19
User-defined Clo.apply collides with the compiler closure sentinel and duplicates a linear File


jasisz:
### What I did

```text
bun bend2/main.ts clo_apply_duplicate_file.bend
bun bend2/main.ts clo_apply_duplicate_file.bend -o out.js
bun out.js
bun bend2/main.ts clo_apply_duplicate_file.bend -o out
./out
```

The source defines an ordinary function named `Clo.apply`. Its body always returns `False{}`, and the checker accepts a proof of that result:

```bend
def static_tag() -> {runtime_tag() == False{} : Bool}:
  {==}
```

`False{}` selects the linear `File` payload and keeps two distinct files. The program closes the first file and writes to the second, so its source result is `True`.

### What happened

The default IO lane and emitted JS print:

```text
False
```

The native binary aborts with:

```text
bend: an alien request
```

Changing only `Clo.apply` to `User.apply` makes the default lane, emitted JS, and native binary all print:

```text
True
```

The compiler uses the source-level string `"Clo.apply"` as its synthetic closure-application sentinel:

```ts
const CLO_APPLY = "Clo.apply";
```

The JS emitter therefore compiles the direct user call `Clo.apply(return_true, 0)` as a dynamic call of `return_true`. The emitted user function exists but is never called:

```js
function $runtime_tag$() {
  return run_tail(run_clo((x_0) => {
    return run_jump($return_true$, [x_0]);
  }), 0);
}

function $Clo$apply$(f_0, x_0) {
  return false;
}
```

That changes the dependent tag from the checker's `False{}` to runtime `True{}`. The `True{}` branch was checked with a `Data` payload, so it duplicates `a`. At runtime, `a` is the first `File` from the statically selected linear branch. Both result fields therefore contain the same file handle. The program closes `fst`; writing through `snd` then fails because it is the same closed handle, while the distinct second file is dropped.

This is observable wrong-code, dependent subject-reduction failure, and contraction of an opaque linear runtime handle. `File` is declared as an opaque `Type` law specifically so programs cannot forge, copy, or reuse it. The reproducer uses no `@unsafe`, user FFI, open user law, template, malformed pattern, or custom built-in datatype. It has the same dependent-grade consequence as #872, but a different cause: a legal source definition collides with a compiler-internal call sentinel.

A smaller pure reproducer isolates the same collision: the evaluator returns `12`, emitted JS returns `22`, and native C returns `0`.

### The file

```bend
import Base

[...]

Issue #800, opened 2026-09-18
All foreign .js sources share one scope, so a private helper in one FFI module silently clobbers another's (C lane refuses)


haasonsaas:
## Summary

`js_lib` concatenates every foreign `.js` source into a single IIFE and reads the
effects back out of one flat table:

```
const effs = rows.length === 0 ? "" : "const $0eff = (() => {\n"
  + srcs.join("\n") + "\nreturn {\n" + rows.join("\n") + "\n};\n})();\n\n";
```

All foreign modules therefore share one function scope. A top-level name in one
foreign file collides with the same name in an unrelated foreign file, and because
duplicate `function` declarations are legal inside a function body, the last one
silently wins. Every earlier definition, and every call already closed over it,
is rebound.

This needs no Bend-level name collision: the two defs can have completely distinct
names and distinct FFI slots. It is enough that two independent `.js` files each
declare a private helper with the same name -- `fmt`, `parse`, `encode`, and so on,
which is close to unavoidable once more than a couple of FFI modules exist.

`tests/io/effect_runtime_shadow.bend` fixed the splice-vs-runtime direction of this
by moving the splice into one function scope behind a typeof-guarded table. That
stops a foreign def from clobbering a `runtime.js` helper, but foreign files still
share that one scope with each other, so splice-vs-splice is still open.

The C lane refuses the same program.

## What you did

`helper.bend`:

```python
import Base

law Alpha.run:
  for a: U32
  IO(U32)

def Alpha.run(a):
  import "./helper_a.js"

law Beta.run:
  for a: U32
  IO(U32)

def Beta.run(a):
  import "./helper_b.js"

law main:
  IO(Unit)

def main():
  do IO<Unit>:
    x : U32 <- Alpha.run(1)
    u : Unit <- IO.print(U32.show(x))
    y : U32 <- Beta.run(1)
    IO.print(U32.show(y))
```

`helper_a.js`:

```js
function fmt(x) {
  return x + 1;
}

function alpha_run(a) {
  return fmt(a);
}
```

`helper_b.js`:

```js
function fmt(x) {
  return x + 1000;
}

function beta_run(a) {
  return fmt(a);
}
```

```
bend helper.bend
bend helper.bend -o helper.js && node helper.js
```

## What happened

Both lanes print:

```
1001
1001
```

`Alpha.run(1)` is `1001`, not `2`: `helper_b.js`'s `fmt` replaced `helper_a.js`'s,
[...]

PR #926, opened 2026-09-21
Prefix generated C locals: ts_32 is a host macro on macOS and the native build fails

nood-co1:
`bend prog.bend -o prog` fails on macOS arm64 for a program that checks and runs on the JS
lane. `name_local` (`bend2/comp.ts:648`) builds every generated C local from the Bend-level
name plus a counter, with **no prefix**:

```ts
function name_local(fl: File, k: Bend.Name): string {
  const base = name_clean(k);          // <- "ts"
  const n = fl.fresh.get(base) ?? 0;
  fl.fresh.set(base, n + 1);
  return base + "_" + n;               // <- "ts_32"
}
```

so a generated local can land on a macro the platform headers already define. On macOS one
does: `ts_32` is `#define ts_32 uts.ts_32` in `<mach/arm/thread_status.h>`, and it is reachable
from `Base` itself — `bend2/base.bend:1459` declares `U32.divmod.go.shl(..., ts: Bool &
Word(32n))`, and compiling it mints `ts_0` … `ts_32`.

### Repro (12 lines)

```bend
import Base

def snd(p: Word(64n) & U32) -> U32:
  (q, r) = p
  r

def half(w: Word(64n), b: U32) -> U32:
  snd(U32.divmod.go(64n, w, b))

def main() -> IO(Unit):
  do IO<Unit>:
    IO.print(U32.show(half(Word.zero(64n), 10)))
```

```
$ bend repro.bend --check-only   # All terms check.
$ bend repro.bend                # the JS lane: 0
$ bend repro.bend -o repro       # native build
…/repro.c:1580:7: error: expected ';' at end of declaration
 1580 |   u32 ts_32 = r34;
/…/MacOSX.sdk/usr/include/mach/arm/thread_status.h:199:18: note: expanded from macro 'ts_32'
  199 | #define ts_32 uts.ts_32
…/repro.c:1582:655: error: member reference base type 'u32' is not a structure or union
```

### The change

```diff
 function name_local(fl: File, k: Bend.Name): string {
-  const base = name_clean(k);
+  const base = name_clean(k).replace(/^bend_/, "");
   const n = fl.fresh.get(base) ?? 0;
   fl.fresh.set(base, n + 1);
-  return base + "_" + n;
+  return "bend_" + base + "_" + n;
 }
```

The strip is required for the case measured, not cosmetic: `seg_open` (`comp.ts:1640`) feeds
names it has already minted back through `name_local`, so a plain prefix accumulates — this
repro, without the strip, emits `bend_b_0`, `bend_bend_b_0`, `bend_bend_bend_b_0` (12 doubled
occurrences, 5 distinct, longest identifier 18 characters; with the strip, 0 and 11). It closes
the observed path; it is not a general idempotence property.

`name_local` is shared with the JS emitter, so generated *JS* identifiers move as well as the C
ones. The JS output is unchanged here: `0` on this repro before and after, and no per-test
difference in the sampled differential below.

Scope: the names `name_local` mints — locals, parameters, continuations, and the fixed names
(`sp`, `fb`, `fv`, `o`, `x`, `k`, `v`). Top-level/def naming is untouched, and `CID_`/`FID_` are
already prefixed.

### Cost

`bend2/comp.ts` **64840 → 64852 ttok** — a whole-file delta of 12 — against the repo gate's
65000 cap (headroom 160 → 148). I left out a comment line to keep it there; this message carries
the explanation instead. `bun gates/repo.ts`: **PASS 46/46**.

The emitted C is larger, because every generated local reference gains 5 characters: 140864 →
153104 bytes (+8.7%) on this 12-line repro. I have not measured that ratio on a large program.

### What I checked

| check | result |
|---|---|
| native build, before / after | `rc=1`, `expected ';' at end of declaration` / **`rc=0`**, binary prints `0` |
| check (`--check-only`) and the JS lane, before / after | `All terms check.` and `0`, both times |
| generated C, bare `ts_N` locals | 33 distinct before, 0 after (33 `bend_ts_N` after) |
| interpreter differential, whole `tests/` corpus | **identical on all 1395 tests** |
| cross-lane sample, 106 tests with a `main` | **0 of 106 per-test records differ**; 51 built on both compiled lanes; the same 2 pre-existing divergences on both sides (`comptime/unsafe.bend`, the interpreter-only unsafe banner; `gfx/window.bend`, which needs a display) |
| `bun gates/repo.ts` | PASS 46/46 |

`gates/test.ts` and `gates/perf.ts` need the SSH cluster and `gates/ping.ts` cannot run outside the
site project. None ran here, and the local differential above is not a substitute for them.

### Not a duplicate

Searching the 920-item issue and PR snapshot, I found no direct duplicate about the C emitter's
generated locals meeting a host macro. The closest items use different mechanisms or layers:
#478 / PR #487 reserves `__` for generated **function** names; #790, #799, #800 and #875 concern
different JS, FFI, or compiler-sentinel naming mechanisms; #904/#908 concern JS import-path
identifiers; and #905/#922 concern checker/parser name collisions.

### What this does not claim

* Not that collisions are impossible after it: `bend_` is an ordinary C identifier prefix and a
  foreign header could still define `bend_x_0`. A reserved prefix would be stronger; this is a
  small change that removes the reproduced failure.
* Not that the compiled lanes are regression-free: the interpreter differential is complete over
  `tests/`, the compiled lanes are sampled.
* Not that `ts_32` is the only such name on any SDK: it is the one reachable from `Base` on this
  host, and the change addresses the class while claiming only the observed instance.

---


VictorTaelin:
Merged into 2.0.25 under your authorship, thank you, with one trim: the prefix is a single `_` (`_ts_32`), which closes the same class (every generated local is a block-scope identifier, where `_x` is not reserved) at +0.9 to +4.6 % of emitted C instead of +8.7 to +23 %. Your 12-line reproduction builds and prints 0. We also found the include chain: `<mach-o/dyld.h>` for `_NSGetExecutablePath` pulls `thread_status.h`, and `ts_32`/`ts_64` are the only lowercase macros it leaks in that shape.

PR #929, opened 2026-09-21
feat: add comprehensive F32 math library

victormeloasm:
Adds a comprehensive F32 math library to Bend, modeled after the C `math.h` API and adapted to Bend's current numeric type system.

The new library includes trigonometric, hyperbolic, exponential, logarithmic, power, rounding, remainder, gamma, error-function, IEEE-754 classification and helper operations, along with common mathematical constants such as PI, E and sqrt(2).

This PR also adds the backend support required for operations that cannot be implemented accurately using only existing F32 primitives, with consistent behavior across the C and JavaScript backends.

Highlights include:

* new `bend2/math.bend` module
* C99/C11-style math functions applicable to F32
* additional C23-style functions such as `sinpi`, `cospi`, `tanpi`, `pown`, `powr`, `rootn`, `rsqrt`, `exp2m1` and `log2p1`
* explicit handling for signed zero, subnormals, NaN and infinities
* fused F32 FMA support
* numerically stable implementations for sensitive operations such as `expm1`, `log1p`, `erf`, `erfc`, scaling and integer powers
* new regression, edge-case and backend consistency tests

Local validation:

* `bun gates/repo.ts`: PASS 48/48
* randomized C/JS differential testing and IEEE-754 edge-case testing passed
* UBSan, float-cast-overflow, ASan and LeakSanitizer testing passed
* `gates/test.ts` could not be executed locally because it requires access to the project's private mini-cluster SSH host


victormeloasm:
Bump

pjcavalcanti:
Thanks for working on this. I’m going to close this PR in its current scope, but would welcome smaller contributions for operations whose benefits justify adding compiler support.

- **It breaks GPU compilation even for programs that use none of the new math functions.** On Metal, macro expansion produces invalid names such as `precise::precise::pow`. On CUDA with NVRTC 12.3, the helpers reference unavailable `NAN` and `INFINITY` macros. These helpers are included unconditionally, so even integer-only GPU programs fail to compile on the affected toolchains.

- **Some functions return incorrect results.** For example, `round(0.4999999701976776123046875)` returns `1` instead of `0`, and `lround` and `llround` inherit the error. The input is strictly below `0.5`; the intermediate F32 addition in `floor(x + 0.5)` rounds away that distinction.

- **The compiler grows substantially, and splitting out `math.ts` moves that growth outside the existing compiler budget.** Counting `comp.ts` and `math.ts` together at `d7d6408`, the change is **64,840 → 72,689 tokens: +7,849 (+12.1%)**. We want the compiler to remain in `comp.ts`; adding another file and allowance does not reduce the code we must understand and maintain.

- **Our measurements do not justify the full set of compiler additions.** We implemented and optimized equivalent functions in pure Bend, then compared our library with the pre-PR compiler against this PR’s library and compiler. Across 96 workloads, single-worker native C was **over 10% faster with the PR in only 8 workloads, slower in 47, and within 10% in 41**. Four-worker CPU execution gave approximately the same comparison. JS benefits more broadly—58/96 workloads—but its compiler is less optimized at present, so those gains can also reflect limitations in existing JS code generation. **We are therefore leaning on the C numbers for this judgement.** GPU performance could not be measured because compilation fails. These comparisons include accuracy differences and measure the two complete implementations, not the compiler changes alone.

Given these results, the broader math collection should be implemented in pure Bend and distributed as a **user library**, using `bend math.bend --publish` so others can import it by its content hash. This needs no PR to this repository and does not expand Base.

What we would welcome here are **small PRs for the compiler operations that justify their cost**. FMA is the strongest starting point: roughly 10× faster in our ordinary and cancellation C workloads, matching the tested reference results. `pown` also merits consideration, with 37–39% gains for large exponents. These and any other proposed additions should demonstrate their performance or accuracy benefits, pass tests across the supported backends, and respect the repository’s organization and token budget, keeping compiler code in `comp.ts`.

PR #932, opened 2026-09-21
A read parked on a FIFO sees its end on macOS: the IO loop waits with select (#928)

### LINKED ISSUE #928: macOS: poll() does not report a FIFO close, so a parked read never sees EOF
> [!NOTE]
> This issue was investigated and written by an AI agent, at the repository
> owner's direction, on the owner's machine. All outputs below come from runs
> of the shown commands. The owner read this text before posting.

## Summary

On macOS, a computation that parks on a FIFO never returns at end of file.
Both lanes wait with `poll()`, and `poll()` does not report a FIFO close on
macOS. The `select()` call does. So reading a FIFO to the end hangs.

## Symptom

An effect that parks on fd 0 reads a pipe and a file to the end. On a FIFO it
stops after the data:

```sh
printf 'hello\n' | ./reader      # LINE[hello] LINE[] — done
./reader < /tmp/two.txt          # LINE[a] LINE[b] LINE[] — done
mkfifo /tmp/f; ( sleep 1; echo hi ) > /tmp/f &
./reader < /tmp/f                # LINE[hi], then nothing, forever
```

The parked call:

```c
return io_wait_on(w, 0, POLLIN, 0, stdin_read_line_more);   // fd 0, no deadline
```

## Scope

One probe waits on a FIFO. The same probe ran on three machines:

| | `poll` | `select` |
|---|---|---|
| macOS 27.0 (26A428), arm64 | never | True |
| Linux 5.15 | `revents=0x10` | True |
| iSH (a system-call emulator) | never | never |

A named FIFO fails and an anonymous pipe does not. `fstat` answers `S_ISFIFO`
for both, so the name is what matters. `< file`, `producer | prog`, a pty and
`<(cmd)` all see EOF.

## Where the runtime waits

- The native lane: `while (poll(fds, n, ms) < 0)` in `io_wait`, in the C that
  `bend prog.bend -o prog.c` emits.
- The JS lane: `io_sys().poll(...)` in `io_wait`, in the emitted JS.

## Reproducer, without Bend

```c
#include <stdio.h>
#include <fcntl.h>
#include <poll.h>
#include <sys/select.h>
#include <unistd.h>

int main(int argc, char** argv) {
  int use_select = argc > 1 && argv[1][0] == 's';
  char c;
  for (int i = 0; i < 20; i += 1) {
    int ready;
    if (use_select) {
      fd_set r; FD_ZERO(&r); FD_SET(0, &r);
      struct timeval tv = {0, 400000};
      ready = select(1, &r, NULL, NULL, &tv);
    } else {
      struct pollfd p; p.fd = 0; p.events = POLLIN; p.revents = 0;
      ready = poll(&p, 1, 400);
    }
    printf("%s -> %d\n", use_select ? "select" : "poll", ready);
    if (ready > 0 && read(0, &c, 1) == 0) break;   // 0 from read is EOF
  }
  return 0;
}
```

```sh
clang -o w w.c
rm -f f; mkfifo f; ( sleep 1; echo hi; sleep 1 ) > f &
./w poll   < f    # 0 forever, also after the writer exits
./w select < f    # ready, then read -> 0 (EOF)
```

The same holds when the program opens the FIFO itself instead of the shell.

## Workaround that works

A deadline on the park makes the effect wake and read again:

- C lane: `io_wait_on(w, 0, POLLIN, io_tick() + 250000000ull, more)`
- JS lane: `io_park_on(0, false, k, go, performance.now() + 250)`

Both lanes then read a FIFO to the end, and their other computations keep
running. `tcp_poll.c` already uses this shape.

## Question

The difference is in Darwin, but Bend decides how it waits. Which do you want?

1. The runtime handles this case on Darwin, for example with a `select` wait.
2. The guide states where a park can wait, and a FIFO as stdin is unsupported.

## Environment

Bend 2.0.20 on macOS 27.0 (build 26A428), arm64. I have one macOS machine, so
I cannot say whether other versions behave the same.

PedroVIOliv:
Fixes #928.

## What changed

macOS's `poll` never reports a FIFO's close, so a computation parked on one never wakes at end of file. Both lanes' `io_wait` now wait with `select`, which does report it. The sets are sized to the highest parked fd, so descriptors above 1023 keep working as they did under `poll`:

- C lane: `_DARWIN_UNLIMITED_SELECT` on macOS; Linux's `select` takes sets of any size.
- JS lane: binds `select$DARWIN_EXTSN` on macOS (the plain symbol refuses an fd past `FD_SETSIZE`), `select` elsewhere.

Probed on macOS 15.5 against a FIFO whose writer has closed: `poll` with `POLLIN` or `POLLIN|POLLPRI|POLLRDNORM|POLLRDBAND` never reports it, `kqueue` `EVFILT_READ` returns no event, plain `select` reports it but fails at fd 1500, and `select$DARWIN_EXTSN` reports it at fd 1500.

## Commits

1. **The IO layer's comments and scheduler say the same in fewer tokens:** a separate cleanup that makes room for the fix under `comp.ts`'s cap (the JS generic park goes through `io_park_on`, its run loop tests `op` once, tighter IO comments, fewer temporaries). comp.ts: 64,936 → 64,682. On its own it changes no test result.
2. **The fix and its test:** 64,682 → 65,000, at the unchanged cap. `gates/repo.ts` passes 46/46.

## Test

`tests/io/fifo_eof.bend` (C and JS twins): a child writes into a FIFO and closes it only once the reader has drained it and parked again (an ack pipe orders the two), so the reader's last wake is the close alone. Three cases: data then close, close with no data, and the reader at fd 1500. Expected `5 / 0 / 5`.

- On 75cb8f3e without this change: hangs on both lanes, 20/20 runs each.
- With it: 40/40 on each lane under 12-way parallel load; the interpreter prints the same.

## Checked

On macOS 15.5 arm64: the tests that go through the IO loop (147 files: io, sockets, timers, channels, fibers) give the same results on both lanes before and after each commit, plus the new test. Not run on Linux or on the cluster gates.


VictorTaelin:
Merged into 2.0.25 as one commit under your authorship, thank you. We reproduced #928 on both lanes (kqueue does not see the close either), confirmed the fix, ran the 26 IO-loop tests on both lanes and trees, and checked `select` past fd 1023 on Linux. Your cleanup commit went in with it.

Issue #874, opened 2026-09-19
IO foreign defs are trusted invisibly: a forged Rfl silently transports Unit{} into F32 arithmetic — the verdict should count foreign defs like it counts ~ instances


Cypoe:
### What you did

Two files: ffi_forge_rwt.bend + its JS twin ffi_forge_rwt.js (below; also at Cypoe/bend-probe). bun bend2/main.ts ffi_forge_rwt.bend

### What happened

```text
> nan
> exit 0
```
The checker demands the {Unit == F32 : Type} evidence live (Rwt checks it) but erases it in codegen ([t.f]), so the forged Rfl is never inspected — Unit{} flows into F32.add unchecked. `All terms check.` is printed with no indication the book's consistency rests on a foreign promise.

Sibling repro (ffi-empty-trap): a foreign def forge() -> IO(Empty) checks clean (ffi_forge_verdict.bend → All terms check.) and fail-stops the runtime when run — IO(Empty) is a payload no honest foreign impl can ever deliver, since Empty has zero constructors.

The ask: #841 already counts ~ instances as unsafe annotations (t.u === true || k.includes("~")). Foreign defs (tld.i set) are the bigger trust boundary and invisible to the same verdict — All terms check, with N foreign defs is the same disclosure applied to FFI. Optionally, IO(E) for ctorless E is statically refusable. (Neither demo is a kernel inconsistency — every lie arrives IO-headed.)

`bend.ts`'s header claims "no closed live inhabitant of `Empty`" and carves out
only `@unsafe`. Foreign books aren't carved out, so the stated invariant is
falsified as written. `IO(Empty)` is also the one foreign payload no honest
implementation can ever deliver to a continuation — provably a lie, not merely
unverified. 

JS emitter path only; C build fails earlier at emit-time layout check.
comp.ts:1737 (val_word) — the C emitter tracks value layouts and catches Unit{} (0 words) being read as F32 (1 word) at compile time. Plus the C backend requires a .c twin per foreign def (comp.ts:2733), which this repro doesn't ship — it would die on that too.

### The file

```python
function forge_eql(kont) {
  return { $: "Rfl" };
}


import Base
 
def forge_eql() -> IO({Unit == F32 : Type}):
  import "./ffi_forge_rwt.js"
 
def main() -> IO(Unit):
  do IO<Unit>:
    p : {Unit == F32 : Type} <- forge_eql()
    x : F32 = %r@p : _; Unit{}
[...]

Issue #805, opened 2026-09-18
The @unsafe count is disclosed only on the no-main path: a harmless main, or any -o emit/build, reports nothing


phenomenon0:
### What you did

One fixture, three commands. `main.bend` imports `PROOF.bend`, which imports `unsafe.bend` (`@unsafe def contradiction() -> {0n == 1n : Nat}`):

- `bend PROOF.bend` (no main)
- `bend main.bend` (main = 7)
- `bend main.bend -o main.js` and `bend main.bend -o main` (emit / build)

### What happened

`bend PROOF.bend`:

```
1 term annotated as unsafe.
The code is well-typed, but may contain logical paradoxes.
```

`bend main.bend`:

```
7
```

`-o main.js` and `-o main`: no output at all. Exit 0 in every case.

The only @unsafe disclosure comes from the no-main path; a harmless `main` (or any `-o`) drops it while the unsafe def stays in the import graph. In bend2/main.ts at 0b7e2b11, `cli_report` (line 397) is the only place the count is printed, `book_run` calls it only in its no-main branch (lines 454-460), and `cli_emit` (line 196) never reports it. Reproduced on 2.0.4 and 2.0.5. (Related: #776 covers the proof-mode exit code; this one is the disclosure vanishing in run/emit.)

### The file

```bend
import Base
import ./PROOF.bend as P

def main() -> U32:
  7
```

(the other two files: PROOF.bend states a law discharged by `U.contradiction()`; unsafe.bend holds the `@unsafe def contradiction() -> {0n == 1n : Nat}: contradiction()`)

### bend --version

bend 2.0.5 (2.0.4 the same)

### uname -sm

Linux x86_64

### clang --version (the first line)

clang version 21.1.7 (Fedora 21.1.7-1.fc43)

VictorTaelin:
Shipped in 2.0.8: the unsafe count also prints on stderr before a run, an emit or a publish, so no path is silent about `@unsafe`. The no-main verdict is unchanged: "All terms check." or "All terms check, with N unsafe annotations."

*Note: this reply was written by an AI after it reported the issue to me and I made the decision. If anything here is wrong, reply and I will review it myself.*

PR #933, opened 2026-09-21
Report unsafe dependencies in the Node and Bun loaders

vicmcorrea:
i'm evaluating Bend for small survey-scoring functions called from Node in our TypeScript monorepo. The CLI warned about an `@unsafe` helper, but importing the same file through Node or Bun was silent.

this reuses `cli_report` in the shared loader, with the same file-level dependency analysis as the CLI. Warnings go to stderr. Safe imports stay quiet, and return values and exit codes stay the same. Follows #805 and #874.

Checked on macOS arm64 with Node 24.19.0 and Bun 1.3.11:

- 37 local checks covering safe, unsafe, transitive, foreign and repeated imports, Bun bundling, invalid programs, and existing diagnostic fixtures.
- Repository gate passed, 46/46. `git diff --check` passed.
- TypeScript 7.0.2 reports the same four errors on untouched main and this branch, two in `bend.ts` and two missing `canvas` imports in documentation scripts. Cluster gates were not run.

<details>
<summary>Small reproduction, run from the checkout root</summary>

```sh
probe_dir=$(mktemp -d)
trap 'rm -rf "$probe_dir"' EXIT
cat > "$probe_dir/value.bend" <<'EOF'
import Base

@unsafe
def value(n: U32) -> U32:
  n
EOF
printf 'import B from "./value.bend"; console.log(B.value(5));\n' > "$probe_dir/main.mjs"
BEND_NO_TELEMETRY=1 bun bend2/main.ts "$probe_dir/value.bend" --check-only
node --import ./bend2/main.ts "$probe_dir/main.mjs"
bun --preload ./bend2/main.ts "$probe_dir/main.mjs"
```

Both imports still print `5` and exit 0. They now also report `value` as relying on unsafe code, on stderr. Before this change, only `--check-only` disclosed it.

</details>


VictorTaelin:
Merged into 2.0.25 under your authorship, thank you. An import of a file that relies on `@unsafe` or foreign code now prints the same verdict the CLI prints, on stderr, from both loaders.

PR #936, opened 2026-09-21
Feature/u32 to nat

MattCozendey:
## A U32 literal is U32.from_nat over a nat literal

Follow-up to #907 and #924, which made a string or nat literal one `Lit`
node. A bare number now parses as `U32.from_nat(3n)`: one `App` over
one `Lit`, where before the parser built a 33-word `U32{WCon{..}}` chain
of about 100 nodes per number. The compiler no longer expands a nat past
256 into `U32.to_nat` over a word either.

In bend.ts the parser emits `App(Ref("U32.from_nat"), Lit(n))` and
`u32_lit` reads that shape back. `term_wnf` takes `U32.from_nat` on a
nat literal straight to its word (`u32_word`, one shared term per
number), so a proof over `4294967295` costs what it did before. That
shortcut is a compressed run of the def's own unfolding, and
definitional equality is unchanged. `lit_ctr` reads any literal as its
constructor, a U32 literal as its word and a string or nat literal one
step, and the termination check, the match frame, `term_compare` and
check-lit call it where each used to step a `Lit` by hand. Patterns and
the printer read the shape too: `case 0:` still matches, and an error
spells `3` rather than `U32.from_nat(3n)`.

In comp.ts a nat `Lit` is a leaf. `lit_word` gives the immediate for
either literal and feeds `term_const` and both emitters, so no
`U32.from_nat` call and no `U32.to_nat` chain reaches the output. The
compile-time fold still reads a literal scrutinee as its constructor, so
a match on a literal argument folds as it did.

In base.bend, `Word.zero`, `Word.inc`, `U32.inc` and `U32.from_nat`
move under `def Word(n)`. Forward references are refused and `IO.fork`
writes `1` early in the file. The base case is `U32{Word.zero(32n)}`.

Measured with both trees on tmpfs:

| | before | after |
|---|---|---|
| 2000 defs, each a wide U32 literal | 1.03 s, 217 MB | 0.26 s, 81 MB |
| demos/app_slash_boss_3d `--check-only` | 1.80 s, 420 MB | 1.43 s, 319 MB |
| demos/proof_numerics `--check-only` | 0.21 s, 74 MB | 0.18 s, 70 MB |

The gain scales with how many numbers a program writes. Emitted C over
tests/ and bench/runtime is byte-identical apart from `0` spelled `0ull`
and one literal list that now lives in the static image. JS inlines
`3000n` where it had `BigInt(3000)`. The bench binaries print the same
results.

Two things a user can observe. A literal now infers its own type, since
it is an application of a typed def, where a bare constructor needed a
goal: `x = 5` with no annotation types as `U32`, and
`tests/parse/rewrite_non_equation` is repinned because the miss is now
the equation rather than the literal. And a file without `import Base`
that writes a bare number reports an undefined `U32.from_nat` instead of
an unknown constructor.

New tests: `check/u32_literal_proof` (wide and wrapping proofs, literal
patterns), `check/u32_literal_show` (a literal prints as its number) and
`compile/u32_literal_word` (immediates on all three lanes). Locally the
interpret lane over all 1415 tests has the same pass set as `main` plus
those three, and the C and JS lanes were built and run on the 118 tests
whose emitted code changed. The cluster gates were not run from here.
Caps move to 43400 for bend.ts (at 43338) and 65300 for comp.ts (at
65193).

bend.ts is edited, in the parser, `term_wnf`, the printer and the
pattern reader, since the change lives there by nature. The `term_wnf`
shortcut is keyed on the def name `U32.from_nat` in three places: the
parser, `u32_lit` and the guard. A rename in base.bend would fall back
to the slow unfolding without an error. bend.lean says literals are base
constructors, not calculus, and may want a line about this reduction.


MattCozendey:
Adding comment to see if it appears again?

nicolas-abril:
Thank you for this. The diagnosis is right and the speedup is real: a U32 literal is 66 constructor nodes today, and parse, lift, check and the report walk each pay per node. One node per literal is the fix.

We are not going to take this shape for it. `U32.from_nat(3n)` makes a literal a redex instead of a value, so every consumer of terms has to recognize it or unfold it: the name-keyed shortcut in `term_wnf`, `lit_word` and `term_const` in the compiler, the `default` arm of `parse_patt`, the printer, `lit_ctr` through four sites, and the base.bend reorder. Each is a place that drifts, and a rename of `U32.from_nat` silently falls back to the slow path.

We could either try what #907 and #924 did for Nat and String: make the literal a `Lit`. A U32 variant whose `lit_step` is the memoized word rides the seams that already exist with no site-specific code, and the emitter already collapses a word to an immediate. Inference of a bare `5` is a separate change, and nat and string literals do not infer today either.

MattCozendey:
I'll make a new PR fixing this soon

MattCozendey:
I've made my branch _into main_ so that the PR remains here with all necessary context. Updates to come soon.

MattCozendey:
I noticed a low hanging fruit: If we are going to do U32 as literals, might as well do F32  on the same path,by using a "k" field to identify the type (the Nat literal optimization requires that the field to be absent)

These are the medians from seven runs after (1) warm-up on an i7-13650HX, WSL2 with and Bun 1.4.2

| Workload | Time: main → branch (ms) | Speedup | Memory: main → branch (MiB) |
|---|---:|---:|---:|
| Mandelbrot compilation | 192.4 → 149.9 | 1.28× | 101.4 → 86.5 |
| N-body compilation | 420.3 → 263.8 | 1.59× | 162.7 → 102.5 |
| Raytrace compilation | 324.8 → 246.5 | 1.32× | 134.1 → 101.0 |
| K-means compilation | 217.1 → 179.5 | 1.21× | 105.6 → 94.6 |
| Repeated literal definitions | 453.7 → 96.6 | 4.70× | 255.2 → 76.8 |
| Distinct literal definitions | 458.9 → 96.0 | 4.78× | 256.3 → 76.6 |
| Repeated-word proofs | 541.1 → 207.8 | 2.60× | 159.5 → 93.0 |
| Distinct-word proofs | 540.8 → 199.2 | 2.71× | 158.1 → 90.2 |

The Definition tests contain 2,000 literals of each numeric type, the proof tests contain 1,000 U32 increment proofs and 1,000 F32 bit-extraction proofs and the five checker benchmarks are between 3% faster to 5% slower with overlapping time ranges ("trees_400" has the largest slowdown[5.4%])

The code grew by 205 tokens and all four compilation benchmarks produce equal C and JS files between both versions.

MattCozendey:
Closing this in favour of more general: https://github.com/bendlang/bend/pull/1004 or https://github.com/bendlang/bend/pull/1003

Issue #884, opened 2026-09-19
Deque in Base


kbrianps:
### What Bend should do

```python
import Base

def main() -> List<&2, U32>:
  d = Deque.new(&2, U32)
  d = Deque.push_back(&2, U32, 1, d)
  d = Deque.push_back(&2, U32, 2, d)
  d = Deque.push_front(&2, U32, 0, d)
  Deque.to_list(&2, U32, d)  # [0, 1, 2]

# Deque.pop_front and Deque.pop_back answer Maybe<&1, A & Deque<a, A>>:
# the element next to the deque that is left, or None when it is empty.
```

### Why

tests/base/heap_queue_deque.bend pins Heap, Queue and Deque as absent, and gives "Map and Set are its containers" as the reason. I would like to know whether that line still stands before writing any code for the repo.

I wrote a Deque while learning the language: two lists, a front and a reversed back, with the back reversed into the front when the front runs out. Each element moves at most once, so pushing and popping on either end is amortized O(1). It is around 60 lines in the Base style with its helpers, and a Queue is the same type using two of the four operations.

What makes it worth having in Base instead of in every project is the proof. Taking to_list as the reading of a deque, the laws are short enough to be free: to_list(push_front(x, d)) == x <> to_list(d) and to_list(new()) == Nil{} both hold by computation, which I have checked. Anyone writing their own deque has to state and prove that again.

If it belongs in Base, I will send a PR with the defs, a tests/base/deque_ops.bend in the #| format, the laws as a tests/proof file, and heap_queue_deque.bend narrowed to Heap and Queue. If it does not belong there, that is a fine answer too.

VictorTaelin:
Thank you for writing this, and sorry for the answer: we are not taking a Deque into Base right now.

Base has a size cap we hold to deliberately, and a deque is both easy to write and easy to get right outside it. That makes it a poor use of the budget, not a poor implementation: your code is fine.

Please publish it on bendhub as a library. That is exactly the kind of thing we want there, and people will find it.
*Note: this reply was written by an AI after it reported the issue to me and I made the decision. If anything here is wrong, reply and I will review it myself.*

kbrianps:
Makes sense, thanks. It's on the hub now:

```
import 0x12ed1da25aa4f36622687d603aeb34e0/deque.bend as Deque
```

[...]

PR #937, opened 2026-09-21
Base compiles a regex and runs it on a Pike VM

phenomenon0:
This follows the regex design in #873, taken in smaller steps. It is the first
step: a regex in Base alone, with no runtime or compiler change.

`Regex.compile(p)` gives a `Result<String, Regex>`. On a compiled regex:

- `Regex.exec(re, s, at)` is `search`.
- `Regex.match_at(re, s, at)` is `match`.
- `Regex.fullmatch(re, s)` is `fullmatch`.
- `Regex.group(s, m, k)` gives a group's text.

The semantics are CPython 3.11 `re` under `re.ASCII`, for literals, `.`,
sets, `\d \w \s \D \W \S \b \B`, `^ $`, groups, `(?:`, `|`, and `* + ?`,
greedy or lazy. Offsets count code points. The VM is a Pike VM with threads
in priority order, so a run is linear in the subject.

A pattern outside the slice fails to compile; it never runs with a different
meaning. The cut constructs each fail this way, and each is a later step:
- flags
- counted repeats (so a lone `{` fails too, where CPython reads a literal)
- `\x \u \0`, backrefs, `\A \Z`
- named groups and lookarounds
- possessive repeats
- `*`/`+` on a body that can match empty

Also not in this step: `findall`/`split`/`sub`, Unicode classes, and a
native VM.

The types are `Regex.*` because a top-level `Match` would collide with
`def Match` in `evals/hell.glob_backtrack_match.bend`.

| check | result |
|---|---|
| `bun gates/repo.ts` | PASS 46 / 46 at each commit. No cap changed. base.bend 25,423 → 31,449 of 32,000 ttok. |
| `tests/regex`, 6 tests | pass check, interp, C and JS (`gates/test.ts`'s steps run locally) |
| differential vs CPython 3.11 | `tests/regex/oracle.bend`, 160 generated rows: 0 diffs. Offline, 5 × 1,000 rows: 0 diffs on C and JS. |
| laws | exec = first `match_at` from `at` on; `match_at(i)` starts at i; `fullmatch` spans all. Each law fails under a VM mutant. |
| `!` smoke | one compiled regex over a list with `!`; on CUDA, `--gpu on` and `off` give the pin |
| everything else | `--checkup` over all 1,412 tests is byte-identical. The C and JS emitted for the 623 tests that build are byte-identical but for one rename (`List.contains`'s specialization `_0` → `_1` in `tests/base/list_fold`). |
| runtime | `comp.ts`, `bend.ts`, `main.ts` untouched |

Not run here: the cluster gate and Metal. The `!` smoke has run on CUDA
only.

Size: +1,061 lines over 7 files. Base is +485 of those, and 369 test
lines are oracle data.

Base has 551 ttok left after this, so the next Base step will need room.


zxv:

pjcavalcanti:
Thanks for working on this and for the tests. I’m going to close this as an addition to Base, but please publish it as a library on [BendHub](https://hub.bend-lang.com/).

We keep Base small. A regex engine brings a substantial parser and matcher, with its own syntax and compatibility rules to maintain. Your implementation already uses ordinary Bend code and needs no compiler or runtime changes, so users can import it as a separate library.

This follows the same decision as [#1031, JSON parsing and encoding](https://github.com/bendlang/bend/pull/1031#issuecomment-5815267941), and [#884, a deque](https://github.com/bendlang/bend/issues/884#issuecomment-5746908694): both were directed to BendHub because they can be useful libraries without becoming part of Base. Publishing this there would let people use it now and let you extend the supported patterns without adding each extension to the core repository.

PR #938, opened 2026-09-21
Fix C macro name collisions

IlyaGulya:
The C backend turns Bend names into `CID_*` and `FID_*` macros by uppercasing them and replacing non-alphanumeric characters with `_`.

This is not one-to-one. For example:

```text
Thing.field  -> FID_THING_FIELD
thing_field  -> FID_THING_FIELD
```

Both are valid Bend names, but the C backend previously rejected the program because both names produced the same macro.

This change assigns C macro names once per compilation. When two ordinary Bend names collide, the later one gets a numeric suffix:

```text
Thing.field  -> FID_THING_FIELD
thing_field  -> FID_THING_FIELD_2
```

Some macro names cannot be renamed because they are written literally by the runtime or by foreign `.c` files. For example:

```c
io_eff(CID_IO_ARGS, ...);
return term_pak(CID_UNIT, 0);
```

Those names keep their original spelling. If two such fixed names require the same macro, the compiler reports an error instead of silently assigning the wrong ID.

Foreign `.c` includes also get temporary aliases from the names written in C to the globally allocated constructor IDs. This covers constructors from the foreign module itself and from `Base`.

For example:

```text
foreign C writes:  CID_OFF
allocated name:    CID_MODULE_OFF_2

CID_OFF -> CID_MODULE_OFF_2
```

The compiler also rejects ambiguous aliases and alias chains, since the C preprocessor would otherwise expand them to a different constructor.

The patch adds regression tests for:

* `.` / `_` name collisions;
* collisions with runtime constructor and segment names;
* generated internal segment names;
* constructors used by imported foreign C;
* root/Base constructor aliases;
* Base constructors used from user foreign effects.

The existing C spelling is preserved where it is part of the runtime/foreign-C interface; only compiler-owned names are renamed.


zxv:

nicolas-abril:
Thanks for working on this! This is a real bug and a tricky one to fix. Renaming source identifiers looks like a promising direction, and your contribution has helped clarify a possible path forward.

The implementation needs to be smaller and fit more closely into the existing compiler. In particular, this fix shouldn’t modify `bend.ts`, and the additional ~2,500 tokens exceed our gate budget. Ideally, it would also work within the existing compiler sections, with fewer comments explaining individual steps, consistent with the repo’s style.

Would you be interested in exploring a more compact implementation of the source-renaming approach? There’s a useful idea here, and simplifying the underlying logic could make it a good fit. Thanks again for putting time into this and moving the issue forward!

IlyaGulya:
> Thanks for working on this! This is a real bug and a tricky one to fix. Renaming source identifiers looks like a promising direction, and your contribution has helped clarify a possible path forward.
> 
> The implementation needs to be smaller and fit more closely into the existing compiler. In particular, this fix shouldn’t modify `bend.ts`, and the additional ~2,500 tokens exceed our gate budget. Ideally, it would also work within the existing compiler sections, with fewer comments explaining individual steps, consistent with the repo’s style.
> 
> Would you be interested in exploring a more compact implementation of the source-renaming approach? There’s a useful idea here, and simplifying the underlying logic could make it a good fit. Thanks again for putting time into this and moving the issue forward!
> 
> _Note: this reply was written by an AI after it reported the PR to me and I made the decision. If anything here is wrong, reply and I will review it myself._

Yeah, sure, I will try to make a more compact version 🙂

PR #940, opened 2026-09-21
refactor(comp): compact native ops and share match/JS program builders

This-Is-NPC:
`bend2/comp.ts` only. Same language, same Base, same `#|` contract.

The C and JS lanes repeated native op entries, Nat/U32 table matching, and the JS program wrapper. This collapses that duplication without changing emitted semantics.

- Native ops go through `both` / `js_op`; `OPERATIONS` and `OPTIMIZED` are the compact tables.
- C and JS table matches share `mat_tab`.
- `js_book` and `io_run` share `js_prog`.
- Fid/cid macros in the C runtime stay equivalent, with the blank lines dropped.

Net: **−121 lines** (`6591` → `6470`; `+92` / `−213`).


zxv:

nicolas-abril:
Thanks for the contribution! Reducing duplication and simplifying the compiler is a good direction, but I’m going to close this PR because several changes feel more like token minimization than improvements in clarity.

In particular, `both` and `js_op` save tokens without really simplifying the logic, and `both` doesn’t clearly describe its purpose. The changes around `OPTIMIZED` also make the code a little harder to follow. We want token savings to come from simpler code, rather than making the same logic more compact.

The `mat_tab` extraction looks promising, though I’d prefer an `emit_...` name and a simpler return type. I also like the direction of `js_prog`, but its current benefit seems small relative to the extra layer of abstraction.

You’re welcome to open a new PR focused on those two pieces, with cleaner interfaces and, ideally, some simplification of the underlying logic. Thanks again for taking the time to work on this!

PR #943, opened 2026-09-21
refactor(comp): share C/JS show_chr and list/tuple brackets

This-Is-NPC:
`bend2/comp.ts` only. Same language, same Base.

The C and JS showers repeated char-escape rules and list/tuple brackets.
One table now emits both.

- `SHOW_ESC` / `SHOW_HEX` emit `show_chr`. JS hex-escapes surrogates and
  code points past U+10FFFF the way C already did.
- `SHOW_BR` emits `Con`/`Nil` as `[]` and `Tuple` as `()`.

Net is about even (`+39` / `−37`). The point is one readback policy, not fewer lines.


zxv:

nicolas-abril:
While it is nice that we get the shared behaviour `show_chr_src` and `show_br_c` obscure the logic significantly and the code no longer follows the organization of the file, so I can't accept it as it is.

I think either hoisting only the part that is shared but keeping the C/JS skeletons there or applying just the C escape logic to JS would be good solutions

nicolas-abril (changes requested):
see comment above

Issue #944, opened 2026-09-21
C backend: wide monomorphic records bypass boxing and hit the 255-word CID limit


jasisz:
### What you did

I compiled this program to the interpreter, JS, and a native executable:

```python
import Base

type R0 is Data:
  R0{a: U32, b: U32, c: U32, d: U32, e: U32, f: U32, g: U32, h: U32}

type R1 is Data:
  R1{a: R0, b: R0, c: R0, d: R0, e: R0, f: R0, g: R0, h: R0}

type R2 is Data:
  R2{a: R1, b: R1, c: R1, d: R1, e: R1, f: R1, g: R1, h: R1}

def r0(+x: U32) -> R0:
  R0{x, x, x, x, x, x, x, x}

def r1(+x: U32) -> R1:
  R1{r0(x), r0(x), r0(x), r0(x), r0(x), r0(x), r0(x), r0(x)}

def r2(+x: U32) -> R2:
  R2{r1(x), r1(x), r1(x), r1(x), r1(x), r1(x), r1(x), r1(x)}

def first0(x: R0) -> U32:
  match x:
    case R0{a, b, c, d, e, f, g, h}:
      a

def first1(x: R1) -> U32:
  match x:
    case R1{a, b, c, d, e, f, g, h}:
      first0(a)

def first2(x: R2) -> U32:
  match x:
    case R2{a, b, c, d, e, f, g, h}:
      first1(a)

def main() -> U32:
  first2(r2(7))
```

```bash
bun bend2/main.ts repro.bend
bun bend2/main.ts repro.bend -o /tmp/repro.js
bun /tmp/repro.js
bun bend2/main.ts repro.bend -o /tmp/repro
```

### What happened

The interpreter and JS both print `7`, but the native C build fails before invoking clang:

```text
Error: an arity over 255
```

This reproduces on 2.0.23 as well, so it appears to be an uncovered case in the #843 fix rather than a 2.0.24 regression.

`R0` is 8 words, `R1` is 64 words, and `R2` is 512 words. Since #843, a datatype past 256 words is supposed to become a heap node, so `R2` should be boxed and compile.

The type layout does become `BOX`, but the constructor table is computed separately. `compile_book` stores `lay_node(fl.book, c.k).ks.length` in `CID_ARITY_T`; for a monomorphic `R2`, `lay_node` flattens the eight `R1` fields again and produces 512. `CID_ARITY_T` is emitted as `u8`, so `compile_tables` rejects it.

[...]

PR #945, opened 2026-09-21
Compute field offsets once in the layout packer

tachytelicdetonation:
`lay_fields` calculated field offsets before `lay_pack` mutated each offset to account for the constructor tag. This makes `lay_pack` accept constructor names and field layouts and assign the final offsets while merging word kinds. It removes the intermediate offset calculation and the caller-supplied offsets in `arr_lay`.

Only `bend2/comp.ts` changes: +17 / -27 lines (net -10), and 64,941 -> 64,881 ttok. No language features or comments are removed.

Validation against upstream commit `99f9c6cd` on macOS arm64:

- Compared CLI compilation of 263 inputs: every direct `.bend` file under `tests/{compile,run,base,show,rfc,cost}` plus all 16 runtime benchmarks. All 235 emitted C files and 235 emitted JavaScript files are byte-identical; the other 28 inputs have identical compilation diagnostics and exit status.
- Ran 16 representative layout programs on native C (one CPU thread, GPU off) and JavaScript: all 32 runs match their `#|` output. Four additional rejection tests match their expected errors and exit status.
- `gates/repo.ts`: PASS 46 / 46. `git diff --check`: pass.
- TypeScript 7.0.2 with the repository tsconfig reports the same four diagnostics before and after: two in `bend2/bend.ts` and missing `canvas` declarations in two documentation generators. No diagnostics in `comp.ts`.
- The private-cluster test, performance, and release gates were not run. Benchmark source generation was compared; runtime benchmark timings were not measured.


VictorTaelin:
Merged into 2.0.25 as one commit under your authorship, thank you. We verified the byte-identical emission over 914 outputs on our side. It landed together with the wide-record fix of #944, so `lay_pack` now also takes the field lists through `lay_wide`; the offsets are yours.

PR #947, opened 2026-09-21
Fix stale C helper calls in the effects guide

tachytelicdetonation:
The C examples for `io_node` and `io_wait_on` no longer match the runtime: the former has an extra argument, and the latter omits the deadline. Correct those calls and explain that a zero deadline waits only for descriptor readiness.

Validation: reproduced clang's argument-count errors against emitted C, then verified that the corrected calls compile. Also extracted and built the guide's complete Clock example and ran it through the default, emitted JavaScript, and native C paths; all three print the documented numeric result. `git diff --check` passes. No runtime code changes.


zxv:

PR #949, opened 2026-09-21
refactor(effs): share the audio ring through // @src

This-Is-NPC:
The C files for Audio.open, Audio.write and Audio.close each carried the same
157-line IO_RING block so any one of them compiled alone. The generated
program then pasted that block three times.

eff_src already skipped a path it had seen. A `// @src name.c` line now
loads a sibling through that same set, so the three effects share
`audio_ring.c` and a program that imports more than one still emits one
copy. A lone effect still compiles: it pulls the ring itself.

JS is unchanged (the silent stub has no ring). Isolated
`tests/io/audio_open.bend` passed interp, JS and C; the generated C has
`#ifndef IO_RING` once.

Repo: 545 → 231 lines in the audio effects, +2 in `eff_src` (−315 net).


VictorTaelin:
Thank you for this. The same duplication is gone in 2.0.25 through PR #939, which reaches the one-copy result with the realpath dedup `eff_src` already has, so the compiler gains no new directive. Closing this one in favour of that; the diagnosis was right and the work is appreciated.

PR #950, opened 2026-09-21
refactor(io): share file read and write implementations

tachytelicdetonation:
The three file readers repeat buffer setup and byte-list packing; the two writers repeat their write loop and completion handling. Keep those operations together in `file_read.{c,js}` and `file_write.{c,js}`, and remove the six redundant source files.

The C entry points remain guarded by their own CIDs so importing one operation does not require the others. Text decoding, positional reads, byte validation, error results, and handle ownership keep their existing behavior. Update Base's import paths and the source inventory accordingly.

Net: **48 fewer lines and 645 fewer ttok** across the changed files. No changes to the compiler or language kernel.

Verification:

- Nine existing IO tests pass through the default runner, emitted JS, and native C on both macOS arm64 and Linux arm64: `file_binary`, `file_roundtrip`, `read_bytes`, `write_nul_bytes`, `fail_keeps_handle`, `read_leading_bom`, `path_utf8`, `codec_seal_shared`, and `tcp_send_slow_peer` (54 checks). The seven direct file tests also passed before the refactor.
- Built and ran all 31 nonempty combinations of the five operations in C and JS on both platforms (124 checks), covering individual-effect imports and shared-source deduplication.
- All 16 runtime benchmarks generate byte-identical C and JS versus `99f9c6cd`. Repository file/token gate: 46/46; `git diff --check` passes.

The private mini-cluster test and performance gates were not available; no cluster timing results are claimed.


VictorTaelin:
Merged into 2.0.25 under your authorship, thank you: −641 ttok in the effects, every file test identical on the three lanes, and each entry still guarded by its own CID.

PR #951, opened 2026-09-21
fix(js): keep foreign effects and scheduling helpers together

### LINKED ISSUE #946: JS FFI: an effect named _need silently replaces another effect's scheduling helper
### What you did

I defined two independent foreign effects whose documented JS names are `a_run` and `a_run_need`. The latter is also the documented scheduling-helper name of the former.

`main.bend`:

```python
import Base

def A.run(ms: U32) -> IO(U32):
  import "./a.c"
  import "./a.js"

def A.run_need() -> IO(U32):
  import "./b.c"
  import "./b.js"

def main() -> IO(Unit):
  do IO<Unit>:
    y : U32 <- A.run_need()
    x : U32 <- A.run(1)
    IO.print(U32.show(x) ++ " " ++ U32.show(y))
```

`a.js`:

```js
function a_run(ms) {
  return 11;
}

function a_run_need() {
  return { time: true };
}
```

`b.js`:

```js
function a_run_need() {
  return 22;
}
```

The equivalent C implementations use distinct entry points and CIDs:

`a.c`:

```c
Term a_run_run(Env e, Term* f, IoWork* w) {
  return (Term)11;
}

static void __attribute__((constructor)) a_run_use(void) {
  io_eff(CID_A_RUN, a_run_run, IO_TIME);
}
```

`b.c`:

```c
Term a_run_need_run(Env e, Term* f, IoWork* w) {
  return (Term)22;
}

static void __attribute__((constructor)) a_run_need_use(void) {
  io_eff(CID_A_RUN_NEED, a_run_need_run, 0);
}
```

```bash
bun bend2/main.ts main.bend
bun bend2/main.ts main.bend -o /tmp/out.js
bun /tmp/out.js
bun bend2/main.ts main.bend -o /tmp/out
/tmp/out
```

### What happened

The default lane and emitted JS exit successfully but print `11 ` followed by a NUL character. The native C binary prints the expected `11 22`.

`eff_name("A.run")` is `a_run`, so its two JS table keys are `a_run` and `a_run_need`. `eff_name("A.run_need")` is itself `a_run_need`. `js_lib` adds only each effect's base name to the duplicate-check array `ms`, while also emitting `n + "_need"` into `$0eff`; it therefore misses this collision.

Each foreign source is correctly isolated in its own closure after #800, but the returned objects are spread into one `$0eff` object. In the shown reachability order, the `a.js` closure is spread last, so its scheduling helper overwrites `b.js`'s effect function. `A.run_need()` then returns `{ time: true }` where the checked Bend program expects a `U32`, and `U32.show` renders the corrupted value as NUL.

Reversing the two effect calls reverses the closure order: the program happens to print `11 22`, but `b.js`'s `a_run_need` then overwrites `A.run`'s scheduling helper, so `A.run` runs immediately instead of honoring `{ time: true }`. Reachability order decides which wrong behavior occurs.

This is adjacent to #799 and #800 but is not rejected by their fixes. It reproduces on 2.0.23 and 2.0.24.

### Expected

The JS emitter should reject the duplicate emitted key, for example with `two names mangle to a_run_need`, or represent effect functions and scheduling helpers in separate namespaces. Adding both `n` and `n + "_need"` to the collision set would reject this case before emitting wrong code.

### Environment

```text
bend 2.0.24 (99f9c6cd)
Bun 1.4.2
Darwin arm64
Apple clang version 21.0.0 (clang-2100.0.123.102)
```

tachytelicdetonation:
Fixes #946.

An effect named `A.run_need` can overwrite `A.run`'s scheduling helper in the flat JS foreign table, or be overwritten by it. Store one `{run, need}` record per effect so both functions retain their module's identity regardless of discovery order.

Use computed property keys so `__proto__` is also stored as an own effect entry, covering the additional reproduction reported by @jasisz on the issue. Existing module isolation and duplicate effect-name checks remain in place.

The compiler change is **4 lines added / 5 removed** (+12 ttok). Three regression cases cover both discovery orders and `__proto__`; the scheduling fixture checks that its helper was invoked without relying on elapsed-time assertions.

Verification:

- Both collision orders failed before the fix; `__proto__` also failed before the computed-key change. All three now pass through the default runner and emitted JS.
- 13 focused tests pass on macOS arm64 and Linux arm64 (34 checks per platform), including shared module state, runtime-name isolation, foreign arity/types, timers, and parked TCP/UDP effects. Native C checks run where fixtures provide a C implementation.
- All 16 runtime benchmarks emit identical C; their emitted JS differs only in the intended foreign-table/request wiring. Repository file/token gate: 46/46; `git diff --check` passes. TypeScript diagnostics match the baseline's four existing errors.

The private mini-cluster test and performance gates were not available. JS checks used Bun; no cluster timing results are claimed.


VictorTaelin:
Merged into 2.0.25 under your authorship, thank you. Both discovery orders and the `__proto__` effect now print the right values on all lanes; we audited every reader of the foreign table against the `{run, need}` record. This closes #946.

PR #952, opened 2026-09-21
refactor(compiler): group native constructor operations into descriptors

ramonzx6:
## Why the change

The JavaScript emitter stores native constructor builders, field readers, and match conditions in separate maps. This repeats constructor names and makes additions or renames easy to apply inconsistently. Keeping these related operations together makes each constructor's encoding easier to review and maintain.

## What changed

Replace the parallel maps in `bend2/comp.ts` with one descriptor per native constructor and update the lookups for construction, field access, and matching. Each descriptor requires a builder, with optional readers and conditions; `satisfies` checks the registry literal before `Object.setPrototypeOf`. Generated expressions, the outer null prototype, existing diagnostics, and field-count checks remain unchanged.

## Validation

- 1,413 checker/interpreter tests and 639 JavaScript tests pass.
- 1,294 emitted C/JS files are byte-for-byte identical; 258 rejection diagnostics match.
- Repository gate: 46/46. Standalone CLI build passes.


VictorTaelin:
Merged into 2.0.25 under your authorship, thank you: one record per native constructor, byte-identical emission, −48 ttok.

PR #960, opened 2026-09-21
perf(compiler): count facts to detect convergence

byronbenharris:
### Why the change

  The compiler checks whether a pass found new facts by converting three sets to arrays and serializing them. This repeats
  work after every pass. Because the sets only grow, their combined size is enough to detect a change.

  ### What changed

  In bend2/comp.ts, compare the combined sizes of own, hot, and stat to decide whether another pass is needed. Update the
  comparison type and the nearby comment. The passes and generated code are unchanged.

### Validation 

  - Checker/interpreter: 1,412 of 1,413 gate tests matched expected output. The remaining TCP test printed its expected
    output but did not exit within 20 seconds.
  - Parent-commit comparison: 350 of 763 runnable cases checked so far; 564 generated C/JS files byte-identical, 68
    rejection results matched, and zero mismatches.
  - Focused C and JS runs passed for hot facts, ownership, and array forks.
  - The official gates not run yet: the cluster host is unavailable, ttok is missing, and the installer’s sibling
    site repo is absent.


VictorTaelin:
Merged into 2.0.25 under your authorship, thank you. We checked the premise: `own`, `hot` and `stat` only ever grow across passes (every `.delete` in the compiler is on another map), so the summed size is a correct convergence test.

PR #961, opened 2026-09-21
Initialize compiler caches before building printer layouts

vicmcorrea:
## Why

i'm evaluating Bend for small survey-scoring helpers in our TypeScript monorepo. While testing repeated compiler calls, I found that two valid books can crash when their datatype names overlap. A flat type followed by a recursive type with the same name overflows the stack. Both compile separately.

This affects repeated `compile_book` / `js_book` calls. Ordinary CLI runs and the Node/Bun import path are not the failing case.

## What changed

this moves `show_main` after the existing per-book initialization in the C and JS entry points. It was reading `NODES` and `CYCLES` from the previous book before `carb_book` cleared them. One file, +3/-2 lines.

## Validation

Checked on macOS arm64 with Bun 1.3.11 and clang 23.1.1:

- 25 compilation sequences covering both backends, both book orders, repeated books and changed recursion. Patched output matches fresh baseline output byte for byte. Eight generated C/JS executions print the expected values.
- 98 existing fixtures across both backends. All 166 generated files and 30 rejection diagnostics match main.
- Repository gate passed, 46/46. `git diff --check` passed. Cluster and GPU tests were not run.

## Reproduction

<details>
<summary>Reproduction, run from the checkout root</summary>

The current test gate compiles each fixture in a separate process, so it would miss this. This repro keeps both books in one process; it is not an automated regression test.

```sh
BEND_NO_TELEMETRY=1 bun -e '
import * as fs from "node:fs";
import * as os from "node:os";
import * as path from "node:path";

const Bend = await import(path.resolve("bend2/bend.ts"));
const Comp = await import(path.resolve("bend2/comp.ts"));
const dir = fs.mkdtempSync(path.join(os.tmpdir(), "bend-cache-repro-"));
try {
  for (const [ctors, value] of [
    ["Leaf{value: U32}", "Leaf{7}"],
    ["Leaf{value: U32}\n  Branch{tail: Tree}", "Branch{Leaf{7}}"],
  ]) {
    const file = path.join(dir, "main.bend");
    fs.writeFileSync(file, `import Base\n\ntype Tree is Data:\n  ${ctors}\n\ndef main() -> Tree:\n  ${value}\n`);
    const book = Bend.book_nil();
    await Bend.book_load(book, file, "", new Map());
    Bend.book_valid(book);
    Comp.compile_book(book);
    Comp.js_book(book);
    console.log("compiled " + value);
  }
} finally {
  fs.rmSync(dir, {recursive: true, force: true});
}

'
```

On main, the first book compiles and the second throws `RangeError: Maximum call stack size exceeded`. With this change, both lines print:

```text
compiled Leaf{7}
compiled Branch{Leaf{7}}
```

</details>


vicmcorrea:
bugbot run

VictorTaelin:
Merged into 2.0.25 under your authorship, thank you. We reproduced both orders: flat-then-recursive overflowed the stack, and recursive-then-flat silently emitted the previous book's layout, which is worse than the PR says. `show_main` now runs after the memos are reset.

PR #963, opened 2026-09-21
One BEND_RTC macro for the six __CUDACC_RTC__ tests, and gpu_hash above the lanes

costamatheus97:
The two pieces of the HIP lane (#958, merged to `hip`) that do not depend on it, as offered in that thread. One commit on `main` at e52cda47, `bend2/comp.ts` only, +23 −14. No behaviour change.

## What it changes

- **`BEND_RTC`.** Defined when the file is being compiled by the device's runtime compiler (`#if defined(__CUDACC_RTC__)`). The six `__CUDACC_RTC__` tests read it instead. A second RTC lane adds its macro to that one line and leaves the six sites alone.
- **`gpu_hash`.** Moved out of the CUDA block to just above the lane blocks, under `#if BEND_CUDA`. It reads only `BEND_SRC` and `CUBE_LOG`, both defined earlier. A lane that keeps a `.gpu` cache shares it by widening that guard. `hip` already builds and runs with the function at this position.

With both on `main`, `hip` differs from `main` on these lines by `|| defined(__HIPCC_RTC__)` and `|| BEND_HIP`, which should keep its rebases quiet in shared code.

## Checks

I have no CUDA or Metal hardware, so the main check is on the preprocessor. I emitted C for `tests/run/gpu_mark.bend` from pristine `main` and from this branch, removed the `#embed __FILE__` line from both (the embedded text differs by construction), and compared `clang -E -P` output:

| configuration | result |
| - | - |
| host, no GPU lane | identical |
| device, `-D__CUDACC_RTC__` | identical |
| device, `-D__METAL_VERSION__` | identical |
| host, `-DBEND_CUDA=1` (stub `cuda.h` / `nvrtc.h`) | identical except `gpu_hash` appears earlier |

- A one-machine copy of `gates/test.ts`'s rules over `gpu_*`, `compile_*`, `io_hello*`, `gfx_*`: 120 / 122 on this branch and 120 / 122 on pristine `main`. The two failures are the same on both, `gfx_clicks` and `gfx_window`, which need a display.
- `comp.ts` is 64967 tokens against the 65000 cap. I do not have `ttok` here; I counted with `cl100k_base`, which gives exactly your 64941 for `main` and 68028 for `hip`.
- Not run: the gates themselves (`gates/repo.ts` needs `ttok`), the full test gate, anything on CUDA or Metal hardware.


VictorTaelin:
Merged into 2.0.25 under your authorship, thank you. We confirmed the preprocessed C is identical on the host, under CUDA RTC and under Metal, and the Metal tests pass here. `hip` should now differ from main on those lines only by its own `|| defined(__HIPCC_RTC__)` and `|| BEND_HIP`.

PR #969, opened 2026-09-21
Share window effects BendWin through // @src

ewoodbury:
The four Window C effects each carried the X11 `BendWin` struct (with includes) so any one of them compiled alone. A program that imports more than one then pasted that block four times, under an `#ifndef BendWin` concatenate-guard.

`eff_src` already skipped a path it had seen. A `// @src name.c` line now loads a sibling through that same set, so the four effects share `window_host.c` and a program that imports more than one still emits the type once.

This is the same `@src` helper as #949 (audio ring). If that lands first, the `comp.ts` hunk here is identical and drops on rebase

Checks

- `tests/gfx/*.bend` `--check-only`: all check.
- `tests/gfx/window.bend` interpret and JS: same `#|` pin as main (no display / ENOTSUP).
- Emitted C for `window.bend` (open+frame+close) and a scratch file that also calls `set_title`: one `BendWin` typedef, one `<X11/Xlib.h>`, no leftover `// @src`.
- `clang -O1` of those programs with `-lX11 -lpthread -lm`: exit 0.

Not run: the cluster gates, Metal, a machine with a display.

Runtime, checker, and JS are untouched. Window behavior is unchanged


zxv:
Rescuing PR by commenting (github issue)

bend2/comp.ts
```diff
@@ -1235,7 +1235,9 @@ function eff_src(path: string, seen: Set<string>): string {
     return "";
   }
   seen.add(path);
-  return fs.readFileSync(path, "utf8");
+  const dir = path.slice(0, path.lastIndexOf("/"));
+  return fs.readFileSync(path, "utf8").replace(/^\/\/ @src ([a-z0-9_]+\.c)$/gm,
+    (_: string, name: string) => eff_src(dir + "/" + name, seen));
```
paulovitin, line 1240: [P2] Include transitive `@src` files in published packages. `pkg_files()` in `main.ts` only collects direct Bend imports, so an effect containing `// @src helper.c` builds locally but leaves `helper.c` out of the package. I reproduced this by rebuilding from the files returned by `pkg_files()`; compilation fails with `ENOENT` for the helper. The package collector needs to follow these dependencies too.

paulovitin, line 1240: [P2] Protect these helpers from output overwrite. The input set in `cli_file()` only tracks direct imports, so `bend main.bend -o helper.c` exits 0 and replaces a helper referenced through `// @src helper.c` with generated C. I reproduced the overwrite with a working custom effect. Add transitive helpers to the protected input set before emitting any output.

PR #974, opened 2026-09-21
feature/js-plains

MattCozendey:
# The JS lane stops paying for calls: selector intrinsics + plain defs off the trampoline

Two changes to `bend2/comp.ts`, both invisible to the C emitter, the device
and the checker (no base.bend, bend.ts or Lean surface is touched).

## 1. Selector intrinsics (b4ebc8bf)

JS-only `OPERATIONS` entries for `bool_and`, `bool_not`, `bool_pick`,
`u32_min`, `u32_max`, `f32_min`, `f32_max`, `nat_min`, `nat_max`.

Before, `U32.min(a, b)` emitted a trampolined call into `$U32$min$`, which
trampolined again into `$Bool$pick$` — three frames and two `$JMP`
allocations for one compare. Now the ternary emits in place.

Each template mirrors its base.bend body exactly:

- `pick c a b = c ? a : b`, `min = a < b ? a : b`, `max = a < b ? b : a`
- F32 NaN answers keep following `is_lt`, as `Bool.pick` of `F32.is_lt` does
- `Nat.min/max`'s Succ-peeling recursion (O(min(a,b)) BigInt steps) becomes
  one BigInt compare
- `bool_pick` stays JS-only because its `A` is polymorphic; on native lanes
  no single word shape fits

## 2. Plain defs are called directly (8ff7eab8)

A def is *plain* when its emitted JS can never return a `$JMP`: no tail
self-call and no closure applied in tail position. Every other return comes
back through the machine stack, whose depth the def call DAG bounds — Bend
has no mutual recursion and no forward calls.

`js_call` now calls a plain callee directly (tail or not), so the JIT can
inline it. A self tail call keeps `run_jump`; a tail call into a trampolined
callee absorbs its `$JMP`s with `run_loop` in place, costing one bounded
frame. `plain_of` reads the property off the body with `term_any`'s tail
flag; a closure body counts as tail, which only errs toward keeping the
trampoline.

## Measurements (WSL, bun, medians of three, identical outputs and RSS)

| workload (10M iterations)                | before | after  | speedup |
|------------------------------------------|--------|--------|---------|
| Base selector ops (min/max/pick/is_even) | 1.30 s | 0.51 s | ~2.5x   |
| chain of three small user defs per step  | 1.26 s | 0.49 s | ~2.6x   |
| single pure Nat tail loop (control)      | 0.98 s | 0.94 s | ~1.0x   |

The control row is expected: a def that never calls anything gains nothing;
its costs (per-iteration `run_jump` allocation, BigInt Nat arithmetic) are
follow-up work, not this PR.

## Verification

- A scratch program exercising every intrinsic through constant and
  variable arguments prints the same value interpreted and emitted.
- 73 tests across `tests/base`, `tests/eval` and `tests/io` pass on both
  the run path and the emitted-JS path.
- The cluster gates did not run (no ssh access from this machine); the
  perf pins and the device lane still need the usual gate pass.


jeromelaurens:
Since 2.0.17 (#860, 4033e90c) `Nat.min`/`Nat.max` peel successors and neither lane has an intrinsic for them, so they cost O(min(a, b)). This PR fixes the JS lane; the native lane still crashes:

```python
import Base

def main() -> IO(Unit):
  do IO<Unit>:
    IO.print(Nat.show(Nat.min(Nat.mul(1000000000n, 1n), Nat.mul(1000000000n, 2n))))
```

`bend min.bend -o min && ./min` prints `bend: memory fault (machine stack overflow?)` on 2.0.27 (10^8 works) and `1000000000` on 2.0.16. Would C templates for `nat_min`/`nat_max` (`($0 < $1 ? $0 : $1)`, `($0 < $1 ? $1 : $0)`) in this PR cover both lanes? Found in a payroll POC, where amounts in micro-euros exceed 10^9. — bend 2.0.27, Darwin arm64

MattCozendey:
You were right, thanks. Since #860 made `Nat.min`/`Nat.max` peel successors, the C lane had been running that recursion as written, one continuation frame per successor, and clang can't turn it back into a compare. I added the C templates you suggested (036e0b47), plus C templates for the other selectors this PR had made JS-only, so both lanes get the same treatment. `bool_pick` still has no C template: the branch it doesn't take may own a heap value the C lane has to drop, so it keeps its match. Your program now prints `1000000000` on both lanes.

WSL, clang 18 -O3 and bun, medians of three, identical outputs; before is `main`:

| workload (10M iterations)                        | lane | before                | after              | speedup |
|--------------------------------------------------|------|-----------------------|--------------------|---------|
| your repro: `Nat.min(10^9, 2*10^9)`, one call    | C    | memory fault (2.0 GB) | 1000000000, 1.9 MB | fixed   |
| your repro: `Nat.min(10^9, 2*10^9)`, one call    | JS   | memory fault          | 1000000000         | fixed   |
| `Nat.min(p, 100)` + `Nat.max(p, 50)` per step    | C    | 1.39 s, 2.0 MB        | 6.0 ms, 2.4 MB     | ~230x   |
| `Nat.min(p, 100)` + `Nat.max(p, 50)` per step    | JS   | 81.2 s, 56.4 MB       | 0.59 s, 39.0 MB    | ~138x   |
| U32/F32 min/max, `Bool.and`/`not` per step       | C    | 8.6 ms, 2.5 MB        | 7.2 ms, 1.9 MB     | ~1.2x   |
| U32/F32 min/max, `Bool.and`/`not` per step       | JS   | 2.13 s, 44.2 MB       | 0.61 s, 40.8 MB    | ~3.5x   |

The details are in the updated PR description, under section 3 and the new measurements table.

nicolas-abril (changes requested):
Thanks, this is a good change, and the measurements hold up here. We'd like to merge it once the one issue below is fixed and tests are added.

**The intrinsics are correct.** Every template matches its `base.bend` body, including F32's NaN and ±0 behaviour. On a cross-lane program covering NaN, ±0, the U32 extremes, every `Bool` combination, `Bool.pick` at `String` and `Nat`, and `Nat` near 10^12, this branch prints exactly what `main` prints on the interpreter, JS and C wherever `main` finishes. Where `main` overflows the stack, it prints the right values.

**The issue: plain defs can recurse on the JS stack.** The plain-def argument relies on "Bend has no mutual recursion and no forward calls". Since `c64e021d`, an `@unsafe` def can call a def written below it, and `tests/run/unsafe_forward.bend` has two that recurse into each other. Two defs that tail-call each other then both count as plain, so they call each other directly, one JS frame per hop. A tail call into a non-plain callee also became `run_loop(call)` in place of `run_jump`, which nests a frame per call and so also grows on a cycle.

```bend
import Base

@unsafe
def even(n: Nat) -> Bool:
  match n:
    case 0n:
      True{}
    case 1n+p:
      odd(p)

@unsafe
def odd(n: Nat) -> Bool:
  match n:
    case 0n:
      False{}
    case 1n+p:
      even(p)

def main() -> IO(Unit):
  do IO<Unit>:
    t : Nat <- IO.now()
    IO.print(Bool.show(even(Nat.add(1000000n, Nat.mul(Nat.mod(t, 1n), 1n)))))
```

`main` prints `True` on every lane. On this branch the JS lane and `bend file.bend` stop with `memory fault (machine stack overflow?)`; the C lane still prints `True`. It passes at 20,000 and fails at 100,000.

**A fix we tested.** A def is plain only if it is foreign, or every tail call in it reaches a plain def. A def that is still being evaluated counts as not plain, so any call cycle is not plain. Every non-plain callee keeps `main`'s behaviour, and `fl.def` is no longer needed:

```ts
// A def is plain when its JS returns no $JMP: a foreign def, or one whose
// tail calls all reach plain defs. A tail self call, a tail closure and a
// def on its own call cycle (unsafe defs call forward) ride the
// trampoline, so a direct call never recurses through the machine stack.
function plain_of(c: Carb, k: Bend.Name): boolean {
  const got = PLAINS.get(k);
  if (got !== undefined) {
    return got;
  }
  PLAINS.set(k, false);
  const tld = def_body(c, k);
  const out = tld?.$ === "Def" && (def_foreign(tld) || tld.h !== undefined
    && !term_any(c, tld.h as HTerm, (s, tail) => {
      const ck = tail ? call_kind(c, s) : null;
      return ck !== null && (ck.k === CLO_APPLY || !plain_of(c, ck.k));
    }));
  PLAINS.set(k, out);
  return out;
}
```

```ts
  return v !== "" ? "(" + v + ") => " + call
    : plain_of(fl, k) ? call
    : tail ? "run_jump(" + js_sat(k) + ", [" + exprs.join(", ") + "])"
    : "run_loop(" + call + ")";
```

With it, the program above prints `True` on every lane and `unsafe_forward` still passes. The gains are kept on the JS lane (medians of three, bun, 10M steps):

| workload | `main` | this branch | with the fix |
| --- | ---: | ---: | ---: |
| chain of three small defs per step | 0.71 s | 0.30 s | 0.30 s |
| U32 min/max, `Bool.pick`/`and`/`not`, `Nat.min` per step | ~41 s | 0.44 s | 0.44 s |
| a wrapper that tail-calls a self-recursive loop | 0.45 s | 0.37 s | 0.45 s |

The last row is the only cost: the in-place `run_loop` is what makes that shape faster, and it is the part that is unsafe on a cycle. Recovering it would take a real call-cycle analysis, which we don't think is worth it.

**A smaller simplification.** The six min/max entries are one pattern with a different comparison, so a helper can write them once:

```ts
// min and max as Base writes them, a Bool.pick over is_lt: a tie or a NaN
// answers min's second argument and max's first.
function sel_ops(pre: string, C: string): Record<string, Intr> {
  return {
    [pre + "min"]: { C: `(${C} ? $0 : $1)`, JS: "($0 < $1 ? $0 : $1)" },
    [pre + "max"]: { C: `(${C} ? $1 : $0)`, JS: "($0 < $1 ? $1 : $0)" },
  };
}
```

```ts
  ...sel_ops("u32_", "U32_BIN($0, <, $1)"),
  ...sel_ops("f32_", "f32_unbox($0) < f32_unbox($1)"),
  ...sel_ops("nat_", "$0 < $1"),
```

For U32 this returns the original word, as `Bool.pick` in Base does, and it prints the same values on all three lanes.

**Tests.** The branch adds none. Please add:
- `tests/base/nat_min_max.bend` from #1037, which we're closing in favour of this PR (Nat.min/max at 0, equal values, both orders and pairs near 2^40, through a template and a closure);
- a cross-lane intrinsics test for F32 NaN and ±0, the U32 extremes and every `Bool` combination;
- the mutual recursion above, so the JS lane keeps this depth.

Merged onto current `main`, the branch passes `gates/repo.ts` (46/46), and a local run of `gates/test.ts` gives 1439/1441, failing only the two display tests `gfx_clicks` and `gfx_window`. With the fix and the helper applied the result is the same. The cluster and GPU gates were not run.

**A further suggestion, optional.** The C lane compiles a def's self tail call as a loop (`WL_AGAIN` in its helper), but the JS lane emits `return run_jump($f$, [...])` on every iteration. If `js_def` compiled a self tail call as a loop instead, a def could only return a `$JMP` through a closure in tail position or through a cycle of unsafe defs. `plain_of` would shrink to that check, nearly every def would be plain (loop defs included), and the per-iteration `$JMP` allocation behind your control row would go away. Hand-editing the emitted JS of a benchmark whose loop tail-calls a wrapper around a small self-recursive loop (3M steps) gave:

| emitted JS | time |
| --- | ---: |
| as this branch emits it | ~0.43 s |
| the inner loop compiled as a loop | ~0.25 s |
| both loops compiled as loops | ~0.17 s |

One thing to watch: a loop must give each iteration fresh bindings for its parameters, or a lambda created in the body would see a later iteration's values.

Take this only if it pays for itself in size: the whole change should end up smaller than this branch's patch. Measured with `ttok`, as `gates/repo.ts` does, this branch adds 630 tokens to `comp.ts` (62215 to 62845), and the fix and helper above bring that to 522. If the loop emission plus the smaller `plain_of` fit under 630, it is worth doing here; otherwise, leave it for a follow-up.

*Note: this review was written by an AI after it reported the issue to me and I made the decision. If anything here is wrong, reply and I will review it myself.*

MattCozendey:
Updated as requested!

nicolas-abril:
@MattCozendey, following up on the review: we tried the loop suggestion and put the result in a draft, #1040. It contains your three commits unchanged, plus one commit on top:

- **A tail self call is a loop on the JS lane**, as the C lane's flat helpers loop. Each turn gets fresh parameter bindings, so closures keep their own turn's values.
- **Plainness is read off the same walk as C's flatness.** `carb_book` records each def's tail callees in its `Src`, and `plain_of` reads them with `flat_of`'s cycle guard. The `PLAINS` map and `plain_of`'s own walk go.
- **The mutual-recursion overflow from the review is fixed**, and the case the review called the only cost, a tail call into a self-recursive loop, gets faster than on this branch rather than slower.

On the JS lane it runs about 2 to 2.7 times faster than this branch on every benchmark we tried, and it adds 579 tokens to `comp.ts` against 630 here. The numbers and the checks are in its description.

It's your PR, so tell us which way you'd like to go: pull the commit into this branch, or we continue on #1040 and close this one. Tests are still missing on both. The list from the review (#1037's `nat_min_max.bend`, a cross-lane intrinsics test and the mutual recursion) applies to whichever lands.

*Note: this comment was written by an AI after it reported the issue to me and I made the decision. If anything here is wrong, reply and I will review it myself.*

MattCozendey:
Thanks, Nicolas. #1040 is the better base. It is smaller, and it reuses the `Src` walk. We can close #974.

I opened #1043. It contains #1040 unchanged, plus 2 commits:

1. **Tests.** It adds the 7 tests from the review: `nat_min_max`, `selector_intrinsics`, the two unsafe tail cycles, closure capture, a tail call through a closure, and mutual jumps across arities.
2. **Mutual tail recursion runs as a loop.** A Tarjan pass finds cycles in the `tails` sets that `carb_book` already records. `plain_of` reads a cycle as one def, so callers still call it directly.
3. **Intrinsic tail calls stay plain.** Intrinsic callees (`U32.add`, `Nat.min`, …) were recorded as tail callees, so a def that ended in one used the trampoline. A one-line change fixes this.

JS lane, compared to #1040. The outputs are identical, and the memory use is the same:

| case | #1040 | #1043 |
|---|---|---|
| 3-def tail cycle | 543 ms | 71 ms |
| 2-def mutual recursion | 578 ms | 182 ms |
| wrapper that calls a self-loop | 48 ms | 46 ms |
| other cases | same | same |

The tests pass on the interpreter, JS and C. I did not run the cluster gates.

If #1043 is good for you, it can replace both #974 and #1040.

PR #976, opened 2026-09-21
fix(compiler): reset probes between compilations

vicmcorrea:
## Why

i'm evaluating Bend for small survey-scoring helpers in our TypeScript monorepo. While testing rebuilds with the Bun plugin, I noticed compiler objects accumulating even after GC. `PROBES` keeps the binder variables from every previous compilation.

## What changed

this resets `PROBES` in `carb_book` after the existing cache clears, keeping `DUMMY` at index zero. Probes stay valid through emission. The per-definition `memo_gc` would reset them too early. One file, one added line.

## Validation

Checked on macOS arm64 with Bun 1.3.11:

- 250 Bun builds in one process. After warmup, main retains 250 more objects per 50 builds. The patched count stays flat. Every bundle is identical and the scoring function returns `22`.
- 99 existing fixtures across C and JS. All 168 generated files and 30 rejection diagnostics match main. Another 96 mixed compilation operations, including empty roots, also match.
- Also checked with #974 at `c03b845`. The changes merge cleanly, and 250 Bun builds still return `22`. From build 50 to 250, #974 alone retains 1,000 more objects; with this fix the count stays flat.
- Repository gate passed, 46/46. `git diff --check` passed. Cluster and GPU tests were not run.

## Reproduction

<details>
<summary>Reproduction, run from the checkout root</summary>

The current test gate uses separate compiler processes, so it cannot catch this accumulation. This executable repro uses repeated Bun builds instead of adding a new test harness.

```sh
BEND_NO_TELEMETRY=1 bun - <<'TS'
import {heapStats} from "bun:jsc";
import * as fs from "node:fs";
import * as os from "node:os";
import * as path from "node:path";

const {default: bend} = await import(path.resolve("bend2/main.ts"));
const dir = fs.mkdtempSync(path.join(os.tmpdir(), "bend-probes-"));
try {
  fs.writeFileSync(path.join(dir, "score.bend"),
    "import Base\ndef score(x: U32) -> U32:\n  (x * 3 + 7 : U32)\n");
  fs.writeFileSync(path.join(dir, "entry.ts"),
    "import scoring from './score.bend'; export const result = scoring.score(5);\n");
  const build = async () => {
    const result = await Bun.build({entrypoints: [path.join(dir, "entry.ts")],
      plugins: [bend], target: "bun", write: false});
    if (!result.success) throw new AggregateError(result.logs);
    return result.outputs[0].text();
  };
  const expected = await build();
  for (let i = 0; i < 10; i++) await build();
  for (let batch = 0; batch <= 5; batch++) {
    if (batch > 0) {
      for (let i = 0; i < 50; i++) {
        if (await build() !== expected) throw new Error("bundle changed");
      }
    }
    await new Promise((resolve) => setTimeout(resolve, 0));
    Bun.gc(true);
    console.log(batch * 50, heapStats().objectTypeCounts.Object);
  }
  const mod = await import("data:text/javascript;base64," + Buffer.from(expected).toString("base64"));
  if (mod.result !== 22) throw new Error("wrong result");
  console.log("result", mod.result);
} finally {
  fs.rmSync(dir, {recursive: true, force: true});
}
TS
```

The columns are rebuilds and retained `Object` instances after GC. Absolute counts can vary. On main the count keeps growing, while the patch levels off:

```text
main              patched
50 84608          50 84308
100 84858         100 84308
150 85108         150 84308
200 85358         200 84308
250 85608         250 84308
```

Both print `result 22`. The fix bounds probe retention to the latest compilation; it does not release all compiler state immediately after a build.

</details>


vicmcorrea:
bugbot run

PR #977, opened 2026-09-21
TCP.listen's backlog is SOMAXCONN, not 16: a burst of 10k connections is answered in full

aldeni:
## What changes

`bend2/effs/tcp_listen.c` calls `listen(fd, SOMAXCONN)` instead of
`listen(fd, 16)`. The JS lane (`bend2/effs/tcp_listen.js`) cannot see the
macro, so it carries the platform's value the same way the file already
carries the numeric `SO_REUSEADDR` constants: 4096 on Linux (glibc's
`SOMAXCONN`, and the kernel's default `net.core.somaxconn`), 128 on macOS.
The kernel caps the backlog at `somaxconn` either way, so `SOMAXCONN` means
"as deep as this system allows", with no magic number of our own.

A new test, `tests/io/tcp_listen_backlog.bend`, listens and then connects
32 times in a row without ever accepting: with a backlog of 16 the
eighteenth SYN is dropped and that connect stalls on retransmits, so a
regression shows as a hang rather than a wrong line; with `SOMAXCONN` all
32 complete at once and the test prints its `#|` line.

## Why

A backlog of 16 is the accept queue's depth. Once 16 handshakes are
waiting for `TCP.accept`, the kernel drops further SYNs and the clients
retransmit after 1 s, then 3 s -- the connection is not refused, it is
silently delayed, and under a burst most of them time out on the client's
side before the server gets to them.

Measured on Lemah (a Bend API framework) with the same `examples/crud`
server built twice from this tree, once with each backlog, on one 4-core
Linux machine, 10,000 connections opened at once each sending one
`GET /items`, 10 s window, one run each:

| backlog | answered | timed out | last answer | probe p99 |
|---|---|---|---|---|
| 16 | 8,221 / 10,000 | 1,779 | 9.27 s | 1,025 ms |
| `SOMAXCONN` (4096) | **10,000 / 10,000** | **0** | **3.30 s** | 156 ms |

The probe is a separate client opening one fresh connection every 50 ms;
its 1 s p99 with backlog 16 is the SYN retransmit timer. `ss -ltn` shows
the listener's `Send-Q` go from 16 to 4096. An earlier, longer sweep on
the same machine (1k/5k/10k, idle/slowloris/burst, against Bun as the
reference with its backlog of 512) found the same signature in every
scenario; this PR is its smallest fix.

## Alternative considered

A fixed larger constant, as Node (511) and Bun (512) use. `SOMAXCONN` was
preferred because it is the identifier the platform defines for exactly
this, the kernel clamps to it anyway, and it leaves no number to revisit.
Making the backlog a parameter of `TCP.listen` was not attempted: the
signature is one port today, and no caller of `TCP.listen` in `tests/`
has a reason to ask for a shallower queue.

## Test output

`tests/io/tcp_listen_backlog.bend`, all four lanes on this tree
(`bun bend2/main.ts <file> [--check-only | -o x.js | -o x]`):

```
All terms check.
32 connected, none accepted
```

Negative check: the same test run with the released `bend` 2.0.22, whose
runtime still has `listen(fd, 16)`, hangs (killed by `timeout 25`, no
output) -- the eighteenth connect never completes.

Existing `tests/io/tcp_listen_close`, `tcp_loopback`, `tcp_poll`,
`tcp_spawn`, `tcp_short_recv` and `port_bound` print their expected
lines on this tree. The cluster-backed `gates/` harness is not available
in this environment; the test file follows the `tests/<ns>/[a-z0-9_]+.bend`
rule and is well under the token ceiling.

## Duplicate check

No open or closed issue or PR in `bendlang/bend` mentions `backlog`,
`SOMAXCONN` or the listen depth; `WONTFIX.txt` does not cover it.

This is independent of #972 (`SO_REUSEPORT` on the same socket); the two
touch different lines of `tcp_listen.c` and apply in either order.


bend2/effs/tcp_listen.js
```diff
@@ -15,7 +15,7 @@ function tcp_listen(port) {
     sys.close(fd);
     return io_fail(22);
   }
-  if (sys.bind(fd, sys.ptr(at), 16) < 0 || sys.listen(fd, 16) < 0
+  if (sys.bind(fd, sys.ptr(at), 16) < 0 || sys.listen(fd, sys.mac ? 128 : 4096) < 0
```
nicolas-abril: We could just have a big constant like bun's 512 in both C and JS, guaranteeing same behaviour. The OS would cap it if it's over the limit (macOS). If it works for bun which is pretty fast it owuld probably work here too

nicolas-abril (changes requested):

aldeni:
Thanks, that's simpler and better: one number in both lanes, the same behaviour everywhere up to the platform's cap. Done in 3188cdbe: `listen(fd, 512)` in C and JS, with a comment that the kernel caps it (`net.core.somaxconn` / `kern.ipc.somaxconn`).

I re-measured the burst with `demos/io_http_server` from this tree (10,000 connections opened at once, each sending one `GET /`, 10 s window):

| backlog | answered | timed out | last answer |
|---|---|---|---|
| 16 | 9,920 | 80 | 6.51 s |
| `SOMAXCONN` (4096) | 10,000 | 0 | 3.39 s |
| 512 | 10,000 | 0 | 2.37 s |

So 512 loses nothing here. `tcp_listen_backlog` and the other TCP tests pass on the check, interpreted, JS and C lanes. I squashed the change into the PR's single commit and updated the description to match.

PR #978, opened 2026-09-22
TCP gets the byte twins File already has: recv_bytes, send_bytes and poll_bytes

aldeni:
## What changes

TCP gets the byte twins `File` already has, with the same types and the
same shape:

| text | bytes |
|---|---|
| `File.read` | `File.read_bytes` (today) |
| `File.write` | `File.write_bytes` (today) |
| `TCP.recv` | **`TCP.recv_bytes`** |
| `TCP.send` | **`TCP.send_bytes`** |
| `TCP.poll` | **`TCP.poll_bytes`** |

`List<&2, U32>`, exactly as `File.read_bytes` and `File.write_bytes`
already use. No new type and no new concept. The text effects keep their
code unchanged, now under `#ifdef CID_TCP_*` the way `effs/file_read.c`
already guards its pair, since each pair shares one effect file; the send
loop and the poll deadline are shared rather than copied, and the JS lane
mirrors the C lane through the same pack-shaped parameter.

`TCP.poll_bytes` is in the set because `TCP.poll` is the only receive
with a deadline: without it, a server that drops idle connections cannot
read bytes.

## Why

A socket carries bytes. `TCP.recv` answers a `String`, so `io_str`
decodes each read as UTF-8 and writes U+FFFD over whatever it cannot
decode. Two kinds of payload do not survive that.

**A body that is not text** — an image, a protobuf frame, a gzip
stream — loses every byte that is not valid UTF-8. `FF 00 80` reads back
as three replacement characters, and `TCP.send` cannot spell those bytes
either, since it encodes the String as UTF-8.

**A valid character split across two reads.** This is the one that bites
a correct program. `"café"` is `63 61 66 C3 A9`; the network may deliver
`C3` at the end of one read and `A9` at the start of the next. Each read
decodes on its own, with no state carried over, so both halves become
U+FFFD. Measured against a Bend receiver:

```
sent:  63 61 66 C3 | A9          ("café", split by the network)
got code points: 99 97 102 65533
got code points: 65533
```

Sent whole, the same bytes give `99 97 102 233`. Nothing in that program
is wrong: the split is the network's choice, it depends on load, and the
loss is silent.

`File` already draws this line. `File.read` decodes; `File.read_bytes`
answers, in the words of the comment in `effs/file_read.c`, "the bytes as
they are (0..255), one List cell each; a text reader would decode them as
UTF-8". A socket is the same seam as a file, and today only the file has
both doors.

This is also where #809 leaves things. That issue found the two lanes
decoding invalid UTF-8 differently, and the fix in 2.0.7 made the C
runtime a forward WHATWG decoder, one U+FFFD per maximal subpart, exactly
as `TextDecoder` does. That is the right decode — and it is precisely
why a decoding reader cannot be the only way to read a socket: now that
both lanes agree, they agree on discarding the bytes. The decoder is
correct; what is missing is the door that does not decode.

## Alternatives considered

**Make `TCP.recv` carry decoder state between calls**, so a character
split across two reads is stitched back. Rejected: it would make a
`recv` stateful and still leave no way to read a byte that is not text,
so the binary case stays broken. It would also change what an existing
`TCP.recv` returns.

**Preserve undecodable bytes inside the String** (surrogate escapes, as
Python's `surrogateescape` does). Rejected: a `String` would stop being
a sequence of valid code points, and the split-character case still needs
the program to re-decode by hand — the work of bytes, with a type that
says text.

**Ship only `recv_bytes`/`send_bytes`** and leave `poll_bytes` out.
Rejected: `TCP.poll` is the only receive with a deadline, so a server
that times out idle connections would be forced back to text.

`UDP.recv_from` and `UDP.poll` have the same seam and are not touched
here, to keep this reviewable; happy to follow up with the same pair for
UDP if you want it.

## Test output

`tests/io/tcp_bytes_split.bend` sends a character in two halves and then
three bytes that are not text, reading both back over loopback in one
program:

```
first read: 99 97 102 195
second read: 169
not text: 255 0 128
```

Checked, interpreted, JS-built and C-built on this tree
(`bun bend2/main.ts <file> [--check-only | -o x.js | -o x]`): all four
lanes print those three lines. The same exchange through `TCP.recv`
gives `99 97 102 65533` / `65533` instead, which is the defect.

No regressions: `tcp_listen_close`, `tcp_loopback`, `tcp_poll`,
`tcp_spawn`, `tcp_short_recv`, `tcp_send_slow_peer`,
`tcp_connect_black_hole` and `port_bound` print their expected lines on
both runnable lanes. The cluster-backed `gates/` harness is not available
in this environment; the test file follows the
`tests/<ns>/[a-z0-9_]+.bend` rule, takes a free port by retrying upward
as `tcp_loopback` does, and is well under the token ceiling.

## Duplicate check

No open issue or PR asks for byte-level socket effects. #809 (closed,
fixed in 2.0.7) is the decode-divergence issue this builds on, and #846
asks about bundled HTTP/TLS/crypto libraries, which is a different
question — this adds no library, only the byte door on the effect that
already exists. `WONTFIX.txt` does not cover it.


guide/GUIDE.md
```diff
@@ -363,6 +363,12 @@ answers the command line, less the runtime's own options (a `--` ends them). A
 handle (`File`, `Socket`, `Window`) is an affine, opaque value, so every effect
 on one hands it back beside its result, and no program can forge or reuse one.
 
+A file and a socket carry bytes, and each has a text reader and a byte reader:
```
nicolas-abril: This is not needed in the guide, it's pretty self explanatory by the function signatures

nicolas-abril:
Not a criticism of your PR, but i think it would make more sense to have these functions take an array of bytes input in the future

nicolas-abril (changes requested):

nicolas-abril:
Tested on macOS (arm64), merged onto 2.0.26. This looks good to me: the new effects match `File.read_bytes`/`File.write_bytes` exactly (same `List<&2, U32>`, same shared-`.c` layout, same `EINVAL` rule), `tests/io/tcp_bytes_split` passes in the check, interpreted, JS and C lanes, the existing TCP and file tests pass in all three runtime lanes, and `gates/repo.ts` passes.

Two requests before merge:

**1. Test the paths `tcp_bytes_split` does not reach.** Nothing exercises `TCP.poll_bytes`, the `EINVAL` path of `TCP.send_bytes`, or EOF. The test below covers them and passes in the interpreted, JS and C lanes on this branch: an empty send, a byte past 255 (fails with `EINVAL`, and a poll right after shows nothing went out), `poll_bytes` idle / with data / at EOF, and `recv_bytes` at EOF.

<details><summary><code>tests/io/tcp_bytes_edges.bend</code></summary>

```python
# io (io_tcp_bytes_edges): the byte effects' edges. An empty send_bytes
# succeeds; a byte past 255 fails with EINVAL and sends nothing (the
# poll after it finds the socket empty); poll_bytes answers None{} when
# idle, Some{bytes} when data waits, and Some{[]} at EOF, as recv_bytes
# answers [] there. The listener retries upward from a base port on
# Fail, as tcp_loopback does.
import Base

law show_bytes:
  for xs: List<&2, U32>
  String

def show_bytes(xs):
  match xs:
    case Nil{}:
      ""
    case h <> t:
      U32.show(h) ++ (" " ++ show_bytes(t))

law show_res:
  for r: Result<&1, &1, (U32 & String), Unit>
  String

def show_res(r):
  match r:
    case Done{u}:
      "sent"
    case Fail{e}:
      (code, message) = e
      U32.show(code) ++ (" " ++ message)

law show_poll:
  for m: Maybe<&1, List<&2, U32>>
  String

def show_poll(m):
  match m:
    case None{}:
      "none"
    case Some{v}:
      "some: " ++ show_bytes(v)

law said:
  for m: Socket & Result<&1, &1, (U32 & String), Unit>
  IO(Socket)

def said(m):
  (s, r) = m
  do IO<Socket>:
    u : Unit <- IO.print(show_res(r))
    return s

law polled:
  for tag: String
  for m: Socket & Result<&1, &1, (U32 & String), Maybe<&1, List<&2, U32>>>
  IO(Socket)

def polled(tag, m):
  (s, r) = m
  do IO<Socket>:
    v : Maybe<&1, List<&2, U32>> <- IO.pass(Maybe<&1, List<&2, U32>>, r)
    u : Unit <- IO.print(tag ++ show_poll(v))
    return s

law got:
  for tag: String
  for m: Socket & Result<&1, &1, (U32 & String), List<&2, U32>>
  IO(Socket)

def got(tag, m):
  (s, r) = m
  do IO<Socket>:
    xs : List<&2, U32> <- IO.pass(List<&2, U32>, r)
    u  : Unit <- IO.print(tag ++ show_bytes(xs))
    return s

law run:
  for l: Listener
  for c: Socket
  for a: Socket
  IO(U32)

def run(l, c, a):
  do IO<U32>:
    s0 : Socket & Result<&1, &1, (U32 & String), Unit> <- TCP.send_bytes(c, [])
    c1 : Socket <- said(s0)
    s1 : Socket & Result<&1, &1, (U32 & String), Unit> <- TCP.send_bytes(c1, 7 <> (256 <> []))
    c2 : Socket <- said(s1)
    p0 : Socket & Result<&1, &1, (U32 & String), Maybe<&1, List<&2, U32>>> <- TCP.poll_bytes(a, 64, 100)
    a1 : Socket <- polled("empty: ", p0)
    s2 : Socket & Result<&1, &1, (U32 & String), Unit> <- TCP.send_bytes(c2, 1 <> (2 <> (255 <> [])))
    c3 : Socket <- said(s2)
    p1 : Socket & Result<&1, &1, (U32 & String), Maybe<&1, List<&2, U32>>> <- TCP.poll_bytes(a1, 64, 1000)
    a2 : Socket <- polled("data: ", p1)
    x  : Unit <- Socket.close(c3)
    r0 : Socket & Result<&1, &1, (U32 & String), List<&2, U32>> <- TCP.recv_bytes(a2, 64)
    a3 : Socket <- got("eof: ", r0)
    p2 : Socket & Result<&1, &1, (U32 & String), Maybe<&1, List<&2, U32>>> <- TCP.poll_bytes(a3, 64, 1000)
    a4 : Socket <- polled("eof poll: ", p2)
    y  : Unit <- Socket.close(a4)
    z  : Unit <- Listener.close(l)
    return 0

law accepted:
  for c: Socket
  for p: Listener & Result<&1, &1, (U32 & String), Socket>
  IO(U32)

def accepted(c, p):
  (l, ar) = p
  do IO<U32>:
    a : Socket <- IO.pass(Socket, ar)
    run(l, c, a)

law main_go:
  for b: Listener & U32
  IO(U32)

def main_go(b):
  (l, port) = b
  do IO<U32>:
    c : Socket <- IO.try(Socket, TCP.connect("127.0.0.1", port))
    p : Listener & Result<&1, &1, (U32 & String), Socket> <- TCP.accept(l)
    accepted(c, p)

law tcp_retry:
  for port : U32
  for r    : Result<&1, &1, (U32 & String), Listener>
  for rest : IO(Listener & U32)
  IO(Listener & U32)

def tcp_retry(port, r, rest):
  match r:
    case Done{l}:
      IO.pure(Listener & U32, (l, port))
    case Fail{e}:
      rest

law tcp_listen_any:
  for tries : Nat
  for +port : U32
  IO(Listener & U32)

def tcp_listen_any(tries, port):
  match tries:
    case 0n:
      IO.die(Listener & U32, 1, "no free tcp port")
    case 1n+t:
      rest = tcp_listen_any(t, (port + 1 : U32))
      do IO<Listener & U32>:
        r : Result<&1, &1, (U32 & String), Listener> <- TCP.listen(port)
        tcp_retry(port, r, rest)

law main:
  IO(U32)

def main():
  do IO<U32>:
    b : Listener & U32 <- tcp_listen_any(32n, 21013)
    main_go(b)

#|sent
#|22 Invalid argument
#|empty: none
#|sent
#|data: some: 1 2 255
#|eof:
#|eof poll: some:
```

</details>

**2. (Optional) Share the receive loop in `tcp_recv.c`.** `tcp_poll.c` shares `tcp_poll_end` and `tcp_send.c` shares `tcp_send_more` between the text and byte effects, but `tcp_recv.c` copies the whole `recv`/`EAGAIN` loop into `tcp_recv_bytes_more`. Sharing it, as the other two files do, would keep the pair from drifting.

**Suggestion for a separate PR:** the runtime has no helper to build or walk a list. Every effect that does it hand-rolls the loop: `args.c`, `file_read.c` and `window_frame.c` build lists; `file_write.c` and `audio.c` walk them; the JS twins do the same. With this PR, the bytes-to-list loop exists three times in C and three times in JS, and the list-to-buffer loop twice in each. Helpers next to `io_str`/`io_cstr` (C) and `io_text`/`io_bytes` (JS), such as `io_list_u8` and `io_take_u8`, or a generic version that takes an element converter, would cover all of them. That is a runtime refactor across several effects, so it doesn't belong in this PR.

FYI on performance, since the byte effects might look like the fast path: they aren't. A `List<U32>` cell is the same 2-word cell a `String` uses per character. Draining 256 MB over loopback and counting every cell, `recv_bytes` is about 15% slower than `recv` in the C lane (0.52 s vs 0.45 s); in the JS lane it is faster on ASCII (1.72 s vs 2.56 s) and slower on 2-byte UTF-8 (1.76 s vs 1.33 s, twice the cells). The reason to use them is correctness, which is the right reason for this PR.

*Note: this comment was written by an AI after it tested the PR for me and I reviewed the findings. If anything here is wrong, reply and I will review it myself.*

aldeni:
Thank you for testing it on macOS, and for writing the edge test. It's in as `tests/io/tcp_bytes_edges.bend`, unchanged, with you as co-author (5b74688e); it passes the check, interpreted, JS and C lanes here (Linux x86-64).

I also took the optional one, since it's the same shape the other two files already have: in 27454fd5 `tcp_recv.c` has a single `tcp_recv_step` shared by `TCP.recv` and `TCP.recv_bytes`, each keeping only its pack function (text or a list of bytes). The guide paragraph is gone too. The 19 `tcp_*`, `udp_*`, `file_*` and `port_bound` tests keep their output on all four lanes.

On arrays of bytes: agreed. Your measurement matches what I saw (a `List` cell per byte costs what a `String` char costs), so when a byte array type lands in Base these signatures should move to it. The list helpers (`io_list_u8` / `io_take_u8`) sound right to me too; I'd be glad to try that refactor as a separate PR if you'd like.

One small edge I noticed while testing, in case it's useful: with `max = 0`, `recv(fd, buf, 0)` returns 0, so when data is waiting `TCP.poll` answers `Some{""}` (and `TCP.poll_bytes` `Some{[]}`), which reads as the peer's close while the peer is still open. `TCP.poll` already behaves this way, so I left it out of this PR; refusing `max = 0` with `EINVAL` in both would fix it, here or separately, whichever you prefer.

aldeni:
Rebased onto main (95317d95): #1042 changed how effect files name things, and the branch no longer merged. The new effects now follow its form in both lanes: `CID(TCP.recv_bytes)` and the other ids in C, under `#ifdef CID(TCP.*)` guards as `file_read.c` has them, and in JS the `CID(Con)` / `CID(Some)` tags and one `io_eff(CID(...), ...)` per effect. Applied to these six files as they stood before #1042, the same respelling gives main's versions byte for byte, so the port changes nothing but the names. The three commits are the same, and your edge test is in unchanged, with you as co-author.

Re-run on the new base, Linux x86-64:

- The 19 `tcp_*`, `udp_*`, `file_*` and `port_bound` tests, `tcp_bytes_split` and `tcp_bytes_edges` included, print their expected lines on the check, interpreted, JS and C lanes.
- The `tests/io` checkup (119 modules) is identical to main's but for the two new tests' sections.
- `gates/repo.ts` passes 46 / 46.
- Each pair shares one effect file, and the runtime now refuses an effect registered twice, so I also built programs that use both effects of a pair: `tcp_bytes_split` already mixes `recv` and `recv_bytes`, and a scratch program mixing `send` / `send_bytes` and `poll` / `poll_bytes` passes all four lanes.

macOS was not re-run here after the rebase.

PR #981, opened 2026-09-22
refactor(effects): inline one-use C result adapters

tachytelicdetonation:
`IO.get_env` and `UDP.bind` each had a C helper with one caller, an output pointer, and a second error check in the effect handler. Fold the helper into that handler so it returns the Bend result directly.

Net **−27 lines / −136 ttok**, across two effect files. The environment lookup still rejects embedded NULs and reports ENOENT for misses. UDP still validates the port, closes a failed socket, and saves errno before closing. No comments or features removed.

Verification on macOS arm64:

- Eleven existing IO tests passed through default execution, emitted JS (Bun), and native C: **33/33**, matching the baseline. Includes environment ownership/misses, NULs, failed IO.try, UDP loopback, parking, polling, truncation, and invalid addresses/ports.
- Held UDP port 19571 open externally; the existing native loopback test recovered by trying the next port before and after the change.
- All 16 runtime benchmarks emitted byte-identical C and JS against main (`a4952426`).
- Repository gate: **46/46**; `git diff --check` passed.

The private cluster test/performance gates and Linux execution were not run.


nicolas-abril:
Thank you, both inlines read right: we walked every path in `get_env.c` and `udp_bind.c` against main and ran tests/io in the C and JS lanes with no change. One more of the same shape is left: `bend2/effs/tcp_listen.c` has the identical one-use adapter (`tcp_listen(port, &out)` plus a run function that re-checks `q != 0`). If you fold that one in too, the PR is complete by its own title and we will merge it.

nicolas-abril (changes requested):
.

tachytelicdetonation:
@nicolas-abril 
Folded `tcp_listen.c` the same way in 2ce7c00c (−10 lines, −47 ttok; PR total now −37 lines / −183 ttok). The socket still closes on every failure path and errno is saved before the close. Checked on macOS arm64: all 7 `TCP.listen` tests in `tests/io` pass through native C, the whole `tests/io` C lane gives the same results before and after, a second listen on a busy port still reports "Address already in use", and the repo gate is 46/46.

PR #982, opened 2026-09-22
refactor(effects): share descriptor close implementations

tachytelicdetonation:
`File.close`, `Socket.close`, and `Listener.close` repeat the same C operation: close the handle's host descriptor and return Unit. Share that handler and its registration in `fd_close.c`, guarding each CID so programs using any subset still compile. The JS socket and listener effects also share their identical function; the JS file effect retains its existing `fs.closeSync` implementation.

Net **−20 lines / −126 ttok**, including the Base import changes. Three fewer source files, with the same public effects and close-error behavior. The shared C file is 134 ttok against its 4,000 cap.

Verification on macOS arm64:

- All 24 existing IO tests that use these close effects passed through default execution, emitted JS (Bun), and native C: **72/72**, matching the baseline.
- All seven nonempty combinations of the three effects passed through those three paths before and after: **21/21 per revision**. These cover each effect alone, every pair, and all three together.
- All 16 runtime benchmarks emitted byte-identical C and JS against main (`a4952426`).
- Repository gate: **46/46**; `git diff --check` passed.

The private cluster test/performance gates and Linux execution were not run. This is independent of #981.


nicolas-abril:
Thank you for this. `fd_close.c` is correct: we ran tests/io in both lanes and inspected the emitted C for File-only, Socket-only, Listener-only and all three, and the `#ifdef` guards register exactly the live closes. We are going to keep one file per effect here, though: the three closes are one syscall each, the files are small and finished and will not drift, and each effect's source being where its name says is worth more to us than the shared lines. That is a stylistic call, not a fault in the work, so closing this one; the work is appreciated.

PR #985, opened 2026-09-22
Share stdout write/print effect sources (negative lines, same behavior)

jkbennitt:
## Summary

`IO.write`, `IO.print`, and (on JS) `IO.print_err` are the same stdout/stderr
wrapper with a file descriptor and optional newline. Today that logic lives in
near-copy files:

- `effs/write.js` / `print.js` / `print_err.js` (7 lines each; same `io_out` shape)
- `effs/write.c` / `print.c` (nearly the same `io_cstr` → `io_out` run)

This follows the same simplification Bend already merged for audio and file
effects: related effects share one source; Base keeps importing that one path;
`eff_src` still embeds a file only once.

`print_err.c` stays its own file: it correctly routes through `io_errs`, which
is not the same as `io_out(stderr, …)`.

No feature removed. No comment stripping. Behavior of `IO.write` / `IO.print` /
`IO.print_err` unchanged.

## Negative diffstat

```
 AGENTS.md               |  1 +
 bend2/base.bend         |  6 +++---
 bend2/effs/print.c      | 28 +++++++++++++++++++++-------
 bend2/effs/print.js     | 10 ++++++++++
 bend2/effs/print_err.js |  7 -------
 bend2/effs/write.c      | 18 ------------------
 bend2/effs/write.js     |  7 -------
 7 files changed, 35 insertions(+), 42 deletions(-)
```

Net **−7** lines (honest: smaller than a rough pre-patch estimate — the
`#ifdef` wrappers use most of what the shared `io_stdout_run` helper saves).

## Test plan

- [x] Local full-matrix mirror of `gates/test.ts` on this branch: **PASS: 1427 / 1427** (check / interp / JS / C; ~158s). Official mini-cluster `gates/test.ts` was not run from our side (no bastion/`cluster` SSH).
- [x] Smoke: `IO.write` / `IO.print` / `IO.print_err` alone and together on interpreter, emitted JS, and native C (including empty strings / UTF-8 / newline vs no-newline); matched pre-patch HEAD.
- [x] `gates/repo.ts`: **PASS: 46 / 46** (with `ttok` on PATH).
- [ ] No bench intent to change; runtime pins should be unaffected.

## Notes

Happy to reshape naming (`stdout.c` vs keeping `print.c` as the shared home) to
match whatever convention you prefer after #950/#939.


nicolas-abril:
Thank you for this. The sharing is correct: we emitted every test on main and on this branch and the only lines that differ are the effect sources themselves, and the per-effect `#ifdef` gating checks out for print-only, write-only, print_err-only and all three together. We are going to keep the one-file-per-effect layout for stdout, though: these files are small, essentially finished and will not drift, and a reader looking for `IO.write` should land in `write.c` rather than on a flag inside `print.c`. That is a stylistic call, not a fault in the work, so closing this one; the work is appreciated.

Issue #970, opened 2026-09-21
C backend: list reconstruction leaks 16 bytes per call in 2.0.24


mewhhaha:
### What you did

bend main.bend -o /tmp/bend-boxing-repro

### What happened

The command above compiled the original test harness; the harness and allocation probe are omitted here for brevity. **repro.bend below is the reduced function to inspect**, not a standalone executable.

Calling rebox with a fresh singleton containing Scalar, then fully consuming the result, leaks **16 bytes per call**. The same test with Pair does not leak. Expected: neither variant accumulates unreachable allocations.

Measured in-use bytes over four iterations: Scalar 1072 → 1088 → 1104 → 1120; Pair 1056 → 1056 → 1056 → 1056. These measurements subtract allocator free-list storage, so this is not RSS growth or allocator retention.

In the generated C, the original two-word list cell is reused when boxing Pair, but neither reused nor freed when boxing Scalar. Both branches allocate a new output list cell, leaving the original 16 bytes orphaned in the Scalar case.

Likely cause: val_box/ctr_build share mutable spare-node bookkeeping across mutually exclusive branches.

### The file

repro.bend

```python
import Base

type Value is Data:
  Scalar{number: U32}
  Pair{left: Nat, right: Nat}

def rebox(values: List<&2, Value>) -> List<&2, Value>:
  match values:
    case Nil{}:
      Nil{}
    case value <> tail:
      value <> tail
```

### bend --version

bend 2.0.24

Obtained with bend version; bend --version reports an unknown option.

### uname -sm

Linux x86_64

### clang --version (the first line)

clang version 22.1.8

PR #987, opened 2026-09-22
A List Cons spare survives val_box over a sum (#970)

YidaWeng:
## Summary
- `val_box` shared one `fl.spares` list across exclusive sum arms, so boxing `Pair` could splice a parked `List` Cons cell that the `Scalar` arm never saw; rebuilding `Con` then always allocated and the Scalar path leaked 16 bytes per call (#970).
- Each arm now starts with an empty spare list, flushes leftovers, and restores the outer list so `Con` keeps the spare; `Pair` allocates its own node.
- Adds `tests/run/list_rebox_sum.bend` (Scalar + Pair rebox).

## Test plan
- [x] `bun bend2/main.ts tests/run/list_rebox_sum.bend` — prints `7` / `3`
- [x] Same via `-o …`.js (bun) and native C
- [x] `tests/run/spare_reuse.bend` still passes (interp + C)
- [x] Generated C for the repro: Pair `heap_alloc`s; Cons reuses `_sp_0`
- [x] `bun gates/repo.ts` — PASS 46 / 46
- [ ] Cluster `gates/test.ts` (no minis here)

Made with [Cursor](https://cursor.com)


nicolas-abril:
Thank you, the fix is right: with a list built at runtime, main's generated C leaves `_sp_0` (the Cons cell) neither reused nor freed on the Scalar path, and this branch heap_allocs the Pair and hands `_sp_0` to the Cons rebuild, so no path orphans the cell. Two things before it can merge.

1. The test does not reach `val_box`. Both calls in `main` take literal lists, the compiler folds them, and the C emitted for tests/run/list_rebox_sum.bend is byte-identical on main and on this branch (no `rebox`, `head_u32` or `head_nat` code at all; it passes on main). The list has to come from a runtime value. Building the `Value` from `List.length` of `IO.args()` does it, for example:

   ```
   def wrap(n: Nat) -> Value:
     match n:
       case 0n:
         Scalar{7}
       case 1n+p:
         Pair{p, 4n}

   def main() -> IO(Unit):
     do IO<Unit>:
       xs : List<String> <- IO.args()
       +n : Nat = List.length(&1, String, xs)
       Unit <- IO.print(U32.show(head_u32(rebox(wrap(n) <> Nil{}))))
       IO.print(Nat.show(head_nat(rebox(wrap(Nat.add(n, 4n)) <> Nil{}))))
   ```

   With that shape the two emitted C files differ in exactly the two lines the fix changes. The harness cannot observe a leak by output, so please say so in the header instead of "Regression for #970": it is a smoke test that keeps the shape compiling and running.

2. The commit message should name the trade-off: hiding the outer spares from the arms means an arm that reused a same-size spare on main now heap_allocs, and the spare is recovered only if a same-size build follows; otherwise it is freed at the next flush. Right for the Cons shape, but a maintainer reading the log should see it.

YidaWeng:
Both points are in f7acd448 (test and commit message amended in place).

1. The test now builds the list at runtime exactly as you sketched (`wrap` from `List.length` of `IO.args`). Emitting C for it on main vs this branch differs in only the two lines the fix moves:
```c
// main                                // this branch
u64 _nd_2 = _sp_0;                     u64 _nd_2 = heap_alloc(e, cls_fit(2));
u64 _nd_3 = heap_alloc(e, cls_fit(2)); u64 _nd_3 = _sp_0;
```
On main `_sp_0` goes into the Pair node and the Cons rebuild allocates, so the Scalar path orphans the cell; here the Pair allocates and the rebuild keeps `_sp_0`. The header no longer says "Regression" — it is a smoke test, notes a leak is not observable in the output, and that a literal list folds away on both. Interp, C and JS all print `7` / `3`; `tests/run/spare_reuse.bend` still prints `10n` (interp + C); `bun gates/repo.ts` PASS 46 / 46.

2. The commit message now carries the trade-off verbatim: arms no longer see outer spares, so an arm that reused a same-size spare on main now heap_allocs, and the spare comes back only if a same-size build follows, otherwise the next flush frees it.

PR #992, opened 2026-09-22
fix(loader): register in Node worker threads

vicmcorrea:
## Why

i'm evaluating Bend for small survey-scoring helpers in our TypeScript monorepo. While trying a Node worker to keep scoring off the main thread, the `.bend` import failed with `ERR_UNKNOWN_FILE_EXTENSION`. The same import works on the main thread.

## What changed

this lets application workers register the existing loader, while still excluding Node's internal loader thread. One condition changed, zero net lines added.

Node added [`isInternalThread`](https://nodejs.org/api/worker_threads.html#worker_threadsisinternalthread) in 22.14.0 and 23.7.0. The explicit `=== false` preserves main-thread-only registration on older versions where that property is unavailable.

## Validation

Checked on macOS arm64:

- Node 22.14.0 and 24.19.0 passed main-thread, inherited-preload, nested-worker and explicit-worker imports. The three worker cases fail on the original code.
- Node 22.13.1 preserves the original behavior. Main-thread imports work, workers still reject `.bend`. Node 22 tests used `--experimental-strip-types`.
- Bun 1.3.11 preload imports and CLI version output passed. Repository gate passed, 46/46, and `git diff --check` passed. Cluster tests were not run.

## Reproduction

<details>
<summary>Run from the checkout root with Node 24</summary>

The existing test gate runs Bend programs through Bun and the native backend, so this checks Node's loader directly.

```sh
dir="$(mktemp -d)"
trap 'rm -rf "$dir"' EXIT
cat > "$dir/score.bend" <<'BEND'
import Base

def score(x: U32) -> U32:
  (x * 3 + 7 : U32)
BEND
cat > "$dir/worker.mjs" <<'JS'
import assert from "node:assert/strict";
import {Worker, isMainThread, parentPort} from "node:worker_threads";

if (isMainThread) {
  const worker = new Worker(new URL(import.meta.url));
  worker.on("message", value => {
    assert.equal(value, 22);
    console.log(value);
  });
  worker.on("error", error => {
    console.error(error);
    process.exitCode = 1;
  });
} else {
  const {default: scoring} = await import("./score.bend");
  parentPort.postMessage(scoring.score(5));
}
JS
BEND_NO_TELEMETRY=1 node --import ./bend2/main.ts "$dir/worker.mjs"
```

Current main exits 1 with `ERR_UNKNOWN_FILE_EXTENSION`. With this change it prints `22` and exits 0.

</details>


vicmcorrea:
bugbot run

nicolas-abril:
Thank you, merged. We reproduced it: on main, a Node worker preloaded with `--import bend2/main.ts` never registered the loader and failed its `.bend` import with `ERR_UNKNOWN_FILE_EXTENSION`. With this change the import works in plain, nested and concurrent workers on Node 22.14, 24 and 26, and 22.13 behaves as before. Bun never reaches this line, and the local test suite is unchanged, so the one-condition fix went in as it is.

PR #997, opened 2026-09-22
Fix the shared array redirect race

### LINKED ISSUE #975: Shared Array redirect lookup races with atomic reference-count updates
## What you did

On unmodified commit `e52cda47a58967aa65d1eb26efe8f42a0b0407df`, compile this program to C and run the result under ThreadSanitizer with at least two workers:

```sh
bun bend2/main.ts redirect-race.bend -o /tmp/redirect-race.c
clang -x c -std=gnu11 -O1 -g -fsanitize=thread -pthread /tmp/redirect-race.c -o /tmp/redirect-race
TSAN_OPTIONS='halt_on_error=0 exitcode=0' /tmp/redirect-race --threads 2 --gpu off
```

## What happened

ThreadSanitizer reliably reports an 8-byte non-atomic read in `blk_loc` racing with a 4-byte atomic write in `term_keep` / `rfc_bump`:

```text
WARNING: ThreadSanitizer: data race
  Read of size 8 ... by thread T1:
    #0 WL_FID_FAN
    #1 monk_step
    #2 pool_work

  Previous atomic write of size 4 ... by thread T2:
    #0 WL_FID_FAN
    #1 monk_step
    #2 pool_work

SUMMARY: ThreadSanitizer: data race in WL_FID_FAN
4
ThreadSanitizer: reported 1 warnings
```

The redirect cell is initialized as:

```c
e.mem[r] = ((u64)term_loc(t) << 24) | cnt;
```

`blk_loc` reads the complete cell as a plain `u64`:

```c
return blk_shr(a) ? H[term_loc(a)] >> 24 : term_loc(a);
```

At the same time, `rfc_bump` and `term_drop` perform atomic 32-bit read-modify-write operations on the low half of that same `u64`. Mixing the overlapping non-atomic 64-bit access and atomic 32-bit access is a C data race even though the returned location appears stable in practice.

I could not manifest this as an incorrect or nondeterministic Bend result. The reproducer prints the expected `4`, including when ThreadSanitizer continues after reporting the race. A deeper version with 262,144 concurrent leaves, hundreds of repetitions, worker counts 2/4/8/16, Apple Clang and Homebrew Clang/GCC at several optimization levels including LTO, and arm64 and Rosetta x86-64 all produced the expected result.

There is a runtime invariant that explains that result: valid reference-count changes are confined to the low 24 bits, while `cell >> 24` returns the immutable target location. A live reader owns a reference, so the redirect wrapper cannot be freed and reused during that read. Refcount overflow is rejected with `ERR_RFCS`. Thus this is a confirmed language-level C race and possible portability/miscompilation issue, but there is currently no known Bend program whose unsanitized result it corrupts.

Using the existing atomic reconstruction in `rfc_view` for shared `blk_loc` calls removes the ThreadSanitizer report. A representation with separately addressable location and reference-count fields would also avoid overlapping accesses.

## The file

```bend
import Base

@unsafe
def leaf(a: Array<U32>) -> Array<U32>:
  Array.set.fin(U32, Array.atomic.add(a, 0, 1))

@unsafe
def fan(+d: Nat, +a: Array<U32>) -> Array<U32>:
  match d:
    case 0n:
      leaf(a)
    case 1n+p:
      l r = fan(p, a) fan(p, a)
      Array.join(U32, l, r)

def done(r: Array<U32> & U32) -> IO(Unit):
  (a, x) = r
  IO.print(U32.show(x))

def main() -> IO(Unit):
  done(Array.get(U32, fan!(2n, [0 : U32*1n]), 0))

#|4
```

## Environment

- `bend --version`: `bend 2.0.24`
- Commit: `e52cda47a58967aa65d1eb26efe8f42a0b0407df`
- `uname -sm`: `Darwin arm64`
- `clang --version`: `Apple clang version 17.0.0 (clang-1700.6.4.2)`

uta080522:
Fixes #975.

`blk_loc` reads a shared array's redirect cell as a plain 64-bit value, even though other handles update its reference count with 32-bit atomics. The location bits stay unchanged, but the overlapping access is still a C data race.

This patch reads shared redirects through `rfc_view`, as suggested in the issue. The helper now takes `Corpus` rather than `Env`, which lets `blk_loc` reuse it. Its existing callers still pass the same memory, and the cell layout does not change.

The test is based on the issue's reproducer, with 256 leaves. Its output is already correct before the fix, so ThreadSanitizer is needed to catch the regression.

## Testing

With Bun 1.4.2 and Clang 21.1.8 on WSL Ubuntu 26.04 x86-64, ThreadSanitizer reports the race on unmodified `ff7a40cc` with two workers; the one-worker control passes. After the fix, ten runs each with 1, 2, 4 and 8 workers all print `256` without sanitizer diagnostics. Restoring only the plain read in the emitted C brings the warning back.

The local checks also covered 65 selected programs: this test and 64 existing array, shared-data and closure tests. Both versions passed through the interpreter, generated JS, and native CPU with 1 and 4 workers. The repository gate passed all 46 checks, and `comp.ts` uses 64,972 of its 65,000-token budget, five more than before. `git diff --check` also passed. The GPU backends and the full cluster test and performance gates were not run.

To run the sanitizer test:

```sh
bun bend2/main.ts tests/run/array_redirect_race.bend -o /tmp/array-redirect.c
clang -std=gnu11 -O1 -g -fsanitize=thread -pthread \
  /tmp/array-redirect.c -lm -o /tmp/array-redirect
TSAN_OPTIONS='halt_on_error=1:exitcode=66' \
  /tmp/array-redirect --threads 2 --gpu off
```


nicolas-abril:
Thank you for your contribution. Unfortunately your patch leads to very big performance regressions, so I can't merge it as it is. Especially large on metal, where I'm assuming you can't test yourself.
I actually also had a patch for the issue but it had similar performance issues.

I'd prefer to either find an alternative fix to the data race or change the underlying operations to something else rther than accept these changes as they are.
Feel free to contact me on discord if you need me to tests things out for you.

uta080522:
I've replaced the `rfc_view` lookup with a split redirect. The count and target are separate fields in the same eight-byte slot. An Array handle supplies the target's low eight bits, so `blk_loc` reads only the immutable upper field: one ordinary read, with no atomic load or fence on each access.

The updated description includes the race check and paired CPU timings. I used the shared-array stencil for the performance comparison because the existing runtime benchmarks all compile with `BLK_SHR=0`; they do not exercise this lookup.

nicolas-abril (changes requested):
Thank you, the split redirect fixes the regression. On an Apple M4, with the scaled `stencil3d` from #999 (2^21 cells, 128 leaves, 160 passes, median of 10 runs), this branch is within noise of main on every lane: 518 → 519 ms single-threaded, 120 → 120 ms on 10 threads, 7.23 → 7.32 s on Metal. A control that reads the redirect through `rfc_view` in `blk_loc` gives 1091 ms, 237 ms and 32.9 s, so the benchmark does see the old cost. The 16 runtime benchmarks stay within ±3% of main in all three perf-gate modes, with outputs matching the pins.

One change before merge, in `term_drop`. The last-reference path now calls `rfc_view(e, r)`, which adds an atomic load of the count word, where main read the redirect with one plain load. That load isn't needed: `a32_sub_rel` already returns the old count word, whose top eight bits are the storage's low bits, and `loc_hi` never changes. So the target can be rebuilt from the value in hand plus the immutable half:

```c
u32 old = a32_sub_rel(p, 1);
if ((old & RFC_CNT) != 1) {
  t = 0;
} else {
  a32_acq(p);
  Loc src = ((u64)rfc_at(H, r)->loc_hi << 8) | (old >> 24);
  t = (t & ~(RFC_BIT | LOC_MASK)) | src;
  heap_free(e, 0, r);
}
```

This stays race-free, since it reads only the returned value and the immutable half, and drops one atomic load per last-reference drop. I haven't measured it: the only benchmark that moved with this branch is tree-bitonic, +2.1% single-threaded and +2.8% on 16 threads over 10 runs, and this path is my best guess for it. Could you check whether the change recovers that?

*Note: this review was written by an AI after it tested the PR for me and I reviewed the findings. If anything here is wrong, reply and I will review it myself.*

PR #998, opened 2026-09-22
fix(io): send empty UDP datagrams in Bun

vicmcorrea:
## Why

`UDP.send_to` with an empty string crashes on Bun before sending a packet. Bun's FFI rejects a pointer to an empty byte array. Native Bend sends the same zero-byte datagram successfully.

i'm evaluating Bend alongside our TypeScript monorepo and found this while checking Bun and native behavior.

## What changed

this passes a null pointer when the payload is empty and still calls `sendto`. Nonempty payloads keep the existing path.

The UDP round-trip test keeps the original `bend-udp` ping and adds an empty ping through the same helpers. It checks both replies, `bend-udp-pong` and `-pong`, including delivery and the sender's return address. Two files changed, six net lines added.

## Validation

Checked on macOS arm64 with Bun 1.3.11 and clang 23.1.1:

- On the original Bun backend, the test prints `bend-udp-pong` and then fails on the empty ping. With the fix, Bun and native C print both expected replies and exit 0.
- All six UDP fixtures and `port_bound.bend` pass checking and execution under Bun and native C.
- A local receiver confirmed the actual bytes for empty, ASCII, Unicode and embedded-NUL payloads.
- Local repository gate passed, 46/46, and `git diff --check` passed.

Cluster gates were not run.

<details>
<summary>Run the regression</summary>

From this PR's checkout, with Bun on PATH:

```sh
(
  set -e
  udp_check_dir="$(mktemp -d)"
  BEND_NO_TELEMETRY=1 bun bend2/main.ts tests/io/udp_send_recv.bend -o "$udp_check_dir/udp.js"
  bun "$udp_check_dir/udp.js"
  rm "$udp_check_dir/udp.js"
  rmdir "$udp_check_dir"
)
```

The program exits 0 and prints:

```text
bend-udp-pong
-pong
```

With the original Bun backend, it prints the first line and exits 1 with the empty-memory pointer error.

</details>


nicolas-abril:
The regressions test should probably be an addition. Either modify it to just add another (empty) message in that test or make a small new regression test just for this

vicmcorrea:
> The regressions test should probably be an addition. Either modify it to just add another (empty) message in that test or make a small new regression test just for this

okss 🫡

nicolas-abril (changes requested):
.

vicmcorrea:
> .

hey, saw the change request. i thought the latest update covered your feedback, but might be missing something.... could you point me to what still needs changing?

PR #1000, opened 2026-09-22
refactor(compiler): share graph traversal worklists

byronbenharris:
## Summary

Borrow propagation, fork propagation, and segment reachability each compute a transitive graph closure with separate loops. `close_over` in `bend2/comp.ts` now provides one iterative worklist for all three. Each pass indexes its edges once, avoiding repeated full scans and per-node segment lookups while preserving the same borrow, fork, host, and device reachability sets.

## Test

  - [ ] `bun gates/repo.ts` — FAIL 45 / 46 
      - `bend2/comp.ts`: 65,067 / 65,000 ttok
  - [x] Targeted C and JS builds/runs — PASS
      - `compile/nofk_closure.bend`
      - `compile/fork_closure_stress.bend`
      - `reg/borrow_fork_hold.bend`
      - `reg/borrow_nested_call.bend`
      - `reg/borrow_loop.bend`
      - `reg/borrow_request.bend`
      - `reg/brw_fork_dead.bend`
      - `reg/brw_readers.bend`
      - `reg/spin_refs_dead_seg.bend`
      - `reg/brw_twins.bend`
      - `reg/dead_fork_binding.bend`
  - [ ] Cluster gate — not run; no access to minis


nicolas-abril (changes requested):
Thank you for your contribution. I like the idea here, but 2 things.
- The code doesn't follow the repo's style: Graph is not a defined type and the function you added should be called `graph_close_over` or something similar (smaller would be even better!).
- Even though it's a simplification, you're adding to the token count, this is more code and not less (even though a bit simpler). Looking at the code it seems like you could be simplified/shortened a little further.

